PR #4, opened 2015-04-03
Support for SplType extension

lukaszwit:
http://php.net/manual/en/book.spl-types.php

### DIFF
--- a/SplType.php
+++ b/SplType.php
@@ -1,0 +1,99 @@
+<?php
+
+/**
+ * Abstract parent class for all SPL types.
+ *
+ * @link http://php.net/manual/en/class.spltype.php
+ */
+abstract class SplType {
+    /**
+     * @var null Default value
+     * @link http://php.net/manual/en/class.spltype.php#spltype.constants.default
+     */
+    const __default = null;
+
+    /**
+     * Creates a new value of some type
+     *
+     * @param mixed $initial_value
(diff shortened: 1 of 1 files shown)


neuro159:

lukaszwit:
Just sent

lukaszwit:
Any problem with this?

lukaszwit:
Dear JetBrains can I help you someway with merging it?

aik099:
@lukaszwit , what tool you're using to create that stub?

lukaszwit:
@aik099 I was using phpstrom (EAP at that time)
SplType was enabled in my php and I wrote test lib: https://github.com/lukaszwit/currency

aik099:
No, I mean how did you generate all that method definition with DocBlocks and such?

lukaszwit:
I've wrote it as a compilation of what docs says and what I've experienced when working with this extension. This is not a trivial extension so generating automatically could be sometimes misleading. However I think it is possible to generate it but I don't think Reflection Api would be enough and possibly the source code of this extension should be parsed to generate proper doc blocks.
If you've got an idea or requirement that it should be automatically generated we can think about it.

aik099:
I've tried to reflect the `svn` extension but got empty DocBlocks and methods without parameters as the result. I guess reflection requires something from extension developer to work properly.

PR #15, opened 2015-04-25
Add return type to Closure::__invoke()

BenMorel:
This is necessary to avoid this inspection message:

> 'void' method '__invoke' result used

When calling `__invoke()` directly:

    $function = function() {};
    $function->__invoke();

### DIFF
--- a/Core_c.php
+++ b/Core_c.php
@@ -311,6 +311,9 @@ final class Closure {
      */
     private function __construct() { }
 
+    /**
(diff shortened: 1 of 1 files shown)


morrisonlevi:
In this case there is nothing returned; unless your use-case is different I think the demonstrated behavior is correct.

neuro159:

neuro159:
fixed as part of other stub updates

PR #16, opened 2015-04-25
Method name case fix

BenMorel:
http://php.net/manual/en/oci-lob.writetemporary.php


neuro159:

PR #26, opened 2015-06-11
Using @since (related to WI-16509)

boekkooi:
This PR adds the [`@since`](http://www.phpdoc.org/docs/latest/references/phpdoc/tags/since.html) to the constant, function, define, class, variable, method where the information was already availible.

This PR relates to the comments made in [WI-16509](https://youtrack.jetbrains.com/issue/WI-16509)

The following regex and searches where used:
```
\* Available since PHP ([0-9\.]+)
\* Since PHP ([0-9\.]+)
\* PHP >= ([0-9\.]+)<br\/>
\* \(PHP [0-9] &gt;= ([0-9\.]+)\)<br\/>
``` 
(`* @since $1`)
`\* \(PHP 4 &gt;= 4\.([0-9\.]+), PHP 5\)<br\/>` (`* @since 4.$1, 5.0`)
`* (PHP 4, PHP 5)<br/>` (`* @since 4.0, 5.0`)
`* (PHP 5)<br/>` (`* @since 5.0`)
`(\s*\*\s*)@since ([0-9\.]+), ([0-9\.])` (`$1@since $2$1@since $3`)

### DIFF
--- a/Core.php
+++ b/Core.php
@@ -10,7 +10,8 @@
 function zend_version () {}
 
 /**
- * (PHP 4, PHP 5)<br/>
+ * @since 4.0
+ * @since 5.0
  * Returns the number of arguments passed to the function
  * @link http://php.net/manual/en/function.func-num-args.php
  * @return int the number of arguments passed into the current user-defined
@@ -19,7 +20,8 @@ function zend_version () {}
 function func_num_args () {}
 
 /**
- * (PHP 4, PHP 5)<br/>
+ * @since 4.0
+ * @since 5.0
  * Return an item from the argument list
  * @link http://php.net/manual/en/function.func-get-arg.php
  * @param int $arg_num <p>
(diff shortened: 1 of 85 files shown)


neuro159:
Impressive work! I'll pull this in 9.0.x after additional checks.
Will also work on language-level dependent checks later..
We need to have accompanying @deprecated VERSION tags too...

boekkooi:
@neuro159 So I added some new `@deprecated` annotations and fixed some others to be according to http://www.phpdoc.org/docs/latest/references/phpdoc/tags/deprecated.html 
There is one issue left and that is **removed** method/functions/classes for example [mysqli::client_encoding](http://php.net/manual/en/function.mysqli-client-encoding.php) which was removed in PHP 5.4. Should I add a `@removed  [<version>] [<description>]` tags? This would be a PHPStorm specific tag but I think it is very much needed to fix WI-16509

neuro159:
Thank you. No need for another custom tag.
The work saving both @since & @deprecated versions in index and taking into account on inspection, completion, etc. is still pending, ETA v10 EAP

PR #27, opened 2015-06-16
Update to PDO::query stub to document the 2nd and follow parameters options

Dragonrun1:
Update so stub shows additional parameters can be used especial the second parameter to set fetch mode.


neuro159:

PR #28, opened 2015-06-18
Added third argument to "is_subclass_of"

ziege:
Added third argument to "is_subclass_of" (supported from PHP 5.3.9), which caused the following Code Smell message: ""Method call uses 3 parameters, but method signature uses 2 parameters".


neuro159:

PR #31, opened 2015-07-02
Added missing CURLOPT_KEYPASSWD

nitso:
http://php.net/manual/en/function.curl-setopt.php
https://youtrack.jetbrains.com/issue/WI-27882


neuro159:

PR #36, opened 2015-07-31
Stub Ev - http://php.net/manual/en/book.ev.php

zwilias:


neuro159:

neuro159:

zwilias:
Done.

Do you still need me to submit a CLA?

neuro159:
Nope, we were able to get rid of that.

PR #49, opened 2015-09-27
Added newrelic stub

acim:
Covers version 4.23.x


neuro159:
//Special thanks for filing Contributor Agreement in advance!

PR #56, opened 2015-11-09
Added mysqli_stmt::$error_list

codemasher:
As per https://twitter.com/codemasher/status/662063737919139840


neuro159:

codemasher:
Took a few days, sent it in by facsimile just now!

A note on the PDF: it would be super helpful (and you'd get more readable results) if the second page would be editable!

PR #61, opened 2015-11-19
PECL/rrd extension stub

Haran:
The PECL/rrd extension provides bindings to the RRDtool C library which open source industry standard for logging and graphing system.


neuro159:

Haran:
aaaaaand it's done (c)

PR #66, opened 2015-11-30
Specify function scope in jsonSerialise interface.

sekjun9878:
This makes the method stub autogeneration to add a function scope which is the expected behaviour.

I have signed the Contributor License Agreement.

### DIFF
--- a/json.php
+++ b/json.php
@@ -17,7 +17,7 @@ interface JsonSerializable  {
 	 * which is a value of any type other than a resource.
 	 * @since 5.4.0
 	 */
-    function jsonSerialize ();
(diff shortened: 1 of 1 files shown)


MaXal:
https://youtrack.jetbrains.com/issue/WI-29970

MaXal:
I believe it's a wrong way to fix an initial request since {{public}] keyword is not required here. We have to update all our Interfaces to achieve desired behavior (if it's desired indeed: https://youtrack.jetbrains.com/issue/WI-29971).

neuro159:
Discussion in ticket.

PR #73, opened 2015-12-21
The key_exists function has never been deprecated

T1grOK:
The key_exists function has never been deprecated. Detail https://bugs.php.net/bug.php?id=71174


neuro159:
Obsolete
Please ensure that you follow rules in CONTRIBUTING.md

PR #79, opened 2016-01-26
added RarArchive extension

AmsTaFFix:


rar.php
```diff
@@ -0,0 +1,620 @@
+<?php
+
+/**
+ * This class represents a RAR archive, which may be formed by several volumes (parts) and which contains a number of
+ * RAR entries (i.e., files, directories and other special objects such as symbolic links).
+ *
+ * Objects of this class can be traversed, yielding the entries stored in the respective RAR archive.
+ * Those entries can also be obtained through {@see \RarArchive::getEntry} and {@see \RarArchive::getEntries}.
+ *
+ * @link http://php.net/manual/ru/class.rararchive.php
```
avsej: it is better strip locale from url, and rely on user defaults. PHP.net will redirect to correct locale
http://php.net/manual/class.rararchive.php

neuro159:

rar.php
AmsTaFFix: my mistake, i will change

6ecuk:
Hi, if  any plans for merge this request in master  ? I need RarArchive stub in PhpStorm

andrey-sokolov:
Merged https://github.com/JetBrains/phpstorm-stubs/pull/234

PR #83, opened 2016-02-23
Add all supported arguments to PDOStatement::setFetchMode stub

h3xx:
**Note: I'm not going to sign a contract in order to submit this pull request. Use my contribution or don't. Credit me or don't. I don't care what you do with this code, or the knowledge that this bug exists in your software. I fixed it just because it annoyed me. If you had an issues section, I'd have used that. You don't, so instead you get code.**

This adds a second and third optional argument to PDOStatement::setFetchMode to support this method's multi-call nature:

>public bool PDOStatement::setFetchMode ( int $mode )
>public bool PDOStatement::setFetchMode ( int $PDO::FETCH_COLUMN , int $colno )
>public bool PDOStatement::setFetchMode ( int $PDO::FETCH_CLASS , string $classname , array $ctorargs )
>public bool PDOStatement::setFetchMode ( int $PDO::FETCH_INTO , object $object )

Source: http://php.net/manual/en/pdostatement.setfetchmode.php

This fixes false-positives coming up when running code inspections on perfectly legal usages.

Example of false positive:

Code:

    $statement->setFetchMode(PDO::FETCH_CLASS, $class);

Inspection results:

>Problem synopsis
>    Method call uses 2 parameters, but method signature uses 1 parameters (at line XXX)

### DIFF
--- a/PDO.php
+++ b/PDO.php
@@ -1316,9 +1316,24 @@ class PDOStatement implements Traversable {
 	 * @param int $mode <p>
 	 * The fetch mode must be one of the PDO::FETCH_* constants.
 	 * </p>
+	 * @param int|string|object $arg2 <p>
+     * If $mode is PDO::FETCH_COLUMN, the integer column index to fetch.
+	 * </p>
+     * <p>
+     * If $mode is PDO::FETCH_CLASS, the name of the class to instantiate with
+     * row data.
+     * </p>
+     * <p>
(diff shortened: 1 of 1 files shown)


neuro159:
> If you had an issues section, I'd have used that
Here it is: https://youtrack.jetbrains.com/issues/WI
Also there's a plenty of relevant links under Help menu in PhpStorm
Please don's submit any future patches while breaking CONTRIBUTING.md

PR #90, opened 2016-04-20
Fix case in Memcache method names

jeromegamez:
When using Memcache methods, those get annotated by the "Method call or class usage is not case sensitive" Inspection.

:octocat:


tableau-torgie:
Yes please!

jeromegamez:
Thanks for merging! :octocat:

PR #92, opened 2016-05-12
Update the XDebug stubs

stof:
I updated the doc based on https://xdebug.org/docs/all_functions to include all missing arguments (and I opened a bug on Xdebug to report the fact that they are not exposed through the reflection API: https://bugs.xdebug.org/view.php?id=1300)


stof:
Looking at the issue tracker, this fixes the following ticket: https://youtrack.jetbrains.com/issue/WI-29603

PR #99, opened 2016-06-29
Added `sAddArray` to `Redis` stub

boesing:


Redis.php
```diff
@@ -3089,4 +3105,4 @@ public function _target($key) {}
      * Use this function when a new node is added and keys need to be rehashed.
      */
     public function _rehash() {}
-}
```
neuro159: please avoid unnecessary changes

boesing: Sorry, its an IDE thing. 
However, as of `PSR-2`, there MUST be a UNIX LF instead of Windows `CRLF` at the end of a file.

PR #111, opened 2016-08-30
Fixing wrong stubs on \MongoCollection

villfa:
\MongoCollection::findOne() has a third parameter `options` ([check docs](http://www.php.net/manual/en/mongocollection.findone.php)).

### DIFF
--- a/mongo.php
+++ b/mongo.php
@@ -1146,9 +1146,10 @@ class MongoCollection {
 	 * @link http://www.php.net/manual/en/mongocollection.findone.php
 	 * @param array $query The fields for which to search.
 	 * @param array $fields Fields of the results to return.
+	 * @param array $options This parameter is an associative array of the form array("name" => <value>, ...).
 	 * @return array|null
 	 */
-    public function findOne(array $query = array(), array $fields = array()) {}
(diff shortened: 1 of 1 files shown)


neuro159:
May be you can add the contents of the "Currently supported options are:..." table? )

villfa:
Any recommendations about the format ?
In the same file I found 3 different formats :
- [`options` in \MongoCollection::insert()](https://github.com/JetBrains/phpstorm-stubs/blob/cdc93c6d5c42145d623ef35dc2efaff8c726e70b/mongo.php#L989) : HTML format (`dl`, `dd`, `dt`)
- [`options` in \MongoCollection::remove()](https://github.com/JetBrains/phpstorm-stubs/blob/cdc93c6d5c42145d623ef35dc2efaff8c726e70b/mongo.php#L1077) : HTML format (`ul`, `li`)
- [`options` in \MongoCollection::update()](https://github.com/JetBrains/phpstorm-stubs/blob/cdc93c6d5c42145d623ef35dc2efaff8c726e70b/mongo.php#L1046) : plain text

Which one should I choose ?

PR #112, opened 2016-08-30
DOMNamedNodeMap implements Traversable

maosmurf:
As of http://php.net/manual/en/class.domnamednodemap.php,  `DOMNamedNodeMap implements Traversable `

Verify by running
`
<?php assert(new DOMNamedNodeMap() instanceof Traversable);
`
Unfortunately, Intellij warns about DOMNamedNodeMap not implementing Traversable 

false error about domnamednodemap not being a traversable


neuro159:
Thanks!

PR #114, opened 2016-09-02
Added missing deprecated doctag for mysql_num_rows

King2500:
Missing `@deprecated 5.5` PhpDoc tag for `mysql_num_rows`

http://php.net/manual/en/function.mysql-num-rows.php


King2500:
@neuro159 I don't understand. Every other phpdoc for mysql functions has it above the description...

PR #115, opened 2016-09-02
added gRPC extension

iMega:


ged15:
any progress on this?

neuro159:

iMega:
Did I understand you correctly?

PR #117, opened 2016-09-08
WI-22743 stubs update according to the new extension point

artspb:
Stubs have to be divided into modules, "standard" is the first one. It has to be split in the future.


neuro159:
merged manually

PR #123, opened 2016-10-05
Add missing method maxTimeMS to mongo extension

cyrodiil:


neuro159:
Can you please avoid unnecessary formatting changes and provide a clean patch?

PR #125, opened 2016-10-12
Usage in other projects

felixfbecker:
Hi, I would open an issue, but since you don't allow issues I will propose this through a PR.
I would like to use these stubs as a dependency in the Open Source [PHP Language Server](https://github.com/felixfbecker/php-language-server). I have some problems with this:
- The language server is ISC licensed. Could you consider a more permissive license like MIT or ISC instead of Apache2?
- It would be nice to have this available as a composer package (just needs a composer.json file and a submission to packagist.org, no manual publishing needed, it will pull from git)
- Current version cannot be parsed by the PHP parser (can be checked through `php -l`):

```
PHP Parse error:  syntax error, unexpected '__halt_compiler' (T_HALT_COMPILER), expecting '(' in .\vendor\JetBrains\phpstorm-stubs\standard\_standard_manual.php on line 19

Parse error: syntax error, unexpected '__halt_compiler' (T_HALT_COMPILER), expecting '(' in .\vendor\JetBrains\phpstorm-stubs\standard\_standard_manual.php on line 19

Errors parsing .\vendor\JetBrains\phpstorm-stubs\standard\_standard_manual.php
```

`__halt_compiler` should really be defined as a keyword/language construct rather than a function, as it gets the `T_HALT_COMPILER` token and not the `T_FUNCTION` token.
- mongodb.php uses illegal nested namespaces
- It would be nice if you could enable issues for this repository

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,8 +13,8 @@ IDE needs them for completion, code inspection, type inference, doc popups, etc.
 TBD: Have a full copy of .git repo within IDE and add it as an external library "PHP Runtime" to the project. It should then be easilly updatable both way via normal git methods.
 
 ### License
-[Apache 2]
+[ISC]
 
 [PHPDOC]:https://github.com/phpDocumentor/fig-standards/blob/master/proposed/phpdoc.md
-[Apache 2]:https://www.apache.org/licenses/LICENSE-2.0
+[ISC]:https://opensource.org/licenses/ISC
 [Relevant open issues]:https://youtrack.jetbrains.com/issues/WI?q=%23Unresolved+Subsystem%3A+%7BPHP+lib+stubs%7D+order+by%3A+votes+


neuro159:
This is not an independent project at and its primary purpose is to improve PHP editing experience in JetBrains products. Hence the syntax nuances and more.

We'll discuss your proposals and get back to you.

As to "issues" - please use https://youtrack.jetbrains.com/issues/WI

felixfbecker:
I totally understand that. But as you get improvements from the community here, it would also be nice if you could "give back" to the community by making it available more easily to other projects 🙂

felixfbecker:
Hey @neuro159, any update? 🙂

neuro159:
We're all sorts of busy prepping our releases.
I'll amend the parsing problems asap, but redist/licensing parts will be
decided later.

Regards, Alex
+49 172 6938951

neuro159:
Added workaround for __halt_compiler.

felixfbecker:
Thanks Alex! I read up on Apache 2 a bit and it seems like I am allowed to redistribute your stubs with my LS (which is ISC). I can install the dependency with Composer directly from git. I will therefor close this PR - thanks for being so helpful :)

> You can choose to release the modified or derived products under different licenses, the unmodified parts of the software, however, must retain the Apache License, and you cannot name your modified version in any way that suggests that the final product is endorsed or created by the ASF.

> Copyleft licenses require the derivative works or modified versions of existing software to be released under the same license. The Apache License doesn’t have any such requirements. It’s a permissive license. It permits you to release the modified parts of the code under any license of your choice. However, you are required to release all the unmodified parts of the software under the same license (the Apache License).

Source: http://www.whitesourcesoftware.com/whitesource-blog/top-10-apache-license-questions-answered/?utm_source=quora&utm_medium=social&utm_term=blog-top-10-apache-license-questions-answered&utm_content=Is-the-MIT-license-compatible-with-the-Apache

PR #129, opened 2016-11-01
Update Collator constructor argument

BenMorel:
The documentation for [Collator::__construct()](http://php.net/manual/en/collator.construct.php) leaves no ambiguity, this should be `string $locale`.


neuro159:
Thanks!

PR #131, opened 2016-11-19
Fix invalid syntaxes

felixfbecker:
These several commits fix various PHP and PHPDoc syntax violations


neuro159:
Thank you!

felixfbecker:
Thanks for merging :)
One thing that still cannot get parsed is the function definition of `__halt_compiler`. The problem with it is that it doesn't parse as a function, but actually as a `T_HALT_COMPILER` token, that is invalid though because it isn't followed by `();`. Do you think PhpStorm could instead add this to the list of suggested _keywords_ (as it is one) and not define it as a function?

neuro159:
Sure. I have it as part of your original reuse request. It will be better
to file a separate per problem though, and into our main tracker)

PR #144, opened 2016-12-13
WI-33799 Fix memcache_increment() docs

Sannis:


Sannis:
@neuro159 fixed!

neuro159:
added manually after move & merge, thanks!

PR #145, opened 2016-12-13
WI-19857 Add stubs for judy extension

Sannis:


neuro159:

PR #147, opened 2016-12-14
Create mqseries.php

EagleEyeJohn:


neuro159:

..and fix that conditional?..

andrey-sokolov:
All required changes will be applied after merge

PR #156, opened 2017-01-13
igbinary support

Denitz:
Support for igbinary extension, see https://github.com/igbinary/igbinary


neuro159:

alexz707:
any progress on this PR ? Ping @Denitz ;)

Denitz:
@alexz707  Move to `/standard/igbinary/igbinary.php` or `/igbinary/igbinary.php` ? I don't see any other separate folders for non-standard extensions yet.

alexz707:
I'm not a maintainer of this repository but I think it seems legit if you create `/igbinary/igbinary.php`. It's not a standard lib so this is the way to go?! ;)

Denitz:
Ready.

PR #159, opened 2017-01-17
fann stub added to standart

Jnoack331:
issue: https://youtrack.jetbrains.com/issue/WI-34445


chernecov:
Any news on this PR?

neuro159:

PR #162, opened 2017-01-20
Added GNUPG Stub

trevrobwhite:
Added GNUPG Stub http://php.net/manual/en/book.gnupg.php


neuro159:

neuro159:
Thanks for contribution!

PR #165, opened 2017-01-26
Update unserialize stub for php 7.0

funivan:


funivan:
Issue reference: https://youtrack.jetbrains.com/issue/WI-30201

PR #167, opened 2017-01-30
Fix bool on failure returns in standards

addshore:
It was pointed out @ https://github.com/etsy/phan/issues/55#issuecomment-276126690 that many of these are wrong.

This adds the bool return type for methods that are marked as returning false on failure.


addshore:
It looks like there are at least 225 more of these low hanging fixes!

```
$ cat ./standard/*.php |grep @return |grep "on failure" |grep -v bool | wc -l
225
```

PR #201, opened 2017-04-19
Set default visibility for object class methods

b1rdex:
I often type `__` in class body to produce some base method like constructor.
If I use method from completion list then I have to provide visibility each time.
I think default `public` visibility should be there as these methods are public by default.

### DIFF
--- a/standard/_types.php
+++ b/standard/_types.php
@@ -221,7 +221,7 @@ class object {
    * @return void
    * @link http://php.net/manual/en/language.oop5.decon.php
    */
-  function __construct() {}
+  public function __construct() {}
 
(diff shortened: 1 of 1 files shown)


b1rdex:
As for `__sleep` signature change: docs states it must return array and if void returned, then it's serialized as `null` and notice provided.

http://php.net/manual/en/language.oop5.magic.php#language.oop5.magic.sleep

b1rdex:
I reverted __sleep signature change and will open another PR for it.

b1rdex:
@neuro159 will you do tests on your own or any instructions for me?

PR #204, opened 2017-04-24
update MongoDB\Driver\Server stub

angyvolin:
Updates `MongoDB\Driver\Server stub`.
ref: http://php.net/manual/en/class.mongodb-driver-server.php

### DIFF
--- a/mongodb/mongodb.php
+++ b/mongodb/mongodb.php
@@ -252,7 +252,7 @@ namespace MongoDB {}
              * Returns an array of tags describing this server in a replica set
              * @link http://php.net/manual/en/mongodb-driver-server.gettags.php
              * @throws InvalidArgumentException on argument parsing errors.
-             * @return array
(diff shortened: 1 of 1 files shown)


angyvolin:
[WI-36115 Incomplete MongoDB\Driver stubs](https://youtrack.jetbrains.com/issue/WI-36115)

mongodb/mongodb.php
```diff
@@ -244,6 +251,8 @@
             /**
              * Returns an array of tags describing this server in a replica set
              * @link http://php.net/manual/en/mongodb-driver-server.gettags.php
+             * @throws InvalidArgumentException on argument parsing errors.
+             * @return array
```
b1rdex: that's string[] probably

angyvolin: @b1rdex Thanks for your time.  I've updated description to expose more information about return value type. I guess, since this method returns array of key-value pairs, `string[]` type hint can lead to wrong expectations: one can expect "list" structure, while it's a "dictionary".

PR #211, opened 2017-05-10
Add RedisCluster

emirb:
Based on https://github.com/zgb7mtr/phpredis_cluster_phpdoc


emirb:
When can we expect to be in the app?

PR #217, opened 2017-05-21
Added ArgumentCountError class

Fleshgrinder:


Core/Core_c.php
```diff
@@ -476,6 +476,15 @@ class AssertionError extends Error {
 }
 
 /**
+ * ArgumentCountError is thrown when too few arguments are passed to a user
+ * defined routine.
+ *
+ * @since 7.1
+ * @see https://php.net/migration71.incompatible#migration71.incompatible.too-few-arguments-exception
+ */
+class ArgumentCountError extends Error {}
```
oshdev: According to the official docs http://php.net/manual/en/class.argumentcounterror.php ArgumentCountError extends \TypeError

PR #224, opened 2017-05-26
Fix PDOStatement::fetchColumn() return type

BenMorel:
The return type should be `mixed`, not `string`. In particular, it returns `false` when there are no more rows.

See: http://php.net/manual/en/pdostatement.fetchcolumn.php

### DIFF
--- a/PDO/PDO.php
+++ b/PDO/PDO.php
@@ -1134,8 +1134,8 @@ class PDOStatement implements Traversable {
 	 * no value is supplied, <b>PDOStatement::fetchColumn</b>
 	 * fetches the first column.
 	 * </p>
-	 * @return string <b>PDOStatement::fetchColumn</b> returns a single column
-	 * in the next row of a result set.
+	 * @return mixed Returns a single column from the next row of a result
+	 * set or FALSE if there are no more rows.
 	 * </p>
 	 * <p>
 	 * There is no way to return another column from the same row if you


andrey-sokolov:
It's better to use string|bool as a return type

BenMorel:
I agree, but it's also good to keep the stubs in sync with the PHP documentation, and they document it as `mixed`.

This would need to be submitted to the PHP docs as well in this case!

andrey-sokolov:
Unfortunately, PHP docs are often inaccurate and so there are many cases in our stubs where they don't follow the docs exactly. In this particular case we still believe that string|bool is better.

BenMorel:
Actually, the return type is not `string|bool`. It can return `null` for a nullable column.

And when using the mysqlnd driver, it can return native types such as integer as well. Try the following:

    $pdo = new PDO('mysql:host=localhost', 'root', '');
    $pdo->setAttribute(PDO::ATTR_EMULATE_PREPARES, false);
    $pdo->setAttribute(PDO::ATTR_STRINGIFY_FETCHES, false);
    var_dump($pdo->query('SELECT NULL')->fetchColumn()); // NULL
    var_dump($pdo->query('SELECT 1')->fetchColumn()); // int(1)

So it would be more something like `@return string|bool|null|int`, so I guess the `mixed` type in the PHP documentation makes sense here.

andrey-sokolov:
OK, let's make it mixed

PR #240, opened 2017-07-09
SimpleXMLElement accessor always returns a single SimpleXMLElement object

BenMorel:
Even when several nodes with the same name exist in the document, the first element is returned when using the magic accessor.

```php
$data = <<<EOF
<feed>
 <products>
  <product>1</product>
  <product>2</product>
  <product>3</product>
 </products>
</feed>
EOF;

$data = simplexml_load_string($data);

echo get_class($data->products->product); // SimpleXMLElement
echo $data->products->product; // 1
```

Therefore the return type of `__get()` should be `SimpleXMLElement`, not `SimpleXMLElement[]`.

### DIFF
--- a/SimpleXML/SimpleXML.php
+++ b/SimpleXML/SimpleXML.php
@@ -24,7 +24,7 @@ class SimpleXMLElement implements Traversable {
 	/**
      * Provides access to element's children
      * @param $name child name
-     * @return SimpleXMLElement[]
+     * @return SimpleXMLElement
      */
     function __get($name) {}
 


IMSoP:
It's a little more complex than that, as discussed on https://youtrack.jetbrains.com/issue/WI-15760

If you mark it as returning a single element, you'll just get an error reported for the opposite case: `$data->products[0]` should *also* be analysed as an instance of `SimpleXMLElement`, but with this patch will be flagged as an error.

I think a better approach is to mark the class as implementing ArrayAccess, allowing both types of access to be analysed successfully, which is what I have here: https://github.com/JetBrains/phpstorm-stubs/pull/245

BenMorel:
Indeed, your approach seems to be the best compromise for both use cases.
Closing in favour of #245.

PR #245, opened 2017-07-20
Add dummy ArrayAccess to SimpleXMLElement to fix WI-15760

IMSoP:
The actual array access methods are not available, but the internal class implements the equivalent C-level handlers, so that [0], ['foo'], etc are always available. 

This removes the need to mark any returns as SimpleXMLElement[], except for ->xpath(), which really does return a true array.

Issue was reported in 2012: https://youtrack.jetbrains.com/issue/WI-15760 This solution is not ideal, but the stub already contains a dummy __get(), so this seems a logical improvement.

### DIFF
--- a/SimpleXML/SimpleXML.php
+++ b/SimpleXML/SimpleXML.php
@@ -217,36 +217,36 @@ class SimpleXMLElement implements Traversable, ArrayAccess {
 
     /**
      * Class provides access to children by position, and attributes by name
-     * Stub only: method not callable directly.
+     * @access private Method not callable directly, stub exists for typehint only
      * @param string|int $offset
      * @return boolean true on success or false on failure.
      */
-    public function offsetExists ($offset) {}
+    private function offsetExists ($offset) {}
 
     /**
      * Class provides access to children by position, and attributes by name
-     * Stub only: method not callable directly.
+     * @access private Method not callable directly, stub exists for typehint only
      * @param string|int $offset
      * @return SimpleXMLElement Either a named attribute or an element from a list of children
      */
-    public function offsetGet ($offset) {}
+    private function offsetGet ($offset) {}
 
     /**
      * Class provides access to children by position, and attributes by name
-     * Stub only: method not callable directly.
+     * @access private Method not callable directly, stub exists for typehint only
      * @param string|int $offset
      * @param mixed $value
      * @return void
      */
-    public function offsetSet ($offset, $value) {}
+    private function offsetSet ($offset, $value) {}
 
     /**
      * Class provides access to children by position, and attributes by name
-     * Stub only: method not callable directly.
+     * @access private Method not callable directly, stub exists for typehint only
      * @param string|int $offset
      * @return void
      */
-    public function offsetUnset ($offset) {}
+    private function offsetUnset ($offset) {}
 }
 
 /**
--- a/SimpleXML/SimpleXML.php
+++ b/SimpleXML/SimpleXML.php
@@ -23,10 +23,11 @@ class SimpleXMLElement implements Traversable, ArrayAccess {
 
 	/**
      * Provides access to element's children
+     * @access private Method not callable directly, stub exists for typehint only
      * @param $name child name
      * @return SimpleXMLElement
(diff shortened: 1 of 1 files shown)


BenMorel:
I 👍 this approach. This is the best compromise at the moment.

The bottom line is, `__get()` **must** return `SimpleXMLElement`, not `SimpleXMLElement[]`.

The only drawback of this stub is that `offsetGet()` etc. methods will be autocompleted even though they're not actually present in `SimpleXMLElement`, but as noted above, this is already the case with `__get()`.

I thought that marking these methods as `@internal` would prevent this, but it turns out that it doesn't.

SimpleXML/SimpleXML.php
```diff
@@ -215,6 +215,38 @@ public function __toString () {}
 	 */
 	public function count () {}
 
+    /**
```
b1rdex: Methods below are not needed as IDE checks interface only and this can lead to errors in code completion.

BenMorel: Isn't `offsetGet()` required to infer the return type of `$element[0]`?

b1rdex: @BenMorel got it. Methods other then `__get` and `offsetGet` should be removed.

IMSoP: It turns out that deleting the methods doesn't help, because they are assumed to be present based on the interface. However, marking them as Private appears to have precisely the desired effect.

IMSoP:
Additional change to mark the new methods, along with `__get`, as private; from my testing, this should keep the magic property and array access, but prevent the non-existent methods showing in completion popups.

For reference, the code I've been using to confirm the changes have the desired effect is here: https://gist.github.com/IMSoP/dab21ce9d97980adbc22fc867e6e9fcf In PhpStorm 2017.2.1, with completion stubs pointed at this branch, all lines in that sample are considered valid, and the variables are detected as `SimpleXMLElement`.

b1rdex:
IMO PhpStorm should not take into account `private` methods for type gathering, that looks like bug for me. Making this methods `private` in real code makes them not accessible (e.g. throwing warnings), so return types are not usable.
I think that behavior is unreliable and shoud not be used.

IMSoP:
@b1rdex In theory, I agree, but it could also be considered a bug that internal class behaviours can't be signalled to the code analyser without writing stubs for methods that don't exist. It *should* be possible to mark the stub method as "virtual", or indicate the array access behaviour without listing the methods at all, but unless someone from JetBrains tells me that they've implemented such a feature, we are left compromising based on the observed behaviour.

Relying on the private method behaviour is certainly a hack, but the worst outcome is that this is "fixed" in a future version and the stub needs to be revised. It could be that by then an alternative approach is available.

andrey-sokolov:
@IMSoP @b1rdex Suggested change breaks on such example that was supported before
```
<?php
$xml = new SimpleXMLElement('<?xml version="1.0"?><request><error>222</error><error>333</error></request>');
echo $xml->error[1]->//invoke completion there
```
Looks like we need some solution coming from our codebase instead

SimpleXML/SimpleXML.php
```diff
@@ -23,10 +23,11 @@ class SimpleXMLElement implements Traversable {
 
 	/**
      * Provides access to element's children
+     * @access private Method not callable directly, stub exists for typehint only
      * @param $name child name
```
b1rdex: There must be `string` before `$name`, because phpdoc assumes `$name instanceof child::class`

b1rdex:
@andrey-sokolov can you please clarify what exactly is wrong with completion list? I created reproducible example one can run: https://gist.github.com/b1rdex/56bc533cd90b2d9880ef55799b09b425
So there is copied `SimpleXMLElement` class with added `2` to its name. Also, I changed `__get` and `offsetGet` methods to return `SimpleXMLElement2` instances.

Original completion list:
image

Class with proposed changes completion list:
image

I see no difference.

andrey-sokolov:
Works for me as well. Sorry for inconvenience.

PR #248, opened 2017-08-01
preg_split doc update

meridius:
added missing `FALSE` as a return value for `preg_split` if error occurs

http://php.net/manual/en/function.preg-split.php


pcre/pcre.php
```diff
@@ -360,7 +360,8 @@ function preg_filter ($pattern, $replacement, $subject, $limit = -1, &$count = n
  * If this flag is set, only non-empty pieces will be returned by
  * <b>preg_split</b>.
  * @return array an array containing substrings of <i>subject</i>
```
b1rdex: These should be `@return string[]|array|false …`.
`string[]` is common usage case (returns [string, …])
`array` is for `PREG_SPLIT_OFFSET_CAPTURE` (returns [array(string,int), …])
And `false` is for error case

meridius: nice catch, thank you

PR #258, opened 2017-08-17
stub for mapscript extension

rjbijl:
Added a stub file for the PHP mapscript extension. Generated by parsing the documentation at https://raw.githubusercontent.com/mapserver/docs/branch-7-0/en/mapscript/php/phpmapscript.txt

Biggest issue: missing values for the constants, because they are not defined in the docs


b1rdex:
Can you please extract constants values from https://github.com/mapserver/mapserver/blob/66309eebb7ba0dc70469efeb40f865a8e88fafbd/mapscript/tcl/mapscript_wrap.html ?

PR #267, opened 2017-09-05
Adding class stub for php_excel extension

rusakov92:
Here is the task in the issue tracker: https://youtrack.jetbrains.com/issue/WI-37929


andrey-sokolov:
Thank you for contribution. Since it's not a standard extension and not PECL extension it would be better to create a separate plugin with these stubs. Here is an example: https://github.com/artspb/phpstorm-library-plugin

PR #268, opened 2017-09-06
Fix callable typehints

b1rdex:
PhpStorm can't navigate to callables, because of wrong typehints on following functions:
- restore_error_handler
- set_exception_handler


andrey-sokolov:
set_error_handler - not restore_error handler

PR #269, opened 2017-09-11
Imagick::RESOURCETYPE_THREAD, ImageMagick ^6.7.8

utlime:
http://php.net/manual/en/imagick.constants.php#imagick.constants.resourcetype-thread


andrey-sokolov:
Thank you for contribution. Unfortunately in ImageMagick 7.0.7-1  Imagick::RESOURCETYPE_THREAD = 7 and there are some other constants that should added to our stub. I've created https://youtrack.jetbrains.com/issue/WI-38010. We are going to update this stub soon.

PR #271, opened 2017-09-12
WI-38016 fix redis del() and delete() docs

ShaneOH:
Most important thing here is that the `delete()` function currently incorrectly reports `void` return type. The signatures of both functions should be the same, and keys are of type `string`.

### DIFF
--- a/redis/Redis.php
+++ b/redis/Redis.php
@@ -353,7 +353,7 @@ class Redis
     /**
      * Remove specified keys.
      *
-     * @param   string|array    $key1 An array of keys, or an undefined number of parameters, each a key: key1 key2 key3 ... keyN
+     * @param   string|string[] $key1 An array of keys, or an undefined number of parameters, each a key: key1 key2 key3 ... keyN
      * @param   string          $key2 ...
      * @param   string          $key3 ...
      * @return  int             Number of keys deleted.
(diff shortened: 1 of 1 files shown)


redis/Redis.php
```diff
@@ -353,10 +353,10 @@ public function setnx( $key, $value ) {}
     /**
      * Remove specified keys.
      *
-     * @param   int|array   $key1 An array of keys, or an undefined number of parameters, each a key: key1 key2 key3 ... keyN
-     * @param   string      $key2 ...
-     * @param   string      $key3 ...
-     * @return int Number of keys deleted.
+     * @param   string|array    $key1 An array of keys, or an undefined number of parameters, each a key: key1 key2 key3 ... keyN
```
b1rdex: string|string[]

redis/Redis.php
```diff
@@ -372,9 +372,11 @@ public function del( $key1, $key2 = null, $key3 = null ) {}
 
     /**
      * @see del()
-     * @param $key1
-     * @param null $key2
-     * @param null $key3
+     * @param   string|array    $key1 An array of keys, or an undefined number of parameters, each a key: key1 key2 key3 ... keyN
```
b1rdex: string|string[]

redis/Redis.php
ShaneOH: @b1rdex Done!

PR #273, opened 2017-09-26
Enhancement: Define TOKEN_PARSE

localheinz:
This PR

* [x] defines `TOKEN_PARSE`

💁‍♂️ For reference, see:

* http://php.net/manual/en/function.token-get-all.php#refsect1-function.token-get-all-parameters
* https://3v4l.org/TKWdk

### DIFF
--- a/tokenizer/tokenizer.php
+++ b/tokenizer/tokenizer.php
@@ -42,6 +42,8 @@ function token_get_all ($source, $flags = 0) {}
  */
 function token_name ($token) {}
 
+define('TOKEN_PARSE', 1);
+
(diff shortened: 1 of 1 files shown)


andrey-sokolov:
Hey, Thank you for contribution. I think it's better to place this constant with other T_* constants, not between functions.

localheinz:
@andrey-sokolov 

The primary intention was to keep the constant definition as close as possible to the stub of `token_get_all()`. Let me know if you need any more amendments.

localheinz:
Thank you, @andrey-sokolov and @b1rdex!

PR #275, opened 2017-10-05
Flags parameter in Regex constructor is optional

cyrodiil:
The parameter $flags for the class MongoDB\BSON\Regex is optional:
http://php.net/manual/en/mongodb-bson-regex.construct.php

### DIFF
--- a/mongodb/mongodb.php
+++ b/mongodb/mongodb.php
@@ -1159,7 +1159,7 @@ namespace MongoDB {}
              * @param string $pattern
              * @param string $flags
              */
-            public function __construct($pattern, $flags)
+            public function __construct($pattern, $flags = "")
             {
             }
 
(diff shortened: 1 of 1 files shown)


b1rdex:
Note: it became optional from 1.2.0 version

mongodb/mongodb.php
```diff
@@ -1159,7 +1159,7 @@ class Regex implements Type
              * @param string $pattern
              * @param string $flags
```
andrey-sokolov: please add [optional] 
string $flags -> string #flags [optional]

cyrodiil: I added the [optional] flat, but I didn't see that anywhere else in the file...

andrey-sokolov: We use it in our stubs for optional parameters

PR #276, opened 2017-10-05
couchbase: update to 2.4.1

avsej:


couchbase/couchbase.php
```diff
@@ -3337,19 +3356,6 @@ function encode() {}
...
-         *
-         * Note that this method accessible only while Analytics support
-         * is experimental. It will be removed after GA release of this
-         * feature, and service location will be supplied in cluster config.
-         *
-         * @param string $hostname location of the analytics service.
-         *    Usually something like "localhost:8095/query/service"
-         * @return AnalyticsQuery
-         */
-        final public function hostname($hostname) {}
```
andrey-sokolov: Is it removed in new version?

avsej: yes. This was temporary method for experimental feature, that at that moment cannot be discovered through the cluster configuration API. So the user have to specify endpoint manually. New version of this service has been integrated with Couchbase orchestrator, and hostname is not needed anymore. The removed phpdoc states that it was experimental, so it does not really break API.

andrey-sokolov: Thank you for clarification.

PR #279, opened 2017-10-08
Add missing newrelic_* functions

p-h-p:
There are about 5 functions missing from the official documents (I have reported, will be fixed) and also from this stub.
Since there are no official docs, I didn't expand more on the docblock, it's better to wait for those then we can open another PR.

I have fetched the parameters like below, but I am unsure of the types since there is no source code available.
```
root@box:~# php --rf newrelic_curl_header_callback
Function [ <internal:newrelic> function newrelic_curl_header_callback ] {

  - Parameters [2] {
    Parameter #0 [ <required> $"curl_resource" ]
    Parameter #1 [ <required> $"header_data" ]
  }
}
```

### DIFF
--- a/newrelic/newrelic.php
+++ b/newrelic/newrelic.php
@@ -356,3 +356,37 @@ function newrelic_set_user_attributes($user, $account, $product) {}
  * @return boolean
  */
 function newrelic_start_transaction($appName, $license = null) {}
+
+/**
+ * @param $curl_resource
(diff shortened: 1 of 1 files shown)


andrey-sokolov:
We are going to remove this stub, because it's not a standard php extension. It would be great to create a separate plugin. Here is an example: https://github.com/artspb/phpstorm-library-plugin

PR #284, opened 2017-10-14
add Phalcon stubs

youbuwei:
The Phalcon stubs.


b1rdex:
Please take a look to https://github.com/phalcon/ide-stubs. There are up to date and better documented stubs.

andrey-sokolov:
Thank you for contribution! Phalcon is not a standard extension and not PECL extension it would be better to create a separate plugin with these stubs. Here is an example: https://github.com/artspb/phpstorm-library-plugin

PR #294, opened 2017-11-10
Add stub for SessionUpdateTimestampHandlerInterface

smuggli:
This interface is not documented on php.net, the only [documentation](https://github.com/php/php-src/blob/f9c3ee9ae8a41e2f47101c638491abd36ed09ce3/ext/session/session.c#L2730) is the code itself. There is also a [polyfill from Symfony](https://github.com/symfony/polyfill/blob/ff0c49221c6ef38caa463b182c17f2513f35c8ea/src/Php70/Resources/stubs/SessionUpdateTimestampHandlerInterface.php).

Is a documentation on php.net needed to get this merged?

### DIFF
--- a/session/SessionHandler.php
+++ b/session/SessionHandler.php
@@ -96,6 +96,40 @@ interface SessionHandlerInterface {
 }
 
 /**
+ * <b>SessionUpdateTimestampHandlerInterface</b> is an interface which
+ * defines a prototype for updating the life time of an existing session.
+ * In order to use the lazy_write option must be enabled and a custom session
+ * handler must implement this interface.
+ * @since 7.0.0
+ */
+interface SessionUpdateTimestampHandlerInterface {
+
+    /**
+     * Validate session id
+     * @param string $session_id The session id
+     * @return bool <p>
+     * Note this value is returned internally to PHP for processing.
+     * </p>
+     */
+    public function validateId($session_id);
+
+    /**
+     * Update timestamp of a session
+     * @param string $session_id The session id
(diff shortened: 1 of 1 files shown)


b1rdex:
I can't find any information on how to use it. Could you please give a link with example or documentation?

smuggli:
@b1rdex I think the best example out there is the implementation in Symfony (introduced in 3.4/4.0).

[updateTimestamp](https://github.com/symfony/symfony/blob/master/src/Symfony/Component/HttpFoundation/Session/Storage/Handler/MemcachedSessionHandler.php)

[validateId](https://github.com/symfony/symfony/blob/master/src/Symfony/Component/HttpFoundation/Session/Storage/Handler/AbstractSessionHandler.php)

I will try to write some documentation on php.net.

session/SessionHandler.php
```diff
@@ -95,6 +95,40 @@ public function read($session_id);
 	public function write($session_id, $session_data);
 }
 
+/**
+ * <b>SessionUpdateTimestampHandlerInterface</b> is an interface which
```
andrey-sokolov: did you use any resource to write this descrption?

smuggli: Actually I wrote it by myself but tried to make it similar to SessionHandlerInterface.

One "resource" has been a RFC which describes the behavior quite good but is not directly related to the implementation. Also I am using it by myself.

https://wiki.php.net/rfc/session-read_only-lazy_write

I had not much time lately so couldn't work on official documentation.

PR #300, opened 2017-11-30
Add stubs for PHP Google Protocol Buffers extension

Antoha-GS:


andrey-sokolov:
Thank you for contribution! Protobuf is not a standard extension and not PECL extension it would be better to create a separate plugin with these stubs. Here is an example: https://github.com/artspb/phpstorm-library-plugin

shmurakami:
Protobuf is now PECL extension. Can it be reopened and merged?

PR #302, opened 2017-12-05
Reference secure.php.net instead of php.net for all stubs.

umulmrum:
Change all links to the PHP documentation to use secure URLs.

I hope you consider accepting this PR despite no ticket is linked. To be honest, I was too lazy to create an account on the JetBrains tracker, file a ticket and reference it. This would have cost more time than the PR itself.


andrey-sokolov:
What is the goal of this change?
It takes much more time to reach secure.php.net and there should be a reason for waiting each time you want to check the documentation online

umulmrum:
The goal is to use HTTPS wherever possible. While privacy doesn't matter that much in this case, integrity of the received data does, which is only ensured by using HTTPS. I think that in 2017 HTTPS should be the default for all web-based communication and that NOT using it requires justification. 

Is it really that much slower for you? On my machine the difference is about 0.4s which I find negligible compared to the security gain.

andrey-sokolov:
@umulmrum this change is quite massive and the secure connection for doc links can be obtained in some other ways (web-browser extension or new option in PhpStorm for example).  I've created a ticket for this request https://youtrack.jetbrains.com/issue/WI-39291 .

PR #304, opened 2017-12-06
Add sodium_crypto_secretbox_keygen() function

roukmoute:


sodium/sodium.php
```diff
@@ -959,3 +959,13 @@ function sodium_crypto_scalarmult_base(
...
     return '';
 }
+
+/**
+ * Creates a random key
+ *
+ * It is equivalent to calling random_bytes() but improves code clarity and can
+ * prevent misuse by ensuring that the provided key length is always be correct.
+ *
+ * @return string
```
roukmoute: Is it useful for phpstorm engine or can I remove it?

roukmoute: Remove it

sodium/sodium.php
```diff
@@ -959,3 +959,11 @@ function sodium_crypto_scalarmult_base(
 
     return '';
 }
+
+/**
+ * Creates a random key
+ *
+ * It is equivalent to calling random_bytes() but improves code clarity and can
+ * prevent misuse by ensuring that the provided key length is always be correct.
+ */
```
andrey-sokolov: What is the source of descritpion? Please add link on php.net. Please add since tag

andrey-sokolov:
Thank you for contribution. Please add `@since` tag, and link to documentation.

sodium/sodium.php
roukmoute: Why would you like to add `@since` ?
All this file is only for `>=7.2`

andrey-sokolov: Thank you for clarification. This file is used for all PHP versions but `@since` is quite useful because you can check it in quick documentation, and we gonna use `@since` info for stubs filtering based on PHP Language level settings. That's why we want to have `@since` annotation for functions/classes/methods in our stubs.

roukmoute: So I must add it to all others function, isn't it?

andrey-sokolov: Let's start with this new function and patch others in a separate pull-request.

roukmoute: 🆗

roukmoute: I've just seen : `The effects of this tag are not yet fully implemented in PhpDocumentor2.`

andrey-sokolov: We have plans to implement full support and that's why we want to have our stubs ready for this.

roukmoute:
So are you interested if I do : https://github.com/JetBrains/phpstorm-stubs/pull/304#discussion_r156090671?

PR #306, opened 2017-12-12
Update Redis

Denitz:
`Redis::exists()` actually returns INT, current Readme.markdown on https://github.com/phpredis/phpredis is incorrect.


b1rdex:
How do you know it? Source code 
https://github.com/phpredis/phpredis/blob/develop/redis.c#L1163 shows it's boolean and never changed.

Denitz:
Argh, sorry, I used `develop` version, 3.1.4 stable is fine.

PR #307, opened 2017-12-14
Add missing imagick constants

JWZH:
This pull request adds some predefined constants only available in (ImageMagick < 7)

* imagick::ALPHACHANNEL_RESET (compiled against ImageMagick version 6.3.8 or higher)
* imagick::ALPHACHANNEL_COPY ( compiled against ImageMagick version 6.4.6 or higher)
* imagick::ALPHACHANNEL_OPAQUE (compiled against ImageMagick version 6.4.6 or higher)

See http://php.net/manual/en/imagick.constants.php for more details

These constants do not exist in the curernt repo and I guess it's because they are not available in ImageMagick >= 7. However, I think most companies are still using ImageMagick 6.


imagick/imagick.php
```diff
@@ -491,6 +491,9 @@ class Imagick implements Iterator, Countable {
 	const ALPHACHANNEL_TRANSPARENT = 15;
 	const ALPHACHANNEL_ASSOCIATE = 2;
 	const ALPHACHANNEL_DISSOCIATE = 7;
+    const ALPHACHANNEL_RESET = 7;
```
andrey-sokolov: We use tabs for in our stubs, please, amend indent accordingly.

PR #319, opened 2018-01-08
add yaf

wxxiong6:
http://php.net/manual/en/book.yaf.php


andrey-sokolov:
Thank you for contribution! Do you plan to fix functions/methods signatures with type annotations?

wxxiong6:
yes, I will

wxxiong6:
Has added the type and annotation.

PR #328, opened 2018-02-01
Optionality of params fix in methods signatures

alexglue:
trouble case:

```
class MyFooMemcache extends \Memcache {

    public function get($key, &$flags){} //this is actual stub's signature
}
```
turns into warning:

```
Missing argument 2 for MyFooMemcache::get()
```


andrey-sokolov:
@alexglue Call of the method with 1 parameter should be highlighted with warning, cause of [optional] annotation in method doc

PR #336, opened 2018-02-18
Correction of return types for "MultipleIterator"

maestroprog:


SPL/SPL_c1.php
```diff
@@ -2206,7 +2206,7 @@ public function key () {}
         /**
          * Gets the registered iterator instances
          * @link http://php.net/manual/en/multipleiterator.current.php
-         * @return void An array of all registered iterator instances,
+         * @return array An array of all registered iterator instances,
          * or false if no sub iterator is attached.
```
b1rdex: Please add `|false` to type definition. Also please add `@throws \RuntimeException`. And please change comment of return value to make it clear that it returns `array of current values of attached iterators` rather than array of attached iterators.

SPL/SPL_c1.php
```diff
@@ -2164,15 +2164,15 @@ public function detachIterator (Iterator $iterator) {}
          * @param Iterator $iterator <p>
          * The iterator to check.
          * </p>
-	 * @return void true on success or false on failure.
+	     * @return bool true on success or false on failure.
```
Majkl578: This indent seems wrong.

PR #339, opened 2018-02-23
Add using the namespace Yaf

wxxiong6:
Add using the namespace Yaf

### DIFF
--- a/yaf/yaf_namespace.php
+++ b/yaf/yaf_namespace.php
@@ -1,0 +1,3402 @@
+<?php 
+namespace
+{
+    
+    define('YAF_VERSION', '3.0.6', true);
+    define('YAF_ENVIRON', 'product', true);
+    define('YAF_ERR_STARTUP_FAILED', 512, true);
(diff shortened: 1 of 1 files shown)


andrey-sokolov:
Please provide a link on namespace Yaf extension

wxxiong6:
[http://php.net/manual/en/yaf.configuration.php](http://php.net/manual/en/yaf.configuration.php)
yaf.use_namespace integer 
Only works as of PHP 5.3, if this value is On, All classes of Yaf will named in namespace style. 
For example: 
`Yaf_Route_Rewrite => \Yaf\Route\Rewrite
Yaf_Request_Http  => \Yaf\Request\Http`
        
There is a exception, that is some classes like Yaf_Controller_Abstract. The last component is a keyword of PHP, could not be used as a class name, so for such classes: 
`Yaf_Controller_Abstract => \Yaf\Controller_Abstract
Yaf_Route_Static => \Yaf\Route_Static`

PR #350, opened 2018-03-16
Add stubs for ioncube 10.2.0 loader API

borgou:
Add stubs for ioncube 10.2.0 loader API.
The full user guide can be found here [USER-GUIDE](https://www.ioncube.com/sa/USER-GUIDE.pdf)

### DIFF
--- a/ioncube/ioncube.php
+++ b/ioncube/ioncube.php
@@ -1,0 +1,204 @@
+<?php
+
+/**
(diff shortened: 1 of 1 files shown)


andrey-sokolov:
Thank you for contribution! Ioncube is not a standard extension and not PECL extension, it would be better to create a separate plugin with these stubs. Here is an example: https://github.com/artspb/phpstorm-library-plugin

PR #351, opened 2018-03-18
Changing "<?" to "<?php" in apc/apc.php

jaydiablo:
The lack of "php" here causes some configurations to interpret the file as non-php.

### DIFF
--- a/apc/apc.php
+++ b/apc/apc.php
@@ -1,4 +1,4 @@
-<?
+<?php
 
 /**
  * Stubs for APC 3.1.4


andrey-sokolov:
Thank you for contribution. Please, elaborate on your use-case - what configurations more specifically in what use-case.

jaydiablo:
@andrey-sokolov I'm actually experimenting with using this project to provide stubs for use with phpstan when optional extensions aren't loaded.  Typically short tags are disabled on systems that I use, so when PHP includes this particular file, it doesn't interpret it as PHP code.

PR #352, opened 2018-03-24
Update return type of `newrelic_disable_autorum`

jderusse:
When using `function foo(): bool {return newrelic_disable_autorum();}`, PhpStorm report an error because true is not a boolean.

### DIFF
--- a/newrelic/newrelic.php
+++ b/newrelic/newrelic.php
@@ -102,7 +102,7 @@ function newrelic_custom_metric($metricName, $value) {}
  *
  * @link https://docs.newrelic.com/docs/agents/php-agent/configuration/php-agent-api#api-rum-disable
  *
- * @return true
+ * @return boolean
  */
 function newrelic_disable_autorum() {}
 


jaydiablo:
I don't think this matches the documentation though:

https://docs.newrelic.com/docs/agents/php-agent/php-agent-api/newrelic_disable_autorum

>Always returns true.

jderusse:
`true` is not a type, it's a value. Here `@return` expect a type.

jaydiablo:
Yes, but in this context it's a php docblock type, not a php typehint, and it's common when documenting types in php docblocks to use just `true` or `false` if the function/method/param wouldn't return/accept `true` **or** `false`, but just one or the other.

http://docs.phpdoc.org/references/phpdoc/types.html
https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc.md#appendix-a-types

Many static analyzers (and I assume PHP Storm as well) will make this distinction, which is why it's complaining about the `bool` typehint on that function.

You'll find that this is one of many cases where Jetbrains has (correctly, IMHO) marked either `true` or `false` instead of `bool` when the type can't be both.

https://github.com/JetBrains/phpstorm-stubs/blob/dc20836a546ce8fcb939d72003d8a409924432e9/apache/apache.php#L25
https://github.com/JetBrains/phpstorm-stubs/blob/6bac808b849f681cdc1ecef64fce388dc2ccfab4/standard/standard_5.php#L402
https://github.com/JetBrains/phpstorm-stubs/blob/78066e1719550ae7925155de6e1c56fd11fbff1a/mbstring/mbstring.php#L104

This isn't my library, so not up to me (I don't know their standards for these stubs), I just wanted to mention that how it's documented now is "correct".

b1rdex:
This is bug in `PHP | Type compatibility | Strict type checking rules violation` inspection IMHO and should be reported to youtrack.jetbrains.com.

artspb:
>PhpStorm report an error because true is not a boolean.

@jderusse Could you please post the error message here?

PR #354, opened 2018-04-03
Amqp stubs updated

martinezdelariva:


amqp/amqp.php
```diff
@@ -1,6 +1,7 @@
 <?php
 /**
  * Stubs for AMQP
+ * @since 2018-04-03T10:00:41+00:00
```
b1rdex: IMHO this should be removed, because `@since` is used to indicate version of php/module things was introduced.

martinezdelariva: Thanks. I misunderstood the `@since` tag

PR #358, opened 2018-04-17
Fix  \DateTimeImmutable methods definitions

Grzesie2k:
Definitions of \DateTimeImmutable methods contains invalid return types.


jaydiablo:
I think being explicit with the object returned here (changing from `static` to `\DateTimeImmutable`) is incorrect, because of code like this:

```php
<?php

class myDate extends DateTimeImmutable {};

$date = new myDate();

$interval = new DateInterval('P2Y4DT6H8M');

$newDate = $date->add($interval);

var_dump($newDate);
```

https://3v4l.org/NQbCZ

Notice how that returns a `myDate` object.

PR #359, opened 2018-04-19
Fix typos caused by case-insensitive method name

mingrammer:


mingrammer:
The hash method name from https://github.com/phpredis/phpredis#hashes

PR #365, opened 2018-04-20
Added constant "\PDO::MYSQL_ATTR_SSL_VERIFY_SERVER_CERT"

Geolim4:
- As per https://bugs.php.net/bug.php?id=71003
- Implemented at http://git.php.net/?p=php-src.git;a=blobdiff;f=ext/pdo_mysql/pdo_mysql.c;h=99fb6d3b77154c2f9928dd90e4f3c4b4f76d9611;hp=32b85ba54736e4f5bc79ceb5a486242d72da511a;hb=247ce052cd0fc7d0d8ea1a0e7ea2075e9601766a;hpb=b224e7426713befd7449117afd68355231f9077e
- As of 7.0.18 and 7.1.4 (see http://php.net/ChangeLog-7.php)
- PR at php/php-src#1972


PDO/PDO.php
```diff
@@ -710,6 +710,18 @@ class PDO  {
...
 	const MYSQL_ATTR_MULTI_STATEMENTS = 1015;
 
+	/**
+	 * <p>
+	 * Disables SSL peer verification when set to FALSE.
+	 * </p>
+	 * <p>
+	 * Note, this constant is not yet documented on the php.net documentation.
+	 * </p>
+	 * @since 7.0.18 and 7.1.4
```
b1rdex: This should be splitted to two `@since` tags

b1rdex:
Is there any request to add this to php.net documentation?

PDO/PDO.php
Geolim4: I can split them but what about this then ?
https://github.com/JetBrains/phpstorm-stubs/blob/master/PDO/PDO.php#L708

Geolim4:
I'm not sure, in my case I did no request at all.

PDO/PDO.php
b1rdex: dunno 🤷‍♂️  If this ever become parsed, then it's better to have them splitted

b1rdex: Also, take a look at `@since` definition at http://docs.phpdoc.org/references/phpdoc/tags/since.html

Geolim4: Okay, I'll fix mine + another PR to fix others, @andrey-sokolov what's your mind on this ?

andrey-sokolov: For new contributions it's better ti split such tags

Geolim4:
I can fix it tonight then, thanks :)

Le mar. 24 avr. 2018 à 15:04, andrey-sokolov <notifications@github.com> a
écrit :

PDO/PDO.php
```diff
@@ -710,6 +710,18 @@ class PDO  {
 	 */
 	const MYSQL_ATTR_MULTI_STATEMENTS = 1015;
 
+	/**
+	 * <p>
+	 * Disables SSL peer verification when set to FALSE.
+	 * </p>
+	 * <p>
+	 * Note, this constant is not yet documented on the php.net documentation.
```
andrey-sokolov: I think, this note is redundant and it's better to remove it

Geolim4: Okay, i thought that it was a good idea for those that have to maintain the project.

Geolim4: @andrey-sokolov It's fixed !

Geolim4: @andrey-sokolov up :)

Geolim4: @andrey-sokolov Up !

PR #367, opened 2018-04-28
Use https for @link annotations

BackEndTea:
A lot of times when looking at the stubs for reference i want to check the documentation of the official php site for more in-depth info of a function. I prefer to do that over https.


b1rdex:
See #302

BackEndTea:
Thank you @andrey-sokolov

PR #368, opened 2018-05-02
Update xhprof.php

AxiosCros:


xhprof/xhprof.php
```diff
@@ -17,7 +17,7 @@
  *
  * @return null
  */
-function xhprof_enable( $flags = 0, array $options )
```
andrey-sokolov: Please add [optional] to the doc for $options parameter

PR #373, opened 2018-05-30
 Create libpostal.php

rg-viza:
stub for php-postal https://github.com/openvenues/php-postal

Please let me know if I need to change anything...


andrey-sokolov:
Thank you for contribution! libpostal is not a standard extension and not PECL extension, it would be better to create a separate plugin with these stubs. Here is an example: https://github.com/artspb/phpstorm-library-plugin

PR #379, opened 2018-06-22
sodium_memzero() - removed invalid param $length

codemasher:
http://php.net/manual/function.sodium-memzero.php


andrey-sokolov:
Thanks for noting this problem!

Please note, that we use Travis continuous integration and this PR have failed checks.
In this case reflection based test fails cause for some reason `ReflectionFunction("sodium_memzero")->getParameters()` returns two parameters but really it accepts only one. 

We need to mute this problem till we find a better way to test stubs signatures or till reflection bug will be fixed. Please, add `sodium_memzero` to `tests/mutedProblems.json` with problem "parameter mismatch" to fix failed test.


Please, remove commit 59d684e with .gitignore, it's better to solve one problem per PR.

codemasher:
ahrg. i should learn to create branches instead of pushing to master... i'm gonna do this again in a clean fork. brb.

PR #381, opened 2018-06-27
Added missing PDO::SQLSRV_* constants

morozov:
These constants are not documented in the PHP [manual](http://php.net/manual/en/ref.pdo-sqlsrv.php#pdo-sqlsrv.constants), therefore the information on their individual purposes and the versions they were first introduced is unavailable.

The documentation on these constants can be found [here](https://docs.microsoft.com/en-us/sql/connect/php/constants-microsoft-drivers-for-php-for-sql-server).


morozov:
@andrey-sokolov could you review this any time soon?

PR #382, opened 2018-06-27
Add stubs for SQL Anywhere sasql extension

Majkl578:
http://dcx.sap.com/index.html#sa160/en/dbprogramming/php-api.html
http://dcx.sap.com/sa160/en/pdf/dbprogramming16.pdf

cc @morozov

### DIFF
--- a/sasql/sasql.php
+++ b/sasql/sasql.php
@@ -1,0 +1,687 @@
+<?php
+/**
+ * Stubs for sasql extension
+ * Taken from official SAP documenation
+ * @link     http://dcx.sap.com/index.html#sa160/en/dbprogramming/php-api.html
+ * @link     http://dcx.sap.com/sa160/en/pdf/dbprogramming16.pdf
+ * @author   Michael Moravec
+ */
+
+const SASQL_NUM = 1;
+const SASQL_ASSOC = 2;
+const SASQL_BOTH = 3;
+
+const SASQL_D_INPUT = 1;
+const SASQL_D_OUTPUT = 2;
+const SASQL_INPUT_OUTPUT = 3;
+
+const SASQL_USE_RESULT = 0;
+const SASQL_STORE_RESULT = 1;
+
+/**
+ * Returns the number of rows affected by the last SQL statement. This function is typically used for
+ * INSERT, UPDATE, or DELETE statements. For SELECT statements, use the sasql_num_rows function.
+ *
+ * @param resource $conn The connection resource returned by a connect function.
+ *
+ * @return int The number of rows affected
+ */
+function sasql_affected_rows($conn) {}
+
+/**
+ * Ends a transaction on the SQL Anywhere database and makes any changes made during the transaction
+ * permanent. Useful only when the auto_commit option is Off.
+ *
+ * @param resource $conn The connection resource returned by a connect function.
+ *
+ * @return bool TRUE on success or FALSE on failure.
+ */
+function sasql_commit($conn) {}
+
+/**
+ * Closes a previously opened database connection.
+ *
+ * @param resource $conn The connection resource returned by a connect function.
+ *
+ * @return bool TRUE on success or FALSE on failure.
+ */
+function sasql_close($conn) {}
+
+/**
+ * Establishes a connection to a SQL Anywhere database.
+ *
+ * @param string $conn_str A connection string as recognized by SQL Anywhere.
+ *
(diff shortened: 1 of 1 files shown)


sasql/sasql.php
```diff
@@ -0,0 +1,685 @@
...
+ */
+
+const SASQL_ASSOC = 0;
+const SASQL_BOTH = 0;
+const SASQL_D_INPUT = 0;
+const SASQL_D_OUTPUT = 0;
+const SASQL_INPUT_OUTPUT = 0;
+const SASQL_NUM = 0;
+const SASQL_STORE_RESULT = 0;
+const SASQL_USE_RESULT = 0;
```
morozov: I believe we need to extract the actual values from the extension distribution binaries. Otherwise, it may lead to errors like "duplicate array keys" when constants are used as array keys.

Majkl578: Yes, I don't have access to any SA installation/extension though. :/ So at least better than udefined constants ATM.

morozov: Let me see if I can extract them during the day. This is where I currently am:
```
↪ ldd php-5.6.0_sqlanywhere.so
        statically linked

↪ php -dextension=$PWD/php-5.6.0_sqlanywhere.so
Warning: PHP Startup: The SQLAnywhere client libraries could not be loaded. Please ensure that libdbcapi.so can be found in your LD_LIBRARY_PATH environment variable. in Unknown on line 0
```
Going to try the Windows DLL later.

sasql/sasql.php
```diff
@@ -0,0 +1,685 @@
...
+ * @return TRUE if binding the variable was successful or FALSE otherwise.
+ */
+function sasql_stmt_bind_param_ex($stmt, $param_number, &$var, $type, $is_null = null, $direction = null) {}
+
+/**
+ * Binds one or more PHP variables to result columns of a statement that was executed, and returns a result set.
+ *
+ * @param resource $stmt A statement resource that was executed by sasql_stmt_execute.
+ * @param mixed ...$vars References to PHP variables that will be bound to result set columns returned by the
+                         sasql_stmt_fetch.
```
morozov: A `*` is missing in the beginning.

sasql/sasql.php
morozov: From the extension [source](https://archive.sap.com/documents/docs/DOC-40537):
```c
REGISTER_LONG_CONSTANT("SASQL_D_INPUT", 1, CONST_CS | CONST_PERSISTENT);
REGISTER_LONG_CONSTANT("SASQL_D_OUTPUT", 2, CONST_CS | CONST_PERSISTENT);
REGISTER_LONG_CONSTANT("SASQL_D_INPUT_OUTPUT", 3, CONST_CS | CONST_PERSISTENT);

REGISTER_LONG_CONSTANT("SASQL_USE_RESULT", 0, CONST_CS | CONST_PERSISTENT);
REGISTER_LONG_CONSTANT("SASQL_STORE_RESULT", 1, CONST_CS | CONST_PERSISTENT);

REGISTER_LONG_CONSTANT("SASQL_NUM", 1, CONST_CS | CONST_PERSISTENT);
REGISTER_LONG_CONSTANT("SASQL_ASSOC", 2, CONST_CS | CONST_PERSISTENT);
REGISTER_LONG_CONSTANT("SASQL_BOTH", 3, CONST_CS | CONST_PERSISTENT);
```

Majkl578: Cool, updated!

andrey-sokolov:
Thank you for contribution! sasql is not a standard extension and not PECL extension, it would be better to create a separate plugin with these stubs. Here is an example: https://github.com/artspb/phpstorm-library-plugin

Majkl578:
Hmm I think I am not going to do that unfortunately, I am not interested in maintaining a plugin for an extension I don't use. It's just that Doctrine uses it and PhpStorm is unable to recognize it's functions.

Majkl578:
@andrey-sokolov Should probably noted that stubs already include multiple non-PECL/non-standard proprietary extensions:
 * suhosin
 * sybase (precedesor of SASQL)
 * newrelic
 * mapscript
 * ZendCache
 * ZendDebugger
 * ZendUtils

So your decision doesn't seem to be objective.

neuro159:
These were there years before opening public contribution and will remain for backward compatibility reasons.

Popularity matters the most now. Do you have some usage statistic for this extension?

PR #383, opened 2018-07-02
Fix OCI-Lob and OCI-Collection class names

Majkl578:
Attempt to fix incorrect names of OCI classes.
Correct names are `OCI-Lob` and `OCI-Collection`, not with underscores:
http://php.net/manual/en/class.OCI-Lob.php
http://php.net/manual/en/class.OCI-Collection.php
Since these names are not valid class names, I haven't found any better way than trick with class_alias().

Is it somehow possible to prevent `__Internal*` from appearing in autocomplete?

References:
https://github.com/doctrine/dbal/pull/3025#discussion_r198236056
https://github.com/phpstan/phpstan/issues/1123
https://github.com/phpstan/phpstan/commit/6149780143f59e82b048ec995f88a88eede7bd53

### DIFF
--- a/oci8/oci8.php
+++ b/oci8/oci8.php
@@ -2,11 +2,14 @@
 
 // Start of oci8 v.2.0.7
 
+class_alias('__InternalOCILob', 'OCI-Lob');
+class_alias('__InternalOCICollection', 'OCI-Collection');
+
 /**
  * OCI8 LOB functionality for large binary (BLOB) and character (CLOB) objects.
  * @link http://php.net/manual/en/class.OCI-Lob.php
  */
-class OCI_Lob  {
+class __InternalOCILob {
 
 	/**
 	 * (PHP 5, PECL OCI8 &gt;= 1.1.0)<br/>
@@ -276,7 +279,7 @@ class OCI_Lob  {
  * OCI8 Collection functionality.
  * @link http://php.net/manual/en/class.OCI-Collection.php
  */
-class OCI_Collection  {
+class __InternalOCICollection {
 
 	/**
 	 * (PHP 5, PECL OCI8 &gt;= 1.1.0)<br/>


b1rdex:
> Is it somehow possible to prevent __Internal* from appearing in autocomplete?

AFAIK no official way. But class can be named with non-ascii… so class named using emojis wouldn't be in list 🙄 😁

andrey-sokolov:
Class identifiers with **dash** are not supported in PHP and PhpStorm parser. That's why we use the names with underscores. Besides, all real-world examples use underscored names as well. Based on this reasoning I'm inclined to decline your PR

Majkl578:
> Based on this reasoning

What reasoning? This breaks real-world code - PHPStan for example.

MaXal:
@Majkl578 Class names with **dash** are not valid in PHP (http://php.net/manual/en/language.oop5.basic.php):
*The class name can be any valid label, provided it is not a PHP reserved word. A valid class name starts with a letter or underscore, followed by any number of letters, numbers, or underscores. As a regular expression, it would be expressed thus: ^[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*$.*
See: https://3v4l.org/AHso6 for results of PHP execution.

According to PSR-5 draft gramatic class names with **dash** can't be used to describe type as well: https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc.md#appendix-a-types

Majkl578:
> Class names with dash are not valid in PHP 

I am very well aware of that, although this only applies to declaring them.

Extensions could declare classes with invalid characters (as you can see here) and you can also instantiate such classes (just not directly by `new ClassName()`: https://3v4l.org/YbOeT

morozov:
@andrey-sokolov your reasoning above is valid if applied to writing PHP code but isn't really valid if applied to documenting _existing_ PHP extensions which these stubs are about. As long as those symbols exist and can be used in the code, they should be stubbed as is.

Similarly to the example above, this is valid PHP code too:
```php
$class = 'OCI-Lob';

if ($value instanceof $class) {
    // we've got a LOB
}
```

MaXal:
Folks, we understand that these class names with dashes are real and exists. But still, they are not compatible with PHP/PSR-5 and the fix produces a lot of questions and possible issues for the dependent projects and user code:
* People start to use `/** @var OCI-Lob */` in their code. That is not valid PHPDoc so any time PHPDocumentator or any other tool that parses PHPDocs may break on this code.
* People already use `/** @var OCI_Lob */` and the fix will break type inference and completion for them.
* Classes in question don't have a private constructor so by mistake people may complete `new OCI-Lob` and PhpStorm parser, as well as PHP parser, will go crazy.

The benifits on the other hand are not at all clear for me. Could you please provide the specific problem with details of what exactly goes wrong in PHPStan with the current class names? May be the problem should be resolved on PHPStan side.

morozov:
> they are not compatible with PHP/PSR-5

> People start to use `/** @var OCI-Lob */` in their code. That is not valid PHPDoc so any time PHPDocumentator or any other tool that parses PHPDocs may break on this code.

Could you please refer to some document explaining why this PHPDoc is invalid? I was able to generate documentation from the doc-block like this:

screenshot from 2018-08-06 18-02-41

> People already use `/** @var OCI_Lob */` and the fix will break type inference and completion for them.

BC could be preserved by keeping the original class name unchanged and adding an alias. @Majkl578 Why did you need to rename the actual class name from `OCI_Lob` to `__InternalOCILob`?

> Classes in question don't have a private constructor so by mistake people may complete `new OCI-Lob` and PhpStorm parser, as well as PHP parser, will go crazy.

That's true. However, PhpStorm could understand that despite the fact that the class `What-Ever` exists, it cannot be used as a symbol but can be used via a variable. This is what PHP is.

> The benifits on the other hand are not at all clear for me. Could you please provide the specific problem with details of what exactly goes wrong in PHPStan with the current class names? May be the problem should be resolved on PHPStan side.

Please see the list of references in the description. The bottom line is, in the absence of the actual extension at the run-time, PHPStan tries to use PhpStorm stubs. The stubs should be the same as the extension since they are only loaded if the extension is not loaded. If they are inconsisten, then the analysis will pass or fail depending on whether the actual or stubbed API is loaded at the run time.

Majkl578:
> Why did you need to rename the actual class name from OCI_Lob to __InternalOCILob?

Just to clearly indicate that _OCI_Lob_ and _OCI_Collection_ are not correct names and don't confuse everyone.

Majkl578:
> since those names goes against the standard

PSRs are not PHP standards, only recommendation by/for some user-land ecosystems.
Also PSR-5 is not even approved.
If you want a standard, [here is PHP language specification](https://github.com/php/php-langspec/blob/63375439ff7e8a1be93d0ef7d3c3b4833d5e35a0/spec/09-lexical-structure.md#names), but it's rather a syntax for declarations.

> Can you use alias on your side so that OCI_Lob and OCI-Lob are treated as equal?

Not for portable code, `OCI-Lob` and `OCI_Lob` are two different classes and generic library can't declare aliases like that (i.e. using `class_alias` in global scope).

morozov:
> Can you use alias on your side so that OCI_Lob and OCI-Lob are treated as equal?

@MaXal as I explained above, here's the problem with using an alias: if the `oci8` extension is loaded, PHPStan uses the extension reflection to load the declared symbols and _doesn't load the stub_ because:

1. Why load the stub if the extension itself is loaded?
2. The symbol the stub declares will conflict with the symbols already declared by the extension.

In this case, if I use `/** @var OCI_Lob */` in the code, PHPStan won't be able to find its declaration and will report an error.

At the same time, if I use `/** @var OCI-Lob */` and the extension is _not_ loaded, PHPStan also won't be able to find its declaration and will report an error.

The key here is to have the stubs _consistent_ with the actual extension so that the result doesn't depend on whether the actual or stubbed extension is used as the source of declarations.

morozov:
@MaXal this inconsistency leads to a new series of problems:

1. In order to make the build pass, we have to whitelist this issue (see https://github.com/doctrine/dbal/commit/8b8af0c306ca4b32b9d8096c146bf92995cca11b#diff-0847156bd64b0de0f4d91da2394ec0e8R46).
2. Given that this error is reproducible inconsistently, we have to [disable](https://github.com/doctrine/dbal/commit/df4343721d9981e58afd15eeaf18c885a2399766#diff-0847156bd64b0de0f4d91da2394ec0e8R7) `reportUnmatchedIgnoredErrors` in PHPStan. Otherwise, PHPStan would report if a whitelisted issue didn't happen and should be removed from the whitelist.
3. In practice, it means that once a whitelisted issue is fixed, it still may remain in the whitelist until manually found and removed.

As you can see, it all stems from the inconsistency between the stubbed and actual symbols.

PR #385, opened 2018-07-12
Updated return types for some of the ob_ functions

maksimovic:
Some of the inspection plugins (e.g. EA Extended) will emit a warning if the type casting is not necessary, which will generate false warning in cases for the signatures updated here, like this one:

screen shot 2018-07-12 at 3 47 30 pm

Function descriptions already contain "... or false if ...", just the type itself does not include it.

~Also, `false` is a valid return type to use, but I've noticed that just `bool` is used in other similar cases, hence `string|bool` here, too. I'd prefer `string|false` though, because `true` will never be returned.~

Edit: amended the commit to use `string|false`.

### DIFF
--- a/standard/standard_8.php
+++ b/standard/standard_8.php
@@ -268,7 +268,7 @@ function ob_end_clean () {}
 /**
  * Flush the output buffer, return it as a string and turn off output buffering
  * @link http://php.net/manual/en/function.ob-get-flush.php
- * @return string the output buffer or false if no buffering is active.
+ * @return string|false the output buffer or false if no buffering is active.
  * @since 4.3.0
  * @since 5.0
  */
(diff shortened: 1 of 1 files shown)


Majkl578:
Although I'm not a fan of `|false` unit type, I think it's the correct one for internal PHP functions (that return `false`on error and never return `true` otherwise). Using `|bool` breaks static analysis (PHPStan) as both `true` and `false` would have to be covered.

maksimovic:
@Majkl578 Agreed. Would you prefer to have these updated as `string|false` then (not sure if the comment was meant to be a requirement or just a note)?

b1rdex:
+1 for `|false`

Majkl578:
Note that using `|true`/`|false` is [explicitly allowed by phpDoc](https://docs.phpdoc.org/guides/types.html#keywords). :blush:

PR #389, opened 2018-07-19
Fix return type and description of stream_copy_to_stream

hollodotme:
* Official documentation says it can return int or false on failure
* See http://php.net/manual/en/function.stream-copy-to-stream.php

### DIFF
--- a/standard/standard_6.php
+++ b/standard/standard_6.php
@@ -573,7 +573,7 @@ function stream_socket_pair ($domain, $type, $protocol) {}
  * @param int $offset [optional] <p>
  * The offset where to start to copy data
  * </p>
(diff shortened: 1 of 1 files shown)


standard/standard_6.php
```diff
@@ -573,7 +573,7 @@ function stream_socket_pair ($domain, $type, $protocol) {}
  * @param int $offset [optional] <p>
  * The offset where to start to copy data
  * </p>
- * @return int the total count of bytes copied.
+ * @return int|bool the total count of bytes copied, or false on failure.
```
b1rdex: Could you please change this to `int|false`?

hollodotme:
Changed the return type declaration to `int|false`.

PR #396, opened 2018-08-03
add .gitignore

codemasher:
a .gitignore with the `.idea` path added.


MaXal:
.idea should be part of the project to simplify project configuration. All unnecessary files are excluded already.

PR #401, opened 2018-08-17
Add functions and constants of the uopz extension

VadimKulagin:
I added the missing functions and constants from here: http://php.net/manual/en/ref.uopz.php


VadimKulagin:
Related issue: https://youtrack.jetbrains.com/issue/WI-43240

PR #404, opened 2018-08-29
hash_hmac_algos

xarhsdev:
Fix hash_hmac_algos return type ([Source](http://php.net/manual/en/function.hash-hmac-algos.php)).

### DIFF
--- a/hash/hash.php
+++ b/hash/hash.php
@@ -232,7 +232,7 @@ function hash_hkdf(string $algo , string $ikm, int $length = 0, string $info = '
 /**
  * @since 7.2
  * Return a list of registered hashing algorithms suitable for hash_hmac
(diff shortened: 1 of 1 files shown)


hash/hash.php
```diff
@@ -232,7 +232,7 @@ function hash_hkdf(string $algo , string $ikm, int $length = 0, string $info = '
 /**
  * @since 7.2
  * Return a list of registered hashing algorithms suitable for hash_hmac
- * @return string Returns a numerically indexed array containing the list of supported hashing algorithms suitable for {@see hash_hmac()}.
+ * @return array Returns a numerically indexed array containing the list of supported hashing algorithms suitable for {@see hash_hmac()}.
```
b1rdex: Could you please make it `string[]`?

xarhsdev: Done

PR #409, opened 2018-09-18
[+]: fix for "mysqli_stmt_init"

voku:
-> http://php.net/manual/en/mysqli.stmt-init.php

---
This change is [Reviewable](https://reviewable.io/reviews/jetbrains/phpstorm-stubs/409)

### DIFF
--- a/mysqli/mysqli.php
+++ b/mysqli/mysqli.php
@@ -2254,10 +2254,11 @@ function mysqli_stmt_field_count ($stmt) {}
 
 /**
  * Initializes a statement and returns an object for use with mysqli_stmt_prepare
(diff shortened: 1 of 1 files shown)


mysqli/mysqli.php
```diff
@@ -2254,10 +2254,11 @@ function mysqli_stmt_field_count ($stmt) {}
 
 /**
  * Initializes a statement and returns an object for use with mysqli_stmt_prepare
- * @link http://fr2.php.net/manual/en/mysqli.stmt-init.php
+ * @link http://php.net/manual/en/mysqli.stmt-init.php
```
b1rdex: Please use https link

PR #410, opened 2018-09-20
Add missing arguments for memcache procedural style functions

tmotyl:
See source code https://github.com/php/pecl-caching-memcache/blob/master/memcache.c 
and documentation examples on php.net. E.g. http://php.net/memcache_decrement

### DIFF
--- a/memcache/memcache.php
+++ b/memcache/memcache.php
@@ -435,19 +435,46 @@ function memcache_delete () {}
  */
 function memcache_debug ($on_off) {}
 
-function memcache_get_stats () {}
+/**
+ * @var object $memcache
+ */
+function memcache_get_stats ($memcache) {}
 
-function memcache_get_extended_stats () {}
+/**
+ * @var object $memcache
+ */
+function memcache_get_extended_stats ($memcache) {}
 
-function memcache_set_compress_threshold () {}
+/**
+ * @var object $memcache
+ * @var int $treshold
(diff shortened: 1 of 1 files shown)


memcache/memcache.php
```diff
@@ -435,19 +435,46 @@ function memcache_delete () {}
  */
 function memcache_debug ($on_off) {}
 
-function memcache_get_stats () {}
+/**
+ * @var object $memcache
+ */
+function memcache_get_stats ($memcache) {}
```
andrey-sokolov: looks like this method accepts more than one parameter http://php.net/manual/en/memcache.getstats.php

memcache/memcache.php
```diff
@@ -435,19 +435,46 @@ function memcache_delete () {}
...
 
-function memcache_get_extended_stats () {}
+/**
+ * @var object $memcache
+ */
+function memcache_get_extended_stats ($memcache) {}
```
andrey-sokolov: http://php.net/manual/en/memcache.getextendedstats.php

andrey-sokolov:
Thank you for finding issues and investigating it. Looks like some signatures can be improved. Please double-check and fix it.

tmotyl:
This is not nice. 
Please improve the patch as you wish and merge it. I dont have time to work on this right now.
The problem is that currently when you follow the advices from phpstorm inspection you will end up with broken code.

PR #412, opened 2018-09-26
removed duplicate entry for setValue

Powerhamster:
#WI-42674

### DIFF
--- a/Reflection/Reflection.php
+++ b/Reflection/Reflection.php
@@ -1517,15 +1517,6 @@ class ReflectionProperty implements Reflector {
 	public function setValue ($object, $value) {}
 
 	/**
-	 * Set static property value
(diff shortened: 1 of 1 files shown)


andrey-sokolov:
`setValue` method can accept one parameter -  the problem is that we can't do method overloading in php stubs - perhaps it's worth to leave one declaration of method but make it compatible for both method signatures

Powerhamster:
How could it be compatible? I don't know any documentation syntax where the first parameter can be made optional.

andrey-sokolov:
https://github.com/JetBrains/phpstorm-stubs/pull/417 I've fixed it in a way I suggested - it will allow to use both kinds of call in PhpStorm without warning. Thank you for participating

PR #414, opened 2018-09-28
Remove duplicate is_iterable()

b1rdex:
Duplicate is at https://github.com/JetBrains/phpstorm-stubs/blob/db07f4569db610bcd6124f8d684d9f9315cd983a/Core/Core.php#L1096-L1103 with phpdoc.


b1rdex:
Update: I created PR to fix phpdoc for Core/Core/is_iterable: #415

PR #431, opened 2018-10-20
 Add stubs for the Saxon/C extension

aboks:
Stubs are based on the API docs at https://www.saxonica.com/saxon-c/doc/html/index.html#php-api,
with a couple of fixes for typos and obvious errors.


andrey-sokolov:
Thank you for your contribution and sorry for the delayed response. We’ve decided to start accepting non-standard extensions but please note that they will be disabled by default.

PR #436, opened 2018-10-30
Update pthreads stub based on latest docs (v3.1.6)

bakedpotat0:
See issue [WI-44013](https://youtrack.jetbrains.com/issue/WI-44013)


bakedpotat0:
Couple of late additions; fixed Threaded::extend() since it's a static function (PHP docs omit the 'static' keyword, but the examples use it as such and the [extension's source code](https://github.com/krakjoe/pthreads/blob/v3.1.6/classes/threaded.h#L113) confirms it), and added a few undocumented (but fully implemented/usable) ref-counting functions.

PR #438, opened 2018-10-30
Fix dir() URL and update its param descriptions based on docs

bakedpotat0:
See [WI-43883](https://youtrack.jetbrains.com/issue/WI-43883) (plus I took the liberties of fixing some other documentation for that same function)

### DIFF
--- a/standard/standard_7.php
+++ b/standard/standard_7.php
@@ -355,14 +355,17 @@ function readdir ($dir_handle = null) {}
 
 /**
  * Return an instance of the Directory class
- * @link https://php.net/manual/en/class.dir.php
- * @param $directory
- * @param $context [optional]
- * @return Directory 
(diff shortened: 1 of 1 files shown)


standard/standard_7.php
```diff
@@ -355,14 +355,17 @@ function readdir ($dir_handle = null) {}
...
- * @link https://php.net/manual/en/class.dir.php
- * @param $directory
- * @param $context [optional]
- * @return Directory 
+ * @link https://php.net/manual/en/function.dir.php
+ * @param string $directory <p>
+ * Directory to open
+ * </p>
+ * @param resource $context [optional]
+ * @return Directory an instance of Directory, or <b>NULL</b> with wrong
```
b1rdex: Please change return type to `Directory|null|false`

PR #430, opened 2018-10-11
Improve docs for DateTimeImmutable constructor

gnumoksha:
Improve documentation for DateTime and DateTimeImmutable constructors. The motivation behind this is the missing @throw on DateTime constructor.

PR #439, opened 2018-10-31
Add @throws Exception on DateTime

tolbon:
Same definition than DateTimeImmutable


tolbon:
I close I didn't see stuff do in PR #430

PR #456, opened 2018-11-22
Made return types of ibm_db2 functions more accurate

morozov:
Similarly to https://github.com/JetBrains/phpstorm-stubs/pull/455, when analyzing the code using `ibm_db2` with PHPStan, some existing stubs cause [false positive issues](https://phpstan.org/r/4d4967ac-46e6-4f1f-a0c7-bfc533ddaaa5).

Additionally, `[optional]` markers have been removed from parameter annotations since they are redundant, non-standard and may conflict with the `[]` ones. Also, parameter annotations like `@param variable type` have been converted to `@param type $variable` for compliance with phpDoc.

Despite the fact that the `$value` argument of `db2_autocommit()` is [documented](https://secure.php.net/manual/en/function.db2-autocommit.php) as boolean, it accepts one of the `DB2_AUTOCOMMIT_*` constants which are integer. The method signature has been modified to avoid false positive type issues.

### DIFF
--- a/ibm_db2/ibm_db2.php
+++ b/ibm_db2/ibm_db2.php
@@ -102,7 +102,7 @@ function db2_pconnect ($database, $username, $password, array $options = null) {
  * A valid database connection resource variable as returned from
  * db2_connect or db2_pconnect.
  * </p>
- * @param bool $value <p>
+ * @param int $value <p>
  * One of the following constants: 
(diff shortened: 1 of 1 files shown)


morozov:
@andrey-sokolov could you take a look at this as well?

PR #461, opened 2018-12-02
Real return types

thesebas:
Return types of those methods are hinted as `static`, it is formally correct but later method generator for methods that consumes variables returned by this method stubs new method with `static` argument type. Example:
```
$startDate = new \DateTimeImmutable();
$endDate = $startDate->add($interval); // hinted as `static`
$object->newMethodDoSomething($startDate, $endDate); // not existing yet, using "Generate method"

// inside class of `$object`

function newMethodDoSomething(\DateTimeImmutable $startDate, \static $endDate) {}
```

### DIFF
--- a/date/date_c.php
+++ b/date/date_c.php
@@ -115,7 +115,7 @@ class DateTimeImmutable implements DateTimeInterface {
      * (PHP 5 &gt;=5.5.0)<br/>
      * Adds an amount of days, months, years, hours, minutes and seconds
      * @param DateInterval $interval
-     * @return static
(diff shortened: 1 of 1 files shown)


b1rdex:
IMO this is bug in PhpStorm, as phpdoc `static`, `self` and `$this` should work as expected.

andrey-sokolov:
I think it should be fixed from refactoring side - I've created an issue https://youtrack.jetbrains.com/issue/WI-44617

andrey-sokolov:
Thank you for reporting the problem.

PR #464, opened 2018-12-06
msgfmt_format_message takes only 3 args

ksimka:
According to docs (https://secure.php.net/manual/en/messageformatter.formatmessage.php)

```
string msgfmt_format_message ( string $locale , string $pattern , array $args )
```


b1rdex:
This was introduced with e598e6d0483eccf23cfc3e6aef5bb09e3c5f19d2

PR #465, opened 2018-12-07
New architecture of test suite

isfedorov:
I've splitted all the classes to separate files. Made autoloader for those classes. Added model classes for php entities to avoid stdClass. Made test run possible without transition generation of json file. Now it's required to perform just `composer install` and `docker-compose -f docker-compose.yml run php` from root folder because I've moved docker-compose.yml file to the root of project


tests/model/PHPElementWithPHPDoc.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+namespace Model;
+
+abstract class PHPElementWithPHPDoc extends BasePHPElement
```
MaXal: Maybe make this one a trait? I have no strong opinion about this one though so up to you.

tests/model/PHPFunction.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
+namespace Model;
+
+use ReflectionException;
+use ReflectionFunction;
+
+class PHPFunction extends PHPElementWithPHPDoc
+{
+    public $name;
```
MaXal: There is already this field in base class.

tests/testdata/MutedProblems.php
```diff
@@ -0,0 +1,92 @@
...
+
+namespace TestData;
+
+
+use stdClass;
+
+class MutedProblems
+{
+    /** @var stdClass */
+    private $mutedProblems;
```
MaXal: This can also be the class like one in `StubParser::$stubs` and `PHPReflectionParser::$data`

tests/TestStubs.php
```diff
@@ -125,43 +24,27 @@ public static function setUpBeforeClass()/* The :void return type declaration th
...
-
     /**
-     * @dataProvider constantProvider
+     * @dataProvider \TestData\Providers\ReflectionTestDataProviders::constantProvider
      */
-    public function testConstants(stdClass $constant)
+    public function testConstants(PHPConst $constant)
     {
         $constantName = $constant->name;
         $constantValue = $constant->value;
```
MaXal: We can actually try to compare const values in stubs and in runtime. In case, there are not a lot of false positives that should be useful.

tests/TestStubs.php
```diff
@@ -125,43 +24,27 @@ public static function setUpBeforeClass()/* The :void return type declaration th
...
     /**
-     * @dataProvider functionProvider
+     * @dataProvider \TestData\Providers\ReflectionTestDataProviders::functionProvider
      */
-    public function testFunctions(stdClass $function)
+    public function testFunctions(PHPFunction $function)
     {
         $functionName = $function->name;
-        $stubFunctions = PhpStormStubsSingleton::getPhpStormStubs()->functions;
+        $stubFunctions = PhpStormStubsSingleton::getPhpStormStubs()[PHPFunction::class];
```
MaXal: Please provide PHPDocs: `/** @var PHPFunction[] $stubFunctions */`

tests/TestStubs.php
```diff
@@ -175,27 +58,15 @@ public function testFunctions(stdClass $function)
...
     /**
-     * @dataProvider classProvider
+     * @dataProvider \TestData\Providers\ReflectionTestDataProviders::classProvider
      */
-    public function testClasses(stdClass $class)
+    public function testClasses(PHPClass $class)
     {
         $className = $class->name;
-        $stubClasses = PhpStormStubsSingleton::getPhpStormStubs()->classes;
+        $stubClasses = PhpStormStubsSingleton::getPhpStormStubs()[PHPClass::class];
```
MaXal: `/** @var PHPClass[] $stubClass */`

tests/TestStubs.php
isfedorov: Well, I've tried. And the result is next: Tests: 1528, Failures: 88. Most of the failures are platform dependent but taking in account that it's just 88 of 1528 constants probably it makes sense to add this test

MaXal: The numbers look indeed ok to include the check. We can turn it off anytime if we will find that most new constants are platform dependant.

.idea/phpstorm-stubs.iml
```diff
@@ -2,6 +2,10 @@
 <module type="WEB_MODULE" version="4">
   <component name="NewModuleRootManager">
     <content url="file://$MODULE_DIR$">
+      <sourceFolder url="file://$MODULE_DIR$/tests/parser/visitor" isTestSource="true" packagePrefix="Parsers\Visitor\" />
```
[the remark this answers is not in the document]
isfedorov: They are marked due to sync with composer.json. Unfortunately it's requred to create custom autoloader to be able to include only root folder in composer.json and load all the classes

tests/TestStubs.php
isfedorov: Done. You can find the list of failing constants in mutedProblems.json

tests/testdata/MutedProblems.php
isfedorov: It will be done in a separate PR later

tests/parser/StubParser.php
```diff
@@ -0,0 +1,55 @@
...
+use RecursiveIteratorIterator;
+use SplFileInfo;
+
+class StubParser
+{
+    public static function getPhpStormStubs(): array
+    {
+        $parser = (new ParserFactory)->create(ParserFactory::PREFER_PHP7);
+        $docFactory = DocBlockFactory::createInstance();
+        $stubs = array();
```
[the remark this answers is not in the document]
isfedorov: It will be done in a separate PR

tests/parser/PHPReflectionParser.php
```diff
@@ -0,0 +1,50 @@
...
+use Model\PHPInterface;
+use RecursiveArrayIterator;
+use RecursiveIteratorIterator;
+use ReflectionClass;
+
+class PHPReflectionParser
+{
+    public static function getStubs(): array
+    {
+        $data = [];
```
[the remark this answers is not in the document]
isfedorov: Will be done in a separate PR

tests/model/PHPElementWithPHPDoc.php
isfedorov: May be..will think about it and look what looks better

tests/TestStubs.php
```diff
@@ -1,391 +1,300 @@
...
+            $params     = $this->getParameterRepresentation($method);
             $methodName = $method->name;
             if (!in_array('missing method', self::$mutedProblems->getMutedProblemsForMethod($className, $methodName), true)) {
-                $this->assertArrayHasKey($methodName, $stubClass->methods, "Missing method $className::$methodName($params){}");
+                static::assertArrayHasKey(
+                    $methodName,
+                    $stubClass->methods,
+                    "Missing method $className::$methodName($params){}"
+                );
                 $stubMethod = $stubClass->methods[$methodName];
```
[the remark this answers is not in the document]
isfedorov: seems this is not needed after adding PHPDoc to $stubClass. Method is resolved properly after that

PR #479, opened 2019-01-07
Add missing SPL filesystem exceptions

bugreportuser:
These exceptions are all documented in the PHP Manual.
https://secure.php.net/manual/en/book.spl.php

Fixes the second part of https://youtrack.jetbrains.com/issue/WI-8146, which was wrongly closed as a duplicate.


bugreportuser:
Fixes the second part of https://youtrack.jetbrains.com/issue/WI-8146, which was wrongly closed as a duplicate.

PR #483, opened 2019-01-16
Correct session_gc signature

tvlooy:
you see the cast to int seems unnecessary but that indication is false because the function returns false is no sessions were deleted

screenshot from 2019-01-16 16-23-26

### DIFF
--- a/session/session.php
+++ b/session/session.php
@@ -190,7 +190,7 @@ function session_create_id($prefix) {}
 
 /**
  * Perform session data garbage collection
- * @return int number of deleted session data for success, FALSE for failure.
+ * @return int|false number of deleted session data for success, false for failure.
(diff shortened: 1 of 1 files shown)


session/session.php
```diff
@@ -190,7 +190,7 @@ function session_create_id($prefix) {}
 
 /**
  * Perform session data garbage collection
- * @return int number of deleted session data for success, FALSE for failure.
+ * @return int|false number of deleted session data for success, false for failure or 0 sessions cleared.
```
b1rdex: I can't find `false for 0 sessions cleared` in documentation. Where do you found that?

tvlooy: yes you are right. Could not clear sessions does not mean 0 sessions cleared

PR #484, opened 2019-01-16
Correct naming and default value of mbstring $before_needle params

signe:
The "$part" parameter is incorrectly named and defines an incorrect default value (null) for haystack/needle mbstring functions.  Correct name is $before_needle and value is false.


andrey-sokolov:
We've decided to revert formatting changes. Now the change contains only `$part` -> `$before_needle` replacements. Thank you for your contribution.

PR #485, opened 2019-01-17
Made return types of mysqli functions more accurate

morozov:
1. Added missing `|false` to some function which can return `false`.
2. For functions which return `true` but don't return `false`, replaced `bool` with `false` since the former implies `true` is also possible.
3. [`mysqli_get_client_version()`](http://php.net/manual/en/mysqli.get-client-version.php#refsect1-mysqli.get-client-version-returnvalues) returns the version ID as an `int`, not as a `string`.
4. [`mysqli_stmt_errno()`](http://php.net/manual/en/mysqli-stmt.errno.php#refsect1-mysqli-stmt.errno-returnvalues) returns the error code, not a `bool`.

### DIFF
--- a/mysqli/mysqli.php
+++ b/mysqli/mysqli.php
@@ -283,7 +283,7 @@ class mysqli  {
 	/**
 	 * Returns statistics about the client connection
 	 * @link https://php.net/manual/en/mysqli.get-connection-stats.php
-	 * @return bool an array with connection stats if success, false otherwise.
+	 * @return array|false an array with connection stats if success, false otherwise.
 	 * @since 5.3.0
 	 */
 	public function get_connection_stats () {}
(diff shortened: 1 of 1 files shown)


mysqli/mysqli.php
```diff
@@ -2125,7 +2125,7 @@ function mysqli_report ($flags) {}
  * @param mysqli $link A link identifier returned by mysqli_connect() or mysqli_init()
  * @param string $query An SQL query
  * @param int $resultmode
- * @return mysqli_result|bool
+ * @return mysqli_result|false
  * For successful SELECT, SHOW, DESCRIBE or EXPLAIN queries, mysqli_query() will return a mysqli_result object.
  * For other successful queries mysqli_query() will return TRUE.
```
b1rdex: Is this wrong or change should be reverted?

morozov: The original type was correct, I reverted the change.

PR #487, opened 2019-01-22
WI-854 Provide expectedArguments for some standard functions/methods

wbars:


b1rdex:
This is 🔥 Can't wait for EAP.

PR #491, opened 2019-01-27
Set correct return type for insert method

jaapio:
The official php-documentation was even incorrect at this point as
shown in https://3v4l.org/X3h8Q
Insert returns true when successfull. It is unclear if it can ever
return false.

### DIFF
--- a/SPL/SPL_c1.php
+++ b/SPL/SPL_c1.php
@@ -1553,7 +1553,7 @@ class SplPriorityQueue implements Iterator, Countable {
          * @param mixed $priority <p>
          * The associated priority.
          * </p>
-         * @return bool 
+         * @return true
          * @since 5.3.0
          */
         public function insert ($value, $priority) {}
--- a/SPL/SPL_c1.php
+++ b/SPL/SPL_c1.php
@@ -1553,7 +1553,7 @@ class SplPriorityQueue implements Iterator, Countable {
          * @param mixed $priority <p>
          * The associated priority.
          * </p>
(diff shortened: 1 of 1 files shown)


orklah:
Shouldn't we change the type to true then?

@b1rdex Does that line means it can return void also? https://github.com/php/php-src/blob/7674a542aa870c4cba462cb2a1b68b08f9b0406e/ext/spl/spl_heap.c#L667

jaapio:
It throws an exception before the return. So I think the return will never be hit. It is an internal corrupted state so I guess it will never happen.

orklah:
Arg! I wanted to link this line:
https://github.com/php/php-src/blob/7674a542aa870c4cba462cb2a1b68b08f9b0406e/ext/spl/spl_heap.c#L660

jaapio:
Same applies there to. The return statement will never be hit because of the check in the if statement. Which triggers an error in php that an argument is missing. I don't think the return is executed.

SPL/SPL_c1.php
```diff
@@ -1553,7 +1553,7 @@ public function compare ($priority1, $priority2) {}
          * @param mixed $priority <p>
          * The associated priority.
          * </p>
-         * @return void 
+         * @return bool
```
b1rdex:
```suggestion
         * @return true
```
@andrey-sokolov I suppose you have write access to accept sugesstion too

jaapio:
Thanks for the improvement

PR #493, opened 2019-01-29
Added the missing @throws tag in SoapClient::__construct()

rgrellmann:


soap/soap.php
```diff
@@ -255,6 +255,7 @@ public function SoapClient ($wsdl, array $options = null) {}
      * <b>SOAP_SSL_METHOD_SSLv3</b> or
      * <b>SOAP_SSL_METHOD_SSLv23</b>.
      * </p>
+	 * @throws SoapFault A SoapFault exception will be thrown if the wsdl URI cannot be loaded.
```
b1rdex:
```suggestion
     * @throws SoapFault A SoapFault exception will be thrown if the wsdl URI cannot be loaded. 
```

PR #499, opened 2019-02-02
feat: Add expected args for IDN functions

mirucon:
This adds the expected args for `idn_to_ascii()` and `idn_to_utf8()`.

source: http://php.net/manual/en/ref.intl.idn.php


meta/.phpstorm.meta.php
```diff
@@ -403,6 +403,10 @@ function argumentsSet($setName, $values) {
     expectedArguments(\json_encode(), 1, JSON_HEX_QUOT | JSON_HEX_TAG | JSON_HEX_AMP | JSON_HEX_APOS | JSON_NUMERIC_CHECK | JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES | JSON_FORCE_OBJECT | JSON_PRESERVE_ZERO_FRACTION | JSON_UNESCAPED_UNICODE | JSON_PARTIAL_OUTPUT_ON_ERROR | JSON_UNESCAPED_LINE_TERMINATORS | JSON_THROW_ON_ERROR);
     expectedArguments(\json_decode(), 3, JSON_BIGINT_AS_STRING | JSON_OBJECT_AS_ARRAY | JSON_THROW_ON_ERROR);
 
+    expectedArguments(\idn_to_ascii(), 1, IDNA_ALLOW_UNASSIGNED | IDNA_CHECK_BIDI | IDNA_CHECK_CONTEXTJ | IDNA_DEFAULT | IDNA_NONTRANSITIONAL_TO_ASCII | IDNA_NONTRANSITIONAL_TO_UNICODE | IDNA_USE_STD3_RULES);
+    expectedArguments(\idn_to_ascii(), 2, INTL_IDNA_VARIANT_2003 | INTL_IDNA_VARIANT_UTS46);
```
wbars: Since `INTL_IDNA_VARIANT_2003` is deprecated starting with 7.2 it will be more convenient to set `INTL_IDNA_VARIANT_UTS46` as first argument since order of expectedArguments control the order in completion list. (Same for `\idn_to_utf8()`)

King2500:
```suggestion
    expectedArguments(\idn_to_ascii(), 2, INTL_IDNA_VARIANT_UTS46, INTL_IDNA_VARIANT_2003);
```
Use `,` instead of `|` delimiter because it can only by one or the other used. (not combined as flags argument)

meta/.phpstorm.meta.php
```diff
@@ -403,6 +403,10 @@ function argumentsSet($setName, $values) {
...
+    expectedArguments(\idn_to_utf8(), 1, IDNA_ALLOW_UNASSIGNED | IDNA_CHECK_BIDI | IDNA_CHECK_CONTEXTJ | IDNA_DEFAULT | IDNA_NONTRANSITIONAL_TO_ASCII | IDNA_NONTRANSITIONAL_TO_UNICODE | IDNA_USE_STD3_RULES);
+    expectedArguments(\idn_to_utf8(), 2, INTL_IDNA_VARIANT_2003 | INTL_IDNA_VARIANT_UTS46);
```
King2500:
```suggestion
    expectedArguments(\idn_to_utf8(), 2, INTL_IDNA_VARIANT_UTS46, INTL_IDNA_VARIANT_2003);
```
(same as commented above)

mirucon:
Thanks @wbars and @King2500, made changes as you suggested and resolved conflicts :)

PR #501, opened 2019-02-05
Add expectedArguments for debug_print_backtrace() and debug_backtrace()

aalwash:
While we're on it, can someone explain me the difference between:

Variant 1
` expectedArguments(\debug_backtrace(), 0, DEBUG_BACKTRACE_PROVIDE_OBJECT|DEBUG_BACKTRACE_IGNORE_ARGS);`

Variant 2
` expectedArguments(\debug_backtrace(), 0, DEBUG_BACKTRACE_PROVIDE_OBJECT, DEBUG_BACKTRACE_IGNORE_ARGS);`

When should we use which variant?


King2500:
Variant 1 is for constants used as flags (bit-wise OR) in code like this:
`debug_backtrace(DEBUG_BACKTRACE_PROVIDE_OBJECT | DEBUG_BACKTRACE_IGNORE_ARGS);`

Variant 2 means you can have only one const in your calls:
`debug_backtrace(DEBUG_BACKTRACE_PROVIDE_OBJECT);`
or
`debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS);`

Basically it depends how the constant is defined (bit-wise constants typically have values like 1, 2, 4, 8, 16, ....).

PR #502, opened 2019-02-05
Updated PHP Doc links (issue WI-45218)

Girgias:
Updated PHP doc links to use https://secure.php.net

YouTrack issue: WI-45218 (https://youtrack.jetbrains.com/issue/WI-45218)


nikic:
Could you please explain in more detail what the motivation for this change is? While https://php.net currently always redirects to https://secure.php.net, this is only because it is the only host currently supporting HTTPS. However, if/when HTTPS is fully supported by the PHP documentation mirror system, different hosts may be chosen based on geolocation, just like it is currently done for HTTP. To properly support this, it is preferable to not hard-code URLs for any particular host.

Girgias:
Just saw the issue on the PHPStorm YouTrack and thought that wouldn't be a difficult change to execute.
Does also https://www.php.net redirect to https://secure.php.net ? If that's the case I'll close the PR otherwise I can rewrite all the links from https://secure.php.net to https://php.net

b1rdex:
Yes, it does redirect too.

```
$ curl --head https://www.php.net
HTTP/1.1 301 Moved Permanently
Server: myracloud
Date: Fri, 08 Feb 2019 04:29:01 GMT
Content-Type: text/html
Content-Length: 185
Connection: keep-alive
Location: https://secure.php.net/
Expires: Fri, 08 Feb 2019 04:29:01 GMT
Cache-Control: max-age=0
ETag: "myra-18b72b1"
```

Girgias:
In that case, I'm closing the PR.

nikic:
To follow up here, it looks like the mirroring system is going away and we'll hopefully get HTTPS on the main php.net domain: https://externals.io/message/104533

PR #503, opened 2019-02-06
DOMNodeList::item() returns DOMNode instead of DOMElement

smoench:
https://secure.php.net/manual/en/domnodelist.item.php

### DIFF
--- a/dom/dom_c.php
+++ b/dom/dom_c.php
@@ -1147,7 +1147,7 @@ class DOMNodeList implements Traversable, Countable {
 	 * @param int $index <p>
 	 * Index of the node into the collection.
 	 * </p>
-	 * @return DOMNode The node at the indexth position in the 
+	 * @return DOMNode|null The node at the indexth position in the 
 	 * DOMNodeList, or &null; if that is not a valid
 	 * index.
 	 * @since 5.0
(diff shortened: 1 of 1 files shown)


orklah:
Seems like it can return null too


> Return Values
> 
> The node at the indexth position in the DOMNodeList, or NULL if that is not a valid index.
>

dom/dom_c.php
```diff
@@ -1147,7 +1147,7 @@ class DOMNodeList implements Traversable, Countable {
 	 * @param int $index <p>
 	 * Index of the node into the collection.
 	 * </p>
-	 * @return DOMElement The node at the indexth position in the 
+	 * @return DOMNode The node at the indexth position in the
```
b1rdex:
```suggestion
	 * @return DOMNode|null The node at the indexth position in the 
```

MaXal:
That is incorrect and should be reverted: https://3v4l.org/4Z5oj

smoench:
@MaXal your code example is incorrect. This would be a valid one https://3v4l.org/EL06S
`DOMElement` extends from `DOMNode` -> http://php.net/DomElement
The php documents says `DOMNodeList::node()` returns `DOMNode|null` -> http://php.net/manual/en/domnodelist.item.php

PR #504, opened 2019-02-06
Added new functions of geoip 1.1.0 WI-45203

aalwash:
YouTrack issue: https://youtrack.jetbrains.com/issue/WI-45203


meta/.phpstorm.meta.php
```diff
@@ -413,6 +413,7 @@ function argumentsSet($setName, $values) {
     expectedArguments(\debug_print_backtrace(), 0, DEBUG_BACKTRACE_IGNORE_ARGS);
     expectedArguments(\debug_backtrace(), 0, DEBUG_BACKTRACE_PROVIDE_OBJECT|DEBUG_BACKTRACE_IGNORE_ARGS);
 
+    expectedArguments(\geoip_database_info(), 0, GEOIP_COUNTRY_EDITION);
```
wbars: According to the [docs](http://php.net/manual/en/function.geoip-database-info.php) the function expect several constants, not just one. Could you please add them as well? 
> You can use the various constants defined with this extension (ie: GEOIP_*_EDITION).

aalwash: Thanks for notifying, I will fix this asap and make a new PR

I did a upstream master merge, but accidentally merged the wrong commits
So this PR is broken, will make a new one tonight

aalwash: Fix the PR and reverted to upstream master changes + my geoip meta changes

Let me know if you're OK with this.

wbars:
Thank you.

PR #508, opened 2019-02-12
Add swoole extension support

SerafimArts:
This is direct subsplit of this repository https://github.com/swoole/ide-helper

It is much more convenient when autocomplete on this extension is already available immediately in php storm.


SerafimArts:
It is necessary to understand what to do with broken tests. If I fix them, then we lose the actual extension code. Those. it will not be possible just to take and copy them once every six months from the original (https://github.com/swoole/ide-helper) repository. It will be necessary each time their merge.

andrey-sokolov:
Thank you for contribution. Unfortunately, we can't accept it as it is since every pull-request should follow our project-specific contracts which are enforced by tests. The build should be green after the request to be accepted. Tests specifies found problems, in this case you should remove `@return` annotations in constructors doc.

SerafimArts:
@andrey-sokolov Those. the fact that you have to follow the changes, and not to copy once a N time to update the source code - is not it scary?

Maybe you have any CI system which will automate fix helpers? Well, or just think about it, because almost all the work done in this repository can be automated.

andrey-sokolov:
@SerafimArts  We are improving phpstorm-stubs data and automating some verification steps so we’ll consider an auto-fixing system. For now, we need to fix found problems manually and we’ll be glad to get ‘green’ PR.

SerafimArts:
@andrey-sokolov kk

GLeoLi:
@SerafimArts 
[swoole/ide-helper](https://github.com/swoole/ide-helper) doesn't make sense. 
All functions' returns are **mixed**. No data type for parameters, no const name for parameters' default values. The **since** tag is used by mistake.

This package is useless by now.

andrey-sokolov:
@SerafimArts Hey. Since there are no updates on this PR for some time I am going to close it. Please, if you fix the problems, create a new PR.

SerafimArts:
@andrey-sokolov ok

PR #509, opened 2019-02-12
Add ffi extension support

SerafimArts:
Adds autocomplete and a brief description of the FFI extension from the current php-src master (PHP 8.0.0-dev) branch.

Relation link: https://github.com/dstogov/php-ffi

### DIFF
--- a/FFI/FFI.php
+++ b/FFI/FFI.php
@@ -1,0 +1,279 @@
+<?php
+
+// Start of FFI v.0.1.0
+
+namespace {
+    use FFI\CData;
+    use FFI\CType;
+    use FFI\ParserException;
+
+    /**
+     * Class FFI
+     *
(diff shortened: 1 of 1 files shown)


SerafimArts:
Hmmm... I do not understand the reason why tests do not pass. Can anyone explain the reason?

MaXal:
@SerafimArts Sorry, it was our fault. I've restarted the build.

andrey-sokolov:
Perhaps, it worth to wait till PHP 8.0 released in case the extension-API will be completely changed. For now, it should be enough to add this stub locally.

SerafimArts:
Then it will be no longer version 0.1.0 of this extension. Approximately as in the case of `ext-ev`, when API was completely rewritten during the transition from 1.0 to 2.0 (if I do not confuse)

PR #511, opened 2019-02-14
Mark Traversable internal

b1rdex:
I'm not sure if this will mark it stroked in client code.


b1rdex:
See https://youtrack.jetbrains.com/issue/WI-45401

PR #515, opened 2019-02-20
Revert "DOMNodeList::item() returns DOMNode instead of DOMElement"

MaXal:
Reverts JetBrains/phpstorm-stubs#503

The return type is actually `DOMElement`: https://3v4l.org/4Z5oj


smoench:
@MaXal Really?

PR #517, opened 2019-02-21
Revert "Revert "DOMNodeList::item() returns DOMNode instead of DOMElement""

b1rdex:
Reverts JetBrains/phpstorm-stubs#515

@MaXal @smoench here is test code: https://3v4l.org/VsnW7

### DIFF
--- a/dom/dom_c.php
+++ b/dom/dom_c.php
@@ -1147,7 +1147,7 @@ class DOMNodeList implements Traversable, Countable {
 	 * @param int $index <p>
 	 * Index of the node into the collection.
 	 * </p>
-	 * @return DOMElement The node at the indexth position in the 
+	 * @return DOMNode|null The node at the indexth position in the 
 	 * DOMNodeList, or &null; if that is not a valid
 	 * index.
 	 * @since 5.0


b1rdex:
NB: `is_a(DOMElement::class, DOMNode::class, true) === true`

smoench:
@b1rdex Thank you :-)

MaXal:
I'm sorry my code sample was really bad and not illustrative at all. But the main point remains so far - the return type is `DOMElement`. Yes `DOMElement extends `DOMNode` but we should specify the most specific class, not the parent.

Please note that my change was caused by new false positive inspections errors that occured after the change on the real-life code in the popular projects (Zend, Mediawiki, b2c).

Consider the following code:
```
<?php
$html = ' 
      <tr></tr>
   ';
$dom = new domDocument;
$dom->loadHTML($html);
$tags = $dom->getElementsByTagName("tr");
f($tags->item(0));

function f(DOMElement $a){
    echo "in f";
}
```
It will produce a warning in case we change the return type to `DOMNode`.

Please also note that I have nothing against adding `|null` to the return type. I just against changing 
`DOMElement` to `DOMNode`. 

If you have any real example when the `item` returns `DOMNode` and not `DOMElement` please provide it and I will merge the request right away.

smoench:
The [PHP documentation](http://php.net/manual/en/domnodelist.item.php) says it returns `DOMNode` so phpstub should do it. Please read the class name carefully: `DOMNodeList`. It says it is a list of `DOMNode`'s. So don't change the API of PHP by your own!

In my case phpstan has been failed and said my code is wrong as they are all relying on the PHP Api. So once again don't change the API of PHP by your own!

derrabus:
I've modified the code snippet to demonstrate that `DOMNodeList::item()` might actually return something that is _not_ a `DOMElement`.

https://3v4l.org/GvRgi

Here, the last call returns an instance of `DOMText` which – like `DOMElement` – extends `DOMNode`.

Depending on the context, it can be safe to assume that a particullar `DOMNodeList` only contains `DOMElement` instances, but in general, you cannot make that assumption.

`DOMNode` is the correct return type.

MaXal:
@smoench Unfortunately, PHP documentation is not a definite source of API and it contains errors and inaccuracies but in this case, it's indeed correct.

Thanks to @nikic for the following code example that produces `DOMNode` elements as a result of `item` call: https://3v4l.org/oSb9N

Once again I'm sorry for jumping to conclusions and thank you all for being persistent!

MaXal:
@derrabus Thank you for another example!

PR #519, opened 2019-02-25
Fix broken https php.net links

blockcounter:
Links were changed to use https without updating the url to be secure.php.net, causing the links to be broken.

### DIFF
--- a/date/date_c.php
+++ b/date/date_c.php
@@ -313,7 +313,7 @@ class DateTimeImmutable implements DateTimeInterface {
 class DateTime implements DateTimeInterface {
     /**
      * (PHP 5 &gt;=5.2.0)<br/>
-     * @link https://secure.php.net/manual/en/datetime.construct.php
+     * @link https://php.net/manual/en/datetime.construct.php
      * @param string $time [optional]
      * <p>A date/time string. Valid formats are explained in {@link www.php.net/manual/en/datetime.formats.php Date and Time Formats}.</p>
      * <p>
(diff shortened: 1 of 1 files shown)


date/date_c.php
```diff
@@ -313,15 +313,15 @@ public function __wakeup() { }
 class DateTime implements DateTimeInterface {
     /**
      * (PHP 5 &gt;=5.2.0)<br/>
-     * @link https://www.php.net/manual/en/datetime.construct.php
+     * @link https://secure.php.net/manual/en/datetime.construct.php
```
andrey-sokolov: https://www.php.net/manual/en/datetime.construct.php works
According to https://github.com/JetBrains/phpstorm-stubs/pull/502#issuecomment-461512744 It's preferable not to hardcode https://secure.php.net/ instead of https://www.php.net/

date/date_c.php
```diff
@@ -313,15 +313,15 @@ public function __wakeup() { }
...
      * @param string $time [optional]
      * <p>A date/time string. Valid formats are explained in {@link www.php.net/manual/en/datetime.formats.php Date and Time Formats}.</p>
      * <p>
      * Enter <b>now</b> here to obtain the current time when using
      * the <em>$timezone</em> parameter.
      * </p>
      * @param DateTimeZone $timezone [optional] <p>
-     * A {@link https://www.php.net/manual/en/class.datetimezone.php DateTimeZone} object representing the
+     * A {@link https://secure.php.net/manual/en/class.datetimezone.php DateTimeZone} object representing the
```
andrey-sokolov: https://www.php.net/manual/en/class.datetimezone.php is also works but it's not navigatable as well as suggested url cause of https://youtrack.jetbrains.com/issue/WI-37030

date/date_c.php
blockcounter: Updated url to not use secure subdomain, but still removed the www subdomain. httpS with www subdomain does NOT work.

date/date_c.php
blockcounter: Updated url to not use secure subdomain, but still removed the www subdomain. httpS with www subdomain does NOT work.

PR #521, opened 2019-02-26
update stub to reflect current MongoDB\Driver\Manager

bartgloudemans:
See http://php.net/manual/en/class.mongodb-driver-manager.php for up-to-date interface

See issue https://youtrack.jetbrains.com/issue/WI-45538

### DIFF
--- a/mongodb/mongodb.php
+++ b/mongodb/mongodb.php
@@ -56,7 +56,7 @@ namespace MongoDB {}
              * @return WriteResult
              * @since 1.4.0 added $options argument
              */
-            final public function executeBulkWrite($namespace, BulkWrite $bulk, array $options = [])
+            final public function executeBulkWrite($namespace, BulkWrite $bulk, $options = [])
             {
             }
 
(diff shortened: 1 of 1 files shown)


bartgloudemans:
A question in general: how does the "stub" solution work for people that are still working with an older mongodb library?

Because now the interface of an already existing, and previously valid function will be changed. (`executeBulkWrite()`)

The `@since` tag doesn't help here since it is an already existing function.

mongodb/mongodb.php
```diff
@@ -81,24 +83,69 @@ final class Manager
              * @link https://php.net/manual/en/mongodb-driver-manager.executequery.php
              * @param string $namespace A fully qualified namespace (databaseName.collectionName)
              * @param Query $query A MongoDB\Driver\Query to execute.
-             * @param ReadPreference $readPreference Optionally, a MongoDB\Driver\ReadPreference to route the command to. If none given, defaults to the Read Preferences set by the MongoDB Connection URI.
+             * @param array $options
```
andrey-sokolov: array|ReadPreference

andrey-sokolov:
For now, signatures should accept both old and new API if it's possible, cause there is only one version of stubs. I've added some suggestions on how to preserve backward compatibility in this case

bartgloudemans:
@andrey-sokolov Thanks! I've followed up on your comment.

The PR is up for review again.

PR #524, opened 2019-02-28
Fix NumberFormatter

antoniopaisfernandes:


standard/standard_7.php
```diff
@@ -185,7 +185,7 @@ function pack ($format, $args = null, $_ = null) {}
  * @since 4.0
  * @since 5.0
  */
-function unpack ($format, $data, $offset) {}
+function unpack ($format, $data, $offset = null) {}
```
wbars: [docs](http://php.net/manual/en/function.unpack.php) specify default value of the `$offset` as `0` rather then `null`

standard/standard_7.php
```diff
@@ -19,7 +19,7 @@
  * <p>
  * If the value returned in errno is
  * 0 and the function returned false, it is an
- * indication that the error occurred before the
```
wbars: Please, try to avoid irrelevant changes to preserve git history.

standard/standard_7.php
wbars: Also it seems this function is already fixed in the earlier PR #522 :)

PR #525, opened 2019-03-02
CHANGED param types for Redis to satisfy phpstan

liamsorsby:
PHP 7.2 + phpstan complains when the value is type hinted to a string as it states it can only accept an int or an array

### DIFF
--- a/redis/RedisCluster.php
+++ b/redis/RedisCluster.php
@@ -209,7 +209,7 @@ class RedisCluster {
     /**
      * Remove specified keys.
      *
-     * @param   int | array $key1 An array of keys, or an undefined number of parameters, each a key: key1 key2 key3
+     * @param   int|string...|array    $key1 An array of keys, or an undefined number of parameters, each a key: key1 key2 key3
      *                            ... keyN
      * @param   string      $key2 ...
      * @param   string      $key3 ...
--- a/redis/RedisCluster.php
+++ b/redis/RedisCluster.php
@@ -209,10 +209,9 @@ class RedisCluster {
     /**
      * Remove specified keys.
      *
-     * @param   int|string...|array    $key1 An array of keys, or an undefined number of parameters, each a key: key1 key2 key3
+     * @param int|string|array $key1 An array of keys, or an undefined number of parameters, each a key: key1 key2 key3
      *                            ... keyN
-     * @param   string      $key2 ...
-     * @param   string      $key3 ...
+     * @param int|string ...$otherKeys
      *
      * @return int Number of keys deleted.
      * @link    https://redis.io/commands/del
--- a/redis/Redis.php
+++ b/redis/Redis.php
@@ -373,7 +373,7 @@ class Redis
     /**
      * Remove specified keys.
      *
-     * @param   string|string[] $key1 An array of keys, or an undefined number of parameters, each a key: key1 key2 key3 ... keyN
+     * @param   string|array    $key1 An array of keys, or an undefined number of parameters, each a key: key1 key2 key3 ... keyN
      * @param   string          $key2 ...
      * @param   string          $key3 ...
      * @return  int             Number of keys deleted.


.gitignore
```diff
@@ -10,6 +10,7 @@ rest-client.private.env.json
 # Default ignored files
 .idea/workspace.xml
 .idea/shelf
+.idea/deployment.xml
```
andrey-sokolov: Please, get rid of this change

redis/RedisCluster.php
```diff
@@ -209,10 +209,10 @@ public function msetnx(array $array) { }
     /**
      * Remove specified keys.
      *
-     * @param   int | array $key1 An array of keys, or an undefined number of parameters, each a key: key1 key2 key3
-     *                            ... keyN
-     * @param   string      $key2 ...
-     * @param   string      $key3 ...
+     * @param   int|array|string    $key1 An array of keys, or an undefined number of parameters, each a key: key1 key2 key3
+     *                              ... keyN
+     * @param   string              $key2 ...
```
andrey-sokolov: It's better to avoid unnecessary formatting changes

redis/RedisCluster.php
```diff
@@ -209,10 +209,10 @@ public function msetnx(array $array) { }
     /**
      * Remove specified keys.
      *
-     * @param   int | array $key1 An array of keys, or an undefined number of parameters, each a key: key1 key2 key3
-     *                            ... keyN
-     * @param   string      $key2 ...
-     * @param   string      $key3 ...
+     * @param   int|array|string    $key1 An array of keys, or an undefined number of parameters, each a key: key1 key2 key3
```
andrey-sokolov:
```
/**
 * Remove specified keys.
 * @param int|string...|array[] $keys
 * @return int Number of keys deleted.
.............
 */
```` 
looks like it's a suitable signature for this case

File: .gitignore (hunk 1 above)
Comment (liamsorsby) [reply .gitignore]: Why do you commit the .idea directory?

File: redis/RedisCluster.php (hunk 3 above)
Comment (liamsorsby) [reply redis/RedisCluster.php]: I agree with the first part but why the braces after array?

File: redis/RedisCluster.php (hunk 2 above)
Comment (liamsorsby) [reply redis/RedisCluster.php]: 👌

File: redis/RedisCluster.php (hunk 3 above)
Comment (andrey-sokolov) [reply redis/RedisCluster.php]: @liamsorsby my mistake - without braces

File: .gitignore (hunk 1 above)
Comment (andrey-sokolov) [reply .gitignore]: it stores the project settings - like exclusion of vendor folders for example. We are open to any pull requests to it but it's better to do it in a separate PR

**Comment by liamsorsby** [comment]
@andrey-sokolov The changes should have been made 😄

File: redis/RedisCluster.php (hunk 3 above)
Comment (andrey-sokolov) [reply redis/RedisCluster.php]: Sorry for the long response. There is a problem with `string...` in docs and tests failed because of it. It's my mistake again, sorry. I believe that the whole signature of this method should be like this - we can delete the third parameter and mark the varidic part in the correct way.
```
    /**
     * Remove specified keys.
     *
     * @param int|string|array $key1 An array of keys, or an undefined number of parameters, each a key: key1 key2 key3
     *                            ... keyN
     * @param int|string ...$otherKeys
     *
     * @return int Number of keys deleted.
     * @link    https://redis.io/commands/del
     * @example
     * <pre>
     * $redisCluster->set('key1', 'val1');
     * $redisCluster->set('key2', 'val2');
     * $redisCluster->set('key3', 'val3');
     * $redisCluster->set('key4', 'val4');
     * $redisCluster->del('key1', 'key2');          // return 2
     * $redisCluster->del(array('key3', 'key4'));   // return 2
     * </pre>
     */
    public function del($key1, ...$otherKeys) { }

```

liamsorsby: Change it to the above.

liamsorsby:
@andrey-sokolov It should now be fixed. Sorry about that.

PR #529, opened 2019-03-05
Fix defaults for mt_rand

antoniopaisfernandes:


jdferreira:
I was about to submit this exact same PR, along with the better description and all! Thanks :)

antoniopaisfernandes:
:) glad to help!

standard/standard_2.php
```diff
@@ -993,7 +993,7 @@ function srand ($seed = null, $mode = MT_RAND_MT19937) {}
 function getrandmax () {}
 
 /**
- * Generate a better random value
+ * Generate a random value via the Mersenne Twister Random Number Generator
```
andrey-sokolov: Please, add description for parameters

andrey-sokolov: In this PR we are going to add default values, but according to description `max` default value is  `mt_getrandmax()` so it'll be better to mention it in docs

antoniopaisfernandes: Done

Virakal:
Just checking - shouldn't these be `@param int $min` and `@param int max` with type definitions?

antoniopaisfernandes:
Nice catch! Thanks

PR #530, opened 2019-03-05
Remove duplicate meta definition for htmlspecialchars

King2500:
`htmlspecialchars()` argument 1 is defined twice...


wbars:
Thank you, we will add the check to the tests to eliminate such issues in the future.

PR #533, opened 2019-03-06
Update http3.php

Leeo97one:
https://youtrack.jetbrains.com/issue/WI-45684


m6w6:
There have been lots of fixes. The current version generated from docs is available here:
* https://github.com/mdref/mdref-http

PR #536, opened 2019-03-13
Fix stub for Yaf 

wxxiong6:
Fix stub for Yaf


wxxiong6:
Fix stub for Yaf

wxxiong6:
Fix stub for Yaf

PR #538, opened 2019-03-14
Master

Huseriato:
Changed some more default parameters for PHP.


wbars:
Hello @Huseriato. Thank you for your first-time contribution, but it seems, that branches in this PR swapped by accident - we expect to merge something in `master`, not from it. Could you please compose valid PR with changes related only to your fix? 
For now I will close this one.

Huseriato:
I never worked with Git. I'm using SVN. I've no time to learn git, just to place a pull request here. May be you can fix it, just search for "[optional]" in the standard_?.php files and check the default parameters. There are 8 functions where it is missing. May be there are more i've not found using this method.

PR #547, opened 2019-03-28
Add PhpDoc for ENT constants

King2500:
*see changes*
Taken from https://www.php.net/manual/en/function.htmlspecialchars.php


standard/standard_defines.php
```diff
@@ -255,13 +255,33 @@
...
 define ('ENT_NOQUOTES', 0);
+
 /**
+ * Silently discard invalid code unit sequences instead of returning an empty string.
+ * Using this flag is discouraged as it may have security implications.
+ * @link https://php.net/manual/en/function.htmlspecialchars.php
  * @since 5.3.0
  */
 define ('ENT_IGNORE', 4);
+
```
wbars:
```suggestion
```

PR #549, opened 2019-03-28
Add expectedArguments common for dirname

King2500:
It's very common to have `__DIR__` and `__FILE__` as arguments for `dirname()`. 
(Getting current directory or parent directories)

This also uses `expectedReturnValues` as a trick, so `dirname()` gets listed when calling it nested:

`dirname(dirname(dirname(__DIR__)))`

image

Also you get completion in comparison for `__DIR__`:

```php
if (dirname($file) == <caret>) { ...
```

### DIFF
--- a/meta/.phpstorm.meta.php
+++ b/meta/.phpstorm.meta.php
@@ -104,6 +104,15 @@ namespace PHPSTORM_META {
     expectedArguments(\variant_cmp(), 0, NORM_IGNORECASE|NORM_IGNORENONSPACE|NORM_IGNORESYMBOLS|NORM_IGNOREWIDTH|NORM_IGNOREKANATYPE|NORM_IGNOREKASHIDA);
 
     expectedArguments(\DOMDocument::schemaValidateSource(), 1, LIBXML_SCHEMA_CREATE);
+    
+    registerArgumentsSet('common_dirname_constants', __DIR__, __FILE__);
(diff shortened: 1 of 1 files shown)


King2500:
From Travis log:
```
1) StubTests\StubsMetaExpectedArgumentsTest::testExpectedArgumentsAreUnique
Expected arguments for dirname with index 0 already registered
```
Intended in this special case. How can we ignore/circumvent it?

wbars:
@King2500 Registration of the same signature multiple times is indeed not expected and basically misusage.
On the other hand there is now another way to implement your case, probably we have to introduce ability to use `argumentsSet` along with another constants like this:
```
registerArgumentsSet('common_dirname_return', __DIR__);
expectedArguments(\dirname(), 0, argumentsSet('common_dirname_return'), __FILE__);
expectedReturnValues(\dirname(), argumentsSet('common_dirname_return'));
```

What do you think about such API?

King2500:
@wbars Sounds good!

King2500:
Should i create an issue on YouTrack?

wbars:
@King2500 Solution [is implemented](https://youtrack.jetbrains.com/issue/WI-46120) and will be available starting with 2019.2 EAP. Could you please update your PR according to the new API?

King2500:
@wbars Done so far :)

PR #552, opened 2019-04-01
search for NULL comments: via regex "(is|return).* NULL"

voku:
This change is [Reviewable](https://reviewable.io/reviews/jetbrains/phpstorm-stubs/552)

### DIFF
--- a/dom/dom_c.php
+++ b/dom/dom_c.php
@@ -50,7 +50,7 @@ class DOMNode  {
     public $childNodes;
 
     /**
-     * @var DOMNode
+     * @var DOMNode|null
      * @since 5.0
      * The first child of this node. If there is no such node, this returns NULL.
      * @link https://php.net/manual/en/class.domnode.php#domnode.props.firstchild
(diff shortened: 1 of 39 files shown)


dom/dom_c.php
```diff
@@ -1419,52 +1419,44 @@ class DOMElement extends DOMNode  {
...
      */
     public $nextSibling;
 
-    /**
-     * @var bool
-     * @since 5.0
-     * Not implemented yet, always return NULL
-     * @link https://php.net/manual/en/class.domelement.php#domelement.props.schematypeinfo
-     */
-    public $schemaTypeInfo ;
```
voku: doublicate entry, so I deleted one

libevent/libevent.php
```diff
@@ -602,7 +602,7 @@ function event_buffer_fd_set($bevent, $fd) {}
  * @param resource $bevent Valid buffered event resource.
  * @param callback|null $readcb Callback to invoke where there is data to read, or NULL if no callback is desired.
  * @param callback|null $writecb Callback to invoke where the descriptor is ready for writing, or NULL if no callback is desired.
- * @param callback|null $errorcb Callback to invoke where there is an error on the descriptor, cannot be NULL.
```
voku: "cannot be NULL"

PR #554, opened 2019-04-04
fix tables in date.php

bmewburn:


pestretsov:
Hello! Thanks for your contribution. Your changes definitely make sense, however they mainly affect our UI for Quick Documentation popup and window, which is not the main purpose of those stubs. Could you please create a [ticket at our tracker](https://youtrack.jetbrains.com/newIssue?project=WI) so we could address it in a separate manner?

PR #555, opened 2019-04-04
Fix ReflectionClass::getMethods

Izumi-kun:


pestretsov:
Thanks for your contribution!

PR #556, opened 2019-04-05
Update user_error stub to match trigger_error

leomoty:
`user_error` is lacking the default `E_USER_NOTICE` that `trigger_error` already has.


pestretsov:

PR #559, opened 2019-04-06
Fix xReadGroup phpdocs for count and block

alexander-schranz:


redis/Redis.php
```diff
@@ -3471,8 +3471,8 @@ public function xRead($arr_streams, $i_count = null, $i_block = null) {}
      * @param   string      $str_group
      * @param   string      $str_consumer
      * @param   array       $arr_streams
-     * @param   int|string  $i_count
-     * @param   int|string  $i_block
+     * @param   int|null    $i_count
```
pestretsov: Does it make sense to explicitly call `xReadGroup('mygroup', 'consumer2', ['s1' => 0, 's2' => 0], null, null)`? I suggest to remove nullability here if it doesn't

alexander-schranz: Removing the nullability would be false as null is allowed and can be set. Its important when you use static code analyzers as they will then not allow to have null there.

PR #561, opened 2019-04-09
Correct type declaration

chrisdevelops:
The type declaration on this argument should be int instead of array, unfortunately I ran into the following error:

```
PHP Fatal error:  Uncaught TypeError: Argument 1 passed to Parle\Parser::sigil() must be of the type integer, array given
```

I've submitted a patch to amend the documentation over at https://www.php.net/manual/en/parle-parser.sigil.php and https://www.php.net/manual/en/parle-rparser.sigil.php.

Please let me know if there are any issues.

### DIFF
--- a/Parle/Parser.php
+++ b/Parle/Parser.php
@@ -141,10 +141,10 @@ class Parser
      * This method is equivalent to the pseudo variable functionality in Bison.
      *
      * @link https://php.net/manual/en/parle-parser.sigil.php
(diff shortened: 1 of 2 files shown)


pestretsov:
Yes, this is a mistake ([more proof](https://github.com/weltling/parle/blob/7f6e64fc8164d6f3c8ec8923fe61c22c264c7368/parle.cpp#L1050)). Thanks for your contribution!

PR #562, opened 2019-04-11
Add support for zAdd options like XX|NX and so on

lovelock:
https://github.com/phpredis/phpredis/issues/760
phpredis now support the options but the stubs not

### DIFF
--- a/redis/Redis.php
+++ b/redis/Redis.php
@@ -2319,7 +2319,7 @@ class Redis
      * @example
      * <pre>
      * <pre>
-     * $redis->('z', 1, 'v1', 2, 'v2', 3, 'v3', 4, 'v4' );  // int(2)
+     * $redis->zAdd('z', 1, 'v1', 2, 'v2', 3, 'v3', 4, 'v4' );  // int(2)
      * $redis->zRem('z', 'v2', 'v3');                           // int(2)
      * $redis->zAdd('z', ['NX'], 5, 'v5');                      // int(1)
      * $redis->zAdd('z', ['NX'], 6, 'v5');                      // int(0)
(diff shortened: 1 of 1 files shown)


redis/Redis.php
```diff
@@ -2306,30 +2306,45 @@ public function brpoplpush( $srcKey, $dstKey, $timeout ) {}
...
+     * @param   string  $value2  Optional value
+     * @param   float   $scoreN  Optional score
+     * @param   string  $valueN  Optional value
      * @return  int     Number of values added
      * @link    https://redis.io/commands/zadd
      * @example
      * <pre>
      * <pre>
-     * $redis->zAdd('z', 1, 'v2', 2, 'v2', 3, 'v3', 4, 'v4' );  // int(2)
+     * $redis->('z', 1, 'v1', 2, 'v2', 3, 'v3', 4, 'v4' );  // int(2)
```
King2500: `->(` is that correct syntax?

lovelock: Ahh thanks, I've fixed a typo and introduced another one.

redis/Redis.php
```diff
@@ -2306,30 +2306,45 @@ public function brpoplpush( $srcKey, $dstKey, $timeout ) {}
...
      * Adds the specified member with a given score to the sorted set stored at key.
      *
-     * @param   string  $key    Required key
-     * @param   float   $score1 Required score
-     * @param   string  $value1 Required value
-     * @param   float   $score2 Optional score
-     * @param   string  $value2 Optional value
-     * @param   float   $scoreN Optional score
-     * @param   string  $valueN Optional value
+     * @param   string  $key     Required key
```
pestretsov: Please, revert the formatting, so the only change in `@params` block is the added `@param array $options Options if needed`. We [do this to preserve git history](https://github.com/JetBrains/phpstorm-stubs/blob/master/CONTRIBUTING.md#code-style).

lovelock:
Thanks for your review and the format has been reverted.

pestretsov:
Thank you!

PR #563, opened 2019-04-11
Add expectedArguments for socket transports

King2500:
*see changes*

### DIFF
--- a/meta/.phpstorm.meta.php
+++ b/meta/.phpstorm.meta.php
@@ -136,6 +136,12 @@ namespace PHPSTORM_META {
 	expectedArguments(\fopen(), 1, 'r', 'r+', 'w', 'w+', 'a', 'a+', 'x', 'x+', 'c', 'c+', 'e');
 	expectedArguments(\popen(), 1, 'r', 'r+', 'w', 'w+', 'a', 'a+', 'x', 'x+', 'c', 'c+', 'e');
 	expectedArguments(\SplFileInfo::openFile(), 0, 'r', 'r+', 'w', 'w+', 'a', 'a+', 'x', 'x+', 'c', 'c+', 'e');
+    
+    registerArgumentsSet('socket_transports',
+        'tcp://', 'udp://', 'ssl://', 'sslv2://', 'sslv3://', 'tls://', 'unix://', 'udg://'
+    );
+    expectedArguments(\fsockopen(), 0, argumentsSet('socket_transports'));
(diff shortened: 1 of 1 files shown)


meta/.phpstorm.meta.php
```diff
@@ -136,6 +136,12 @@ function argumentsSet($setName) {
 	expectedArguments(\fopen(), 1, 'r', 'r+', 'w', 'w+', 'a', 'a+', 'x', 'x+', 'c', 'c+', 'e');
 	expectedArguments(\popen(), 1, 'r', 'r+', 'w', 'w+', 'a', 'a+', 'x', 'x+', 'c', 'c+', 'e');
 	expectedArguments(\SplFileInfo::openFile(), 0, 'r', 'r+', 'w', 'w+', 'a', 'a+', 'x', 'x+', 'c', 'c+', 'e');
+    
+    registerArgumentsSet('socket_transports',
+        'tcp://', 'udp://', 'ssl://', 'sslv2://', 'sslv3://', 'tls://', 'unix://', 'udg://'
+    );
+    expectedArguments(\fsockopen(), 0, argumentsSet('socket_transports'));
```
wbars: So this one is actually misusage  - we want to provide actual expected arguments rather then help to complete part of the some of possible values.

King2500: I understand.
But why not allow something like this to help with prefixes and similar?

King2500:
It is true that these completion make only sense inside string literal `fsockopen('<caret>')` because you have to write down the rest of the the URL: `fsockopen('tcp://<caret>')`.

When completion here `fsockopen(<caret)` you may either use variable (standard PHP completion) or start writing literal with quote char `'`.

meta/.phpstorm.meta.php
wbars: We want to extend the feature in the future and this concept contradicts it's purpose. It should be implemented separately by another instruments.

wbars:
Thank you for paying attention to such possible use-case. Unfortunately, `expectedArguments`  feature was not designed to support such case and expanding it in this way will break future plans of it's development.

I will close the PR, but it will be kept there as useful reference for the feature.

King2500:
I have reimplemented this in [PHP Advanced AutoComplete](https://github.com/King2500/idea-php-advanced-autocomplete) plugin:
https://github.com/King2500/idea-php-advanced-autocomplete/commit/0f3ccf204ac5fc5cfa230250966626570045ad75

PR #565, opened 2019-04-11
Add expectedArguments for streams in I/O functions

King2500:
expectedArguments completion for PHP standard protocols and stream wrappers
see: https://www.php.net/manual/de/wrappers.php
and: https://www.php.net/manual/de/wrappers.php.php

Some are only readable and some only writable.

### DIFF
--- a/meta/.phpstorm.meta.php
+++ b/meta/.phpstorm.meta.php
@@ -136,6 +136,25 @@ namespace PHPSTORM_META {
 	expectedArguments(\fopen(), 1, 'r', 'r+', 'w', 'w+', 'a', 'a+', 'x', 'x+', 'c', 'c+', 'e');
 	expectedArguments(\popen(), 1, 'r', 'r+', 'w', 'w+', 'a', 'a+', 'x', 'x+', 'c', 'c+', 'e');
 	expectedArguments(\SplFileInfo::openFile(), 0, 'r', 'r+', 'w', 'w+', 'a', 'a+', 'x', 'x+', 'c', 'c+', 'e');
+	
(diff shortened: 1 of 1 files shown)


meta/.phpstorm.meta.php
```diff
@@ -136,6 +136,25 @@ function argumentsSet($setName) {
...
+	expectedArguments(\file_get_contents(), 0, 'php://stdin', 'php://input', 'php://fd/', 'php://temp', 'php://temp/maxmemory=', 'php://filter/resource=', 'php://filter/read=', 'file://', 'http://', 'ftp://', 'ftps://', 'zip://', 'compress.zlib://', 'compress.bzip2://', 'data://text/plain,', 'data://text/plain;base64,', 'phar://', 'ssh2.shell://', 'ssh2.exec://', 'ssh2.tunnel://', 'ssh2.sftp://', 'ssh2.scp://', 'rar://', 'ogg://', 'expect://');
+	expectedArguments(\file_put_contents(), 0, 'php://stdout', 'php://stderr', 'php://output', 'php://fd/', 'php://temp', 'php://temp/maxmemory=', 'php://filter/resource=', 'php://filter/write=', 'file://', 'ftp://', 'ftps://', 'compress.zlib://', 'compress.bzip2://', 'phar://', 'ssh2.shell://', 'ssh2.exec://', 'ssh2.tunnel://', 'ssh2.sftp://', 'ogg://', 'expect://');
+	expectedArguments(\DirectoryIterator::__construct(), 0, 'glob://', 'rar://', 'phar://');
+	expectedArguments(\RecursiveDirectoryIterator::__construct(), 0, 'glob://', 'rar://', 'phar://');
+	expectedArguments(\glob(), 0, 'phar://');
+
+	expectedArguments(\fwrite(), 0, STDOUT, STDERR);
```
wbars: Same, this line will means that basically it's not expected to see anything except `STDOUT` or `STDERR` which is not true.

King2500: I understand. Sorry.

King2500: I thought expectedArguments mean it is just a "weighted" suggestion and does not prevent you using other constants and scalars as well...

King2500:
For string literals see my comment here: https://github.com/JetBrains/phpstorm-stubs/pull/563#issuecomment-483331649

wbars:
Thomas, I will close this PR for the same reasons as #563. Thank you again for paying attention for such cases.

PR #567, opened 2019-04-15
Remove possible formatting left-overs from PhpDocs

King2500:
Is there any reason for these strange tags?
Otherwise we can remove them, because they show up in PhpStorm Quick Doc:

image

### DIFF
--- a/standard/standard_4.php
+++ b/standard/standard_4.php
@@ -27,7 +27,6 @@ function error_get_last () {}
  * Note that the parameters for call_user_func are
  * not passed by reference.
  * call_user_func example and references
- * ]]>
  * &example.outputs;
  * </p>
  * @return mixed the function result, or false on error.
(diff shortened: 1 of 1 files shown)


standard/standard_4.php
```diff
@@ -1161,4 +1151,4 @@ function getmxrr ($hostname, array &$mxhosts, array &$weight = null) {}
  * </table>
  * @since 5.0
  */
-function dns_get_record ($hostname, $type = DNS_ANY, array &$authns = null, array &$addtl = null, &$raw = false) {}
\ No newline at end of file
+function dns_get_record ($hostname, $type = DNS_ANY, array &$authns = null, array &$addtl = null, &$raw = false) {}
```
pestretsov: Please, revert formatting here. The reason is that we want to preserve changes history

King2500: It looks like GitHub is inserting this automatically :(

PR #569, opened 2019-04-15
Fix PhpDocs for standard defines

King2500:
*see changes*

### DIFF
--- a/standard/standard_defines.php
+++ b/standard/standard_defines.php
@@ -136,26 +136,30 @@ define ('INF', INF);
 define ('NAN', NAN);
 
 /**
- * Round halves up [5.3.0]
+ * Round halves up
  * @link https://php.net/manual/en/math.constants.php
+ * @since 5.3.0
  */
 define ('PHP_ROUND_HALF_UP', 1);
 
 /**
- * Round halves down [5.3.0]
+ * Round halves down
  * @link https://php.net/manual/en/math.constants.php
+ * @since 5.3.0
  */
 define ('PHP_ROUND_HALF_DOWN', 2);
 
 /**
- * Round halves to even numbers [5.3.0]
+ * Round halves to even numbers
  * @link https://php.net/manual/en/math.constants.php
+ * @since 5.3.0
  */
 define ('PHP_ROUND_HALF_EVEN', 3);
 
 /**
- * Round halves to odd numbers [5.3.0]
+ * Round halves to odd numbers
  * @link https://php.net/manual/en/math.constants.php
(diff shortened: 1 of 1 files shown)


standard/standard_defines.php
```diff
@@ -911,165 +910,130 @@
...
+define('IMAGETYPE_COUNT', 18);
 
 /**
  * IPv4 Address Resource
  * @link https://php.net/manual/en/network.constants.php
  */
 define ('DNS_A', 1);
 
-define('DNS_CAA', 8192);
+define ('DNS_CAA', 8192);
```
pestretsov: Please, revert formatting here. The reason is that we want to preserve changes history.

standard/standard_defines.php
```diff
@@ -911,165 +910,130 @@
...
-define ('IMAGETYPE_COUNT', 18);
 
 /**
- * Image type constant used by the image_type_to_mime_type() and image_type_to_extension() functions
+ * Image type constant used by the {@link image_type_to_mime_type()} and {@link image_type_to_extension()} functions.
  * @link https://php.net/manual/en/image.constants.php
+ * @since 7.1.0
  */
 define('IMAGETYPE_WEBP', 18);
+define('IMAGETYPE_UNKNOWN', 0);
```
pestretsov: Why move those?

King2500: Logical order: _actual image types_ -> _others (unknown/count)_

I can revert if it's a problem here.

pestretsov: I guess it's better to revert (if the logical order is the only reason), because our stubs are not intended to be read by users:)

King2500: I tell you from daily experience with PhpStorm:

My collegues (including myself from time to time) tend to Ctrl+Click on standard PHP identifiers to go to declaration instead of using Ctrl+Q for Quick Definition.
So we end up in stubs file looking at the "code" and keep looking around. It just helps a little if things are ordered somehow there to see "what else" is defined for example with `IMAGETYPE_`.

King2500: I claim that more people are looking into stubs files then people are watching the git history of them on Github or wherever else you see it.

pestretsov: Okay, I see. We'll take it into consideration then. Still, please, resolve my first remark and I'll merge your changes :)

standard/standard_defines.php
King2500: Done.

pestretsov:
@King2500 thank you!

PR #570, opened 2019-04-18
SimpleXML: better return annotation for xpath

quasilyte:
If `xpath` is called from `SimpleXMLIterator`, which extends `SimpleXMLElement`, the result type of `xpath` would be `SimpleXMLIterator[]`, not `SimpleXMLElement[]`. Currently, Phpstorm would report non-existing `current`/`next`/`rewind` methods that are called on `xpath` results even if the underlying type is `SimpleXMLIterator` (which has those methods). A simplest fix is to annotate return type as `static[]`, because this way Phpstorm is able to infer proper type of the elements and stops making false reports about undefined methods.

Simple example that demonstrates that:

```php
<?php

$strXml= <<<XML
<a>
 <b>
  <c>text</c>
  <c>stuff</c>
 </b>
 <d>
  <c>code</c>
 </d>
</a>
XML;

// $xml = new SimpleXMLElement($strXml, 0);
// $root = $xml->xpath("/a");
// var_dump($root);

$xml = new SimpleXMLIterator($strXml, 0);
$root = $xml->xpath("/a");
var_dump($root);

$iter = $root[0];
$iter->rewind();
$iter->next();
var_dump($iter->current());
```

### DIFF
--- a/SimpleXML/SimpleXML.php
+++ b/SimpleXML/SimpleXML.php
@@ -66,7 +66,7 @@ class SimpleXMLElement implements Traversable, ArrayAccess, Countable {
 	 * @param string $path <p>
 	 * An XPath path
 	 * </p>
-	 * @return SimpleXMLElement[] an array of SimpleXMLElement objects or <b>FALSE</b> in
+	 * @return static[] an array of SimpleXMLElement (or derived) objects or <b>FALSE</b> in
 	 * case of an error.
 	 * @since 5.2.0
 	 */


pestretsov:
@quasilyte thanks for your contribution!

MaXal:
I suggest reverting the PR until https://youtrack.jetbrains.com/issue/WI-28728 is fixed. Otherwise, it does more harm than good:
```
<?php

$xml    = simplexml_load_file("file.xml");
$result = $xml->xpath("foo");
foreach ($result as $element)
{
	$result2 = $element->xpath("bar");
	foreach ($result2 as $element2)
	{
		$result3 = $element->xpath("f");
		foreach ($result3 as $element3)
		{
			$element3->xpath(); //is undefined
		}
	}
}
```

We've got quite a lot of new false-positives on the most popular framework after this change.

quasilyte:
I wasn't aware of that issue. Thanks.

PR #572, opened 2019-04-18
Add GitHub templates

SerafimArts:
1) Adding a template will allow new or inexperienced participants to formulate their problem more easily.
2) Having a description of the problem to be solved will allow reviewers to quickly understand what the participant is trying to do with the help of this PR (or issue).
3) Well, in general, these are the rules of good form =)

More about the templates here: https://help.github.com/en/articles/about-issue-and-pull-request-templates

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+/.github export-ignore
+.travis.yml export-ignore
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,9 @@
+### What is a problem?
+
+// Please briefly describe the problem you encountered.
+
+### How is this problem resolved?
+
+// Please describe the steps you can take to solve this problem.
+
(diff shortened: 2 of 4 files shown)


voku:
We could also add different issue templates e.g. https://github.com/kalessil/phpinspectionsea/tree/master/.github/ISSUE_TEMPLATE/

SerafimArts:
However, this repository does not contain `issues` section, so even my template addition looks rather dubious

King2500:
I don't see an issue tab at this GitHub repository... ?

SerafimArts:
@King2500 issues is an illusion :sparkles:

wbars:
Thank you for bringing attention to the problem. Right now we're in process of composing GitHub templates, which will make contribution process easier for both us and contributors. 

Unfortunately, suggested template are quite general and I don't see how they can improve process of contributing, so I will close this PR. 

Thank you again for pointing this out.

PR #575, opened 2019-04-25
Applied optional second parameter for `ReflectionMethod::__construct`

webmaster777:
Adapted docs to match php doc


pestretsov:
@webmaster777 hi! Thanks for your contribution. Although you're correct that current php doc doesn't match the [official docs](https://php.net/manual/en/reflectionmethod.construct.php) and your change doesn't break anything, it was intentionally changed to be that way (https://youtrack.jetbrains.com/issue/WI-21227). I'll close the PR in favour of current version because it is a bit better – the declared type of needed argument in both `__construct` signatures is `string`.

PR #578, opened 2019-04-26
Add stubs for pecl-stomp php extension

shakura:
https://www.php.net/manual/en/book.stomp.php
issue - https://youtrack.jetbrains.com/issue/WI-46583


pestretsov:
@shakura thank you very much!

PR #579, opened 2019-04-30
MongoDB: Timestamp needs to implement TimestampInterface

timwhitlock:
https://php.net/manual/en/class.mongodb-bson-timestamp.php


wbars:
Thank you.

PR #580, opened 2019-05-05
Redis::info phpdoc

shuchkin:
"return string" to "return string|array"

### DIFF
--- a/redis/Redis.php
+++ b/redis/Redis.php
@@ -2132,7 +2132,7 @@ class Redis
      * - vm_enabled
      * - role
      * @link    https://redis.io/commands/info
-     * @return string
+     * @return string|array
      * @example
      * <pre>
      * $redis->info();


redis/Redis.php
```diff
@@ -2132,7 +2132,7 @@ public function sort( $key, $option = null ) {}
      * - vm_enabled
      * - role
      * @link    https://redis.io/commands/info
-     * @return string
+     * @return string|array
```
wbars:
```suggestion
     * @return array
```
According to the [docs](https://github.com/phpredis/phpredis#info) and to the [sources](https://github.com/phpredis/phpredis/blob/d7450b2f59700e4662624317438c456a0dd36165/library.c#L853) this method can return only `array`, not `string`.

wbars:
Hello.

Since this PR did't get any response in quite a while, I will close it to keep our PR list actual. Feel free to reopen it, if you will update it.

PR #581, opened 2019-05-06
make strftime() $timestamp optional

tjallingt:
I think this should be optional and default to the value returned by `time()`: https://www.php.net/manual/en/function.strftime.php

### DIFF
--- a/date/date.php
+++ b/date/date.php
@@ -769,7 +769,7 @@ function checkdate ($month, $day, $year) {}
  * @since 4.0
  * @since 5.0
  */
-function strftime ($format, $timestamp) {}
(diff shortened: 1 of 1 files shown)


date/date.php
```diff
@@ -769,7 +769,7 @@ function checkdate ($month, $day, $year) {}
  * @since 4.0
  * @since 5.0
  */
-function strftime ($format, $timestamp) {}
+function strftime ($format, $timestamp = 'time()') {}
```
wbars:
```suggestion
function strftime ($format, $timestamp = null) {}
```

tjallingt:
makes complete sense, i just copied it from the `strtotime` function since it seemed to do the same thing 😅

PR #585, opened 2019-05-09
Move AssertionError to right extension

kukulich:
https://3v4l.org/4N3Ig

### DIFF
--- a/Core/Core_c.php
+++ b/Core/Core_c.php
@@ -483,15 +483,6 @@ class ParseError extends CompileError {
 }
 
 /**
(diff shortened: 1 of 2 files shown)


Core/Core_c.php
```diff
@@ -621,7 +612,7 @@ static function bind(Closure $closure, $newthis, $newscope = 'static') { }
      * @since 7.0
      */
     function call ($newThis, ...$parameters) {}
-
```
wbars: Please revert unrelated formatting changes.

kukulich: Done

PR #586, opened 2019-05-10
added missing com_exception to com_dotnet stub

dktapps:
This class doesn't have a manual page in the traditional format (see https://php.net/manual/en/com.error-handling.php).

Several other classes appear when writing `php --re com_dotnet`, but I did not take the time to investigate those.

### DIFF
--- a/com_dotnet/com_dotnet.php
+++ b/com_dotnet/com_dotnet.php
@@ -73,6 +73,12 @@ class VARIANT {
 }
 
 /**
+ * This extension will throw instances of the class com_exception whenever there is a potentially fatal error reported by COM. All COM exceptions have a well-defined code property that corresponds to the HRESULT return value from the various COM operations. You may use this code to make programmatic decisions on how to handle the exception.
+ * @link https://php.net/manual/en/com.error-handling.php
+ */
+class com_exception extends \Exception {}
+
+/**
  * (PHP 5, PHP 7)<br/>
  * Generate a globally unique identifier (GUID)
  * @link https://php.net/manual/en/function.com-create-guid.php


wbars:
> All COM exceptions have a well-defined code property

Am I get it correctly that introduced exception class should have `$property` field. If so, maybe we will add it as well?

dktapps:
According to `php --re com_dotnet`, there is not:
```
    Class [ <internal:com_dotnet> final class com_exception extends Exception implements Throwable ] {

      - Constants [0] {
      }

      - Static properties [0] {
      }

      - Static methods [0] {
      }

      - Properties [4] {
        Property [ <default> protected $message ]
        Property [ <default> protected $code ]
        Property [ <default> protected $file ]
        Property [ <default> protected $line ]
      }

      - Methods [10] {
        Method [ <internal:Core, inherits Exception, ctor> public method __construct ] {

          - Parameters [3] {
            Parameter #0 [ <optional> $message ]
            Parameter #1 [ <optional> $code ]
            Parameter #2 [ <optional> $previous ]
          }
        }

        Method [ <internal:Core, inherits Exception> public method __wakeup ] {
        }

        Method [ <internal:Core, inherits Exception, prototype Throwable> final public method getMessage ] {
        }

        Method [ <internal:Core, inherits Exception, prototype Throwable> final public method getCode ] {
        }

        Method [ <internal:Core, inherits Exception, prototype Throwable> final public method getFile ] {
        }

        Method [ <internal:Core, inherits Exception, prototype Throwable> final public method getLine ] {
        }

        Method [ <internal:Core, inherits Exception, prototype Throwable> final public method getTrace ] {
        }

        Method [ <internal:Core, inherits Exception, prototype Throwable> final public method getPrevious ] {
        }

        Method [ <internal:Core, inherits Exception, prototype Throwable> final public method getTraceAsString ] {
        }

        Method [ <internal:Core, inherits Exception, prototype Throwable> public method __toString ] {
        }
      }
    }
```

wbars:
I'm sorry I meant `code` property. Anyway, these properties can be added as separate PR.

dktapps:
@wbars It's inherited from the base `Exception` class, so no action is necessary.

PR #588, opened 2019-05-14
xmlrpc_server_register_method: use callable type

glensc:
xmlrpc_server_register_method third argument is any kind of callable:
-  `array($bar, "foo_func")`


glensc:
thanks!

PR #589, opened 2019-05-14
Fixed GMP stub for PHP >= 5.6

dktapps:
GMP transitioned to objects in PHP 5.6, which was not reflected by the stub, causing false positives wherever it was used.


timmyrs:
Is there any ETA when this month-old fix will finally be available in PhpStorm?

PR #591, opened 2019-05-15
Fixed names of TRUE, FALSE and NULL constants

kukulich:
See https://3v4l.org/A6El3

### DIFF
--- a/Core/Core_d.php
+++ b/Core/Core_d.php
@@ -295,9 +295,9 @@ define ('LOG_NOWAIT', 16);
  * @link https://php.net/manual/en/network.constants.php
  */
 define ('LOG_PERROR', 32);
-define ('true', true, true);
-define ('false', false, true);
-define ('null', null, true);
+define ('TRUE', true, true);
+define ('FALSE', false, true);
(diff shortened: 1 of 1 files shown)


b1rdex:
Actually this 3 ones are magic in php. You can use them in any case (e.g. `tRuE` or `nuLL`).
I'm personally 👎 on this change.

kukulich:
@b1rdex Yes, it works because of the third parametr of function `define`. However the right names are uppercased.

wbars:
Thank you for the investigation. 
Although formally you are correct, I agree with @b1rdex at this one - these constants are case-insensitive. 
Also I don't see how this change will improve PhpStorm code analysis.

kukulich:
@wbars This PR (and my others) helps using this repository as composer package, mostly currently in https://github.com/Roave/BetterReflection

BetterReflection is used in https://github.com/humbug/php-scoper
PHP-Scoper is used in https://github.com/humbug/box

And Box is used to creating PHAR for tools like PHPUnit, PHPStan etc.

So it will really help PHP ecosystem :)

PR #596, opened 2019-05-18
added pecl-crypto stub

dktapps:
adapted from https://github.com/bukka/php-crypto/blob/5f26ac91b0ba96742cc6284cd00f8db69c3788b2/docs/Crypto.php

### DIFF
--- a/crypto/crypto.php
+++ b/crypto/crypto.php
@@ -1,0 +1,781 @@
+<?php
+
+namespace Crypto{
+
+/**
+ * Class providing cipher algorithms
+ */
+class Cipher {
+    const MODE_ECB = 1;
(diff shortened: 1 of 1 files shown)


MaXal:
For the future, please avoid using expressions in default values since it's invalid code:
`public static function seed($buf, $entropy = (float) strlen($buf)) {}`

Please use instead something like:
```
 /**
         * Mixes bytes in $buf into PRNG state
         * @param string $buf
         * @param float $entropy [optional] The default value is (float) strlen($buf)
         * @return null
         */
        public static function seed($buf, $entropy) {}
```

dktapps:
I didn't notice that. It was in the original stub from ext-crypto's repository.

PR #602, opened 2019-05-23
Add MDN link to PhpDocs of HTTP headers related functions

King2500:
I think this is a good place to link a well-documented page where developers get more information about HTTP headers, when using `header` function and similar.


wbars:
Hello, thank you for your contribution.

While the information in the links is quite useful, I think that phpdocs is a not best place to place them. We're trying to keep only PHP-related information in PHPDocs.

King2500:
@wbars why?

King2500:
I see this from developer experience point of view not "PHP language only".

PR #605, opened 2019-05-25
Improve curl_setopt(...) completion performance

mihaelartz:


wbars:
Hello. Thank you for you contribution. As far as I can see, documentation of this function is not redundant and matches existing one from php.net

On the other hand, if you experience performance problems, please submit them into [our issue tracker ](https://youtrack.jetbrains.com/newIssue?project=WI)

PR #607, opened 2019-05-28
Add ReflectionReference

ntzm:


wbars:
Also please add documentation pieces for the methods from RFC. e.g.
`Returns ReflectionReference if array element is a reference, null otherwise`
`Returns unique identifier for the reference. The return value format is unspecified`

ntzm:
Updated

PR #615, opened 2019-06-10
DOM: document nullable return types & properties

BenMorel:
This is in line with other stubs.

### DIFF
--- a/dom/dom_c.php
+++ b/dom/dom_c.php
@@ -1978,7 +1978,7 @@ class DOMNotation  extends DOMNode{
 class DOMEntity extends DOMNode  {
 
     /**
-     * @var string
+     * @var string|null
      * @since 5.0
      * The public identifier associated with the entity if specified, and NULL otherwise.
      * @link https://php.net/manual/en/class.domentity.php#domentity.props.publicid
@@ -1986,7 +1986,7 @@ class DOMEntity extends DOMNode  {
     public $publicId ;
 
     /**
-     * @var string
+     * @var string|null
      * @since 5.0
      * The system identifier associated with the entity if specified, and NULL otherwise. This may be an
      * absolute URI or not.
@@ -1995,7 +1995,7 @@ class DOMEntity extends DOMNode  {
     public $systemId ;
 
     /**
-     * @var string
+     * @var string|null
      * @since 5.0
      * For unparsed entities, the name of the notation for the entity. For parsed entities, this is NULL.
      * @link https://php.net/manual/en/class.domentity.php#domentity.props.notationname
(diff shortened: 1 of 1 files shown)


dom/dom_c.php
```diff
@@ -574,12 +574,12 @@ class DOMDocument extends DOMNode  {
...
 
     /**
-     * @var string
+     * @var string|null
      * @since 5.0
      * The location of the document or NULL if undefined.
      * @link https://php.net/manual/class.domdocument.php#domdocument.props.documenturi
      */
-    public $documentURI ;
+    public $documentURI;
```
AlexMovsesov:
```suggestion
    public $documentURI ;
```
This formatting change is unrelated to the fix

dom/dom_c.php
```diff
@@ -885,7 +885,7 @@ public function getElementsByTagNameNS ($namespaceURI, $localName) {}
      * @param string $elementId <p>
      * The unique id value for an element.
      * </p>
-     * @return DOMElement the DOMElement or &null; if the element is
+     * @return DOMElement|null The DOMElement or &null; if the element is
```
AlexMovsesov:
```suggestion
     * @return DOMElement|null the DOMElement or &null; if the element is
```

dom/dom_c.php
```diff
@@ -1520,7 +1520,7 @@ public function getAttributeNode ($name) {}
      * @param DOMAttr $attr <p>
      * The attribute node.
      * </p>
-     * @return DOMAttr old node if the attribute has been replaced or &null;.
+     * @return DOMAttr|null Old node if the attribute has been replaced or &null;.
```
AlexMovsesov:
```suggestion
     * @return DOMAttr|null old node if the attribute has been replaced or &null;.
```

dom/dom_c.php
BenMorel: I took the opportunity to fix this (in addition to the 2 formatting changes below) because they were on the same line, and make the code more consistent with the rest of the stubs. If someone from JetBrains wants to keep formatting changes in another PR, I'll happily accept your changes, otherwise I think it's doing its bit to improve the stub quality.

AlexMovsesov: Ok, let's leave it as it is, thx for your contribution. I suggest to add nullability to the following properties: DomNode.parentNode (with corresponding documentation update), DomNode.firstChild, DomNode.lastChild, DomNode.previousSibling, DomNode.nextSibling and the same properties for the DOMElement

BenMorel: Very good point! I updated these properties in `DOMNode` + `DOMElement`, and also for [`DOMNode::$ownerDocument`](https://www.php.net/manual/en/class.domnode.php#domnode.props.ownerdocument). @AlexMovsesov do you work for JetBrains?

AlexMovsesov: Thx, yes, I am working in JetBrains

PR #619, opened 2019-06-19
fixing __soapCall() method signature

delboy1978uk:
Storm complains when the correct method signature is in place. Using the stub's signature throws a warning in the `error_log`, so I updated the method signature.

Please see https://youtrack.jetbrains.com/issue/WI-47198


AlexMovsesov:
Hello, thx for your contribution!

delboy1978uk:
Thanks for the speedy merge! :-D

PR #623, opened 2019-06-24
Actualize FFI interface using latest API info

SerafimArts:


SerafimArts:
Thx @AlexMovsesov 

This is actually a more correct option, because the second parameter does not allow null (https://github.com/dstogov/php-ffi/blob/master/ffi.c#L4043).

PR #631, opened 2019-07-07
ReflectionType::getName() should be ReflectionNamedType::getName()

Majkl578:
`ReflectionType::getName()` method doesn't exist [1][2], it's ReflectionNamedType::getName() instead [3].
[1] https://php.net/reflectionType
[2] https://3v4l.org/EgCmC
[3] https://www.php.net/manual/en/reflectionnamedtype.getname.php

### DIFF
--- a/Reflection/Reflection.php
+++ b/Reflection/Reflection.php
@@ -1999,15 +1999,6 @@ class ReflectionType
 	{
 	}
 
-	/**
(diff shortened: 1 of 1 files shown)


ondrejmirtes:
It's important to mention that `ReflectionNamedType` is the only subclass of `ReflectionType` right now so in practice this method is currently always there.

PR #632, opened 2019-07-09
json encode/decode throws JsonException when JSON_THROW_ON_ERROR is set

b1rdex:


MaXal:
Sorry but we had to revert the change see https://github.com/JetBrains/phpstorm-stubs/pull/634

PR #635, opened 2019-07-11
Fix return types of DateTimeImmutable methods

dtakken:
Change the return types of several methods of the DateTimeImmutable class. Most of these methods currently use `static` as return type hint, which does not work well when chaining several method calls on a DateTimeImmultable instance.

In case this PR is indeed the proper way to fix this problem, then the same fix could be applied to the DateTime class. It displays the same problem. I can extend the PR if desired.

### DIFF
--- a/date/date_c.php
+++ b/date/date_c.php
@@ -115,7 +115,7 @@ class DateTimeImmutable implements DateTimeInterface {
      * (PHP 5 &gt;=5.5.0)<br/>
      * Adds an amount of days, months, years, hours, minutes and seconds
      * @param DateInterval $interval
-     * @return static
+     * @return \DateTimeImmutable
      */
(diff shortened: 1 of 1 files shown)


b1rdex:
Could you please provide example code with the problem you're trying to fix? It looks like some PhpStorm side bug to me… `static` should work as expected even when `chaining several methods calls on a DateTimeImmultable instance`.

dtakken:
You appear to be right, this looks like a PHPStorm bug. I tried to isolate the problem but did not manage to do so. Whatever the problem is, replacing `static` with `DateTimeImmutable` in the stub works around it.

Should I close this PR?

Leneshka-jb:
@dtakken Please provide the problem once you'll figure it out. It's better to fix it in PhpStorm than change all `static` to exact types.

PR #637, opened 2019-07-15
array_unshift() has not been introduced in PHP 7.3

BenMorel:
[array_unshift()](https://www.php.net/manual/fr/function.array-unshift.php) has been there since PHP 4.

It's only the possibility of calling the function with a single argument that has been introduced in PHP 7.3.


BenMorel:
@Leneshka-jb done!

PR #640, opened 2019-07-24
RedisArray->info() does not accept any parameter

CyberLine:


Leneshka-jb:
Actually `https://redis.io/commands/info` provides way more keys. Also does it actually return string or associative array?

bzixilu:
@CyberLine Hi, any updates regarding the question asked above?

CyberLine:
I have added the missing info values and return types

bzixilu:
Thank you for your contribution, merged!

PR #645, opened 2019-08-08
Fixed stub for socket_export_stream function

Frago9876543210:
When you import this function using
```php
use function socket_export_stream;
```
PhpStorm says this:
изображение


bzixilu:
FYI, socket_export_stream may also return false in the case of failure - 84c7a76d28ea91318b0009e383cf056475498435

PR #650, opened 2019-08-13
[mongodb] Fix CommandStartedEvent::getCommand() return type

Lctrs:
According to https://www.php.net/manual/en/mongodb-driver-monitoring-commandstartedevent.getcommand.php, it returns an object


bzixilu:

PR #653, opened 2019-08-22
mysql_xdevapi_8_0_11

mhgs:
Updated mysql_xdevapi stub based on official documentation (version 8.0.11)


mhgs:
@bzixilu I see the build has failed, however I cannot see why? If there is a issue with the changes I would like to fix it of course. Can you point me in the right direction?

soulshockers:
> @bzixilu I see the build has failed, however I cannot see why? If there is a issue with the changes I would like to fix it of course. Can you point me in the right direction?

The commited stub map is not up to date. Please regenerate it using ./generate-stub-map

mhgs:
@soulshockers Thanks, that fixed it.

bzixilu:
The tests passed successfully against https://github.com/nexilo/xdevapi-php/blob/master/docker/php/Dockerfile

bzixilu:
Thank you for the contribution!

PR #654, opened 2019-08-24
Mute parameter mismatch failures for strtr function

soulshockers:
Mute parameter mismatch failures for strtr function


bzixilu:
Thank you for contributing!

PR #665, opened 2019-10-02
add tests to check deprecated/since versions 

isfedorov:
check that deprecated/since versions don't end with redundant .0 like 7.2.0


MaXal:
Let's extend the test a bit further:)

Please create a PHP array that reflects PhpStorm mapping: folder => extension type, test that "core" PHP doesn't have `@since/@deprecated` that is not equal to PHP version.

"Core" PHP is everything except PECL.

tests/TestData/Providers/PhpCoreStubsProvider.php
```diff
@@ -0,0 +1,153 @@
...
+            'yaml',
+            'yaf',
+            'yar',
+            'zookeeper',
+        ]
+    ];
+
+    public static function getCoreStubsDirectories(): array
+    {
+        $coreStubs = [];
```
MaXal: This line can be removed and replaced by `$coreStubs = self::$StubDirectoryMap['CORE'];`

tests/Parsers/StubParser.php
```diff
@@ -61,8 +68,26 @@ public static function processStubs(NodeVisitorAbstract $visitor, callable $file
             $traverser = new NodeTraverser();
             $traverser->addVisitor(new ParentConnector());
             $traverser->addVisitor($nameResolver);
-            $traverser->addVisitor($visitor);
+            $coreStubDirectories = PhpCoreStubsProvider::getCoreStubsDirectories();
```
MaXal: Let's move this line out of `foreach`

tests/Parsers/StubParser.php
```diff
@@ -61,8 +68,26 @@ public static function processStubs(NodeVisitorAbstract $visitor, callable $file
...
+            if (self::stubBelongsToCore($file, $coreStubDirectories)){
+                $coreStubVisitor = new CoreStubASTVisitor(self::$stubs);
```
MaXal: Let's create `CoreStubASTVisitor` in `getPhpStormStubs` and pass it to the function.
This will eliminate the static variable and provide a bit of optimization (instance creation inside foreach).

isfedorov: Yeah, I thought about that, but `processStubs` is also used in `MetaExpectedArgumentsCollector` and I've decided not to change to much but just create visitor locally.

PR #666, opened 2019-10-03
add return type to sodium_base642bin()

johnnie502:
The return type is missing.

### DIFF
--- a/sodium/sodium.php
+++ b/sodium/sodium.php
@@ -1280,7 +1280,7 @@ function sodium_bin2base64(string $bin, int $id): string {}
  * @since 7.2.0
  * @see https://www.php.net/manual/en/function.sodium-base642bin.php
  */
-function sodium_base642bin(string $b64, int $id, string $ignore = ''){}
+function sodium_base642bin(string $b64, int $id, string $ignore = ''): string{}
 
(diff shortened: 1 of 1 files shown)


sodium/sodium.php
```diff
@@ -1280,7 +1280,7 @@ function sodium_bin2base64(string $bin, int $id): string {}
  * @since 7.2.0
  * @see https://www.php.net/manual/en/function.sodium-base642bin.php
  */
-function sodium_base642bin(string $b64, int $id, string $ignore = ''){}
```
andrey-sokolov: https://github.com/jedisct1/libsodium-doc/blob/3b6d45a7d0ede7698c85d84bb336a289c309acf6/helpers/README.md looks like it can return int as well, please, use phpdoc to provide these types

isfedorov: Actually based on https://github.com/jedisct1/libsodium/blob/927dfe8e2eaa86160d3ba12a7e3258fbc322909c/src/libsodium/sodium/codecs.c#L266 seems that this method can return only integer but not string

andrey-sokolov:
@johnnie502 Hello, would you, please, provide us any update?

PR #693, opened 2019-10-19
Turkish Comment

selocan1563:


isfedorov:
@selocan1563 please, note that you've broken the tests since Dockerfile is not valid after adding comments to it. Also please note that we use only English comments/code/docs/etc in this repository as far as it is international.

PR #696, opened 2019-10-28
Updated PHPUnit to version 8

grogy:
Updated library for unit testing - locally it works great, I mean that CI say the same :)


grogy:
Error in CI server solve my previous PR number #695

PR #697, opened 2019-10-28
Update phpredis to 5.0.1-beta6

ukko:
https://github.com/ukko/phpredis-phpdoc/releases/tag/5.0.1-beta6

### DIFF
--- a/redis/Redis.php
+++ b/redis/Redis.php
@@ -1508,7 +1508,7 @@ class Redis
      * @param string $key
      *
      * @return int the cardinality of the set identified by key, 0 if the set doesn't exist.
-     * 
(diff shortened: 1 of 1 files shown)


redis/Redis.php
```diff
@@ -1508,7 +1508,7 @@ public function sContains($key, $value)
      * @param string $key
      *
      * @return int the cardinality of the set identified by key, 0 if the set doesn't exist.
-     *
+     *
```
andrey-sokolov: Would you please, get rid of this spaces at the end of empty string?

ukko: ok, fixed

PR #700, opened 2019-10-30
round() never returns false

puggan:
round() never returns false,
not in documentation https://www.php.net/manual/en/function.round.php ,
and not in tests: https://3v4l.org/cAF8J

### DIFF
--- a/standard/standard_3.php
+++ b/standard/standard_3.php
@@ -126,7 +126,7 @@ function floor ($value) {}
  * PHP_ROUND_HALF_EVEN, or
  * PHP_ROUND_HALF_ODD.
  * </p>
- * @return float|false The rounded value
+ * @return float The rounded value
  * @since 4.0
  * @since 5.0
  */


puggan:
Or it dose when it gets an empty array, its not documented in https://www.php.net/round

but as shown by f2912734b064c200f0a52b98e9427a4499429362
and the test https://3v4l.org/vSSu4
it dose return false

puggan:
According to https://bugs.php.net/bug.php?id=74422 the false respone on bad input is due to how 
internal function works.

"If the parameters given to a function are not what it expects, such as passing an array where a string is expected, the return value of the function is undefined. In this case it will likely return NULL but this is just a convention, and cannot be relied upon."
https://www.php.net/manual/en/functions.internal.php

So should the stubs:
1. match the manual, and show the possible values when given the right types for parameters.
2. mark function as return mixed, if the function can take the wrong type of parameters.
3. mark the function with the results that can be return by testing the function

Me myself prefer nr 1, but can't force everyone to agrre on that.

orklah:
I prefer the third for these reasons:
- The manual is not really up to date on egde cases. I've had cases where static tools would advice me to remove a check in my code that was required to keep the program running. The returned type was not documented.
- "wrong type for parameter" is somewhat subjective. Depending on the internal of one function, it may not react the same for empty strings or null for example. Some encoding functions will return weird values if the string containing the encoding name doesn't match a supported encoding (the type of the parameter is right, but the value is wrong)
- A lot of those functions are covered by polyfills in libraries. The polyfills are made to match those weird behaviour of internal functions and IDE's resolver will take those types in consideration and it may actually confuse people when the types in the stubs doesn't match
- There are cases when you can't control input types in a function. A lot of those cases may be caused by bad practices (magic, too permissive APIs..) but they happen. My opinion is that stubs should not dictate that the developper has to filter input types versus them checking the return type instead.

andrey-sokolov:
In general, we are trying to inform about all possible return types in stubs so for now I believe it's better to have an info about false in PHPDoc

PR #702, opened 2019-11-06
Stubs for PECL XXTEA

sirian:
https://pecl.php.net/package/xxtea


isfedorov:
Hi @sirian 
Thank you for your contribution but it's required to do one more step for successful PR. Please note that tests have failed and the error message says "The commited stub map is not up to date. Please regenerate it using ./generate-stub-map". Please execute `php generate-stub-map` and commit the resulting `PhpStormStubsMap.php`

PR #709, opened 2019-11-30
Algo param is a string for php 7.4

fezfez:


standard/password.php
```diff
@@ -162,7 +162,7 @@ function password_get_info ($hash) {}
  * Creates a password hash.
  * @link https://secure.php.net/manual/en/function.password-hash.php
  * @param string $password The user's password.
- * @param int $algo A <a href="https://secure.php.net/manual/en/password.constants.php" class="link">password algorithm constant</a>  denoting the algorithm to use when hashing the password.
+ * @param string $algo A <a href="https://secure.php.net/manual/en/password.constants.php" class="link">password algorithm constant</a>  denoting the algorithm to use when hashing the password.
```
MaXal: Please use `int|string` to keep the function compatible with the previous versions of PHP

fezfez: In 7.4 its only string, the next version will be release for multiple php version ?

MaXal: Our stubs are unified, they support all versions of PHP starting from 5.3 so they should be compatible. That is why we don't remove old methods (just mark them as @Deprecated) and keep a couple of signatures when necessary. In this case, simple `int|string` will do the trick.

fezfez: @MaXal : done :)

PR #710, opened 2019-12-09
Dynamic str_replace return type

b1rdex:
Str_replace returns array if the subject is an array and returns string otherwise. So my intention is to set return type based on the third argument type. Does this work?

### DIFF
--- a/meta/.phpstorm.meta.php
+++ b/meta/.phpstorm.meta.php
@@ -81,6 +81,7 @@ namespace PHPSTORM_META {
   override(\array_values(0), type(0));
   override(\array_combine(0), type(1));
 
+  override(\str_replace(0), elementType(2));
 
     function expectedArguments($functionReference, $argumentIndex, $values) {
         return "expectedArguments " . $functionReference . "at " . $argumentIndex . ": " . $values;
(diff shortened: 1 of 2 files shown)


KapitanOczywisty:
Shouldn't be `type` instead of `elementType`?

meta/.phpstorm.meta.php
```diff
@@ -81,6 +81,7 @@ function elementType($argNum) {
   override(\array_values(0), type(0));
   override(\array_combine(0), type(1));
 
+  override(\str_replace(0), elementType(2));
```
b1rdex:
```suggestion
  override(\str_replace(0), type(2));
```

b1rdex:
> Shouldn't be `type` instead of `elementType`?

Yep, I think you're right. Thanks, fixed.

orklah:
Cool! Maybe we could add preg_replace too?

b1rdex:
> Cool! Maybe we could add preg_replace too?

I'm waiting for JetBrains team response, still not sure if it works or not.

isfedorov:
@b1rdex Thank you for your contribution! We will merge your change but unfortunately due to https://youtrack.jetbrains.com/issue/WI-36161 scalar return type still will not be inferred using meta.
Regarding preg_replace_callback and preg_filter please file a separate PR.

PR #712, opened 2019-12-11
DOMDocument copy return type from param

KapitanOczywisty:
Some methods will return same type as is provided in param.

https://www.php.net/manual/en/domnode.appendchild.php

### DIFF
--- a/meta/.phpstorm.meta.php
+++ b/meta/.phpstorm.meta.php
@@ -84,7 +84,6 @@ namespace PHPSTORM_META {
   override(\DOMDocument::importNode(0), type(0));
   override(\DOMNode::appendChild(0), type(0));
   override(\DOMNode::insertBefore(0), type(0));
-  override(\DOMNode::cloneNode(0), type(0));
   override(\DOMNode::removeChild(0), type(0));
   override(\DOMNode::replaceChild(0), type(1));
   override(\DOMElement::insertBefore(0), type(0));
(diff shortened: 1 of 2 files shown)


meta/.phpstorm.meta.php
```diff
@@ -81,6 +81,16 @@ function elementType($argNum) {
   override(\array_values(0), type(0));
   override(\array_combine(0), type(1));
 
+  override(\DOMDocument::importNode(0), type(0));
+  override(\DOMNode::appendChild(0), type(0));
+  override(\DOMNode::insertBefore(0), type(0));
+  override(\DOMNode::cloneNode(0), type(0));
```
wbars: `DOMNode::cloneNode` accepts boolean as first parameter, not `DOMNode`

meta/.phpstorm.meta.php
```diff
@@ -81,6 +81,16 @@ function elementType($argNum) {
...
+  override(\DOMNode::removeChild(0), type(0));
+  override(\DOMNode::replaceChild(0), type(1));
+  override(\DOMElement::insertBefore(0), type(0));
+  override(\DOMElement::removeChild(0), type(0));
```
wbars: Duplicated lines 88-89

wbars: Since these methods are inherited from `DOMNode` there is no need of specifying them

meta/.phpstorm.meta.php
KapitanOczywisty: I don't know how this got here. Fixed

meta/.phpstorm.meta.php
KapitanOczywisty: I need to remove methods from DOMElement then

KapitanOczywisty:
I've cleaned up and made new PR #717

PR #714, opened 2019-12-11
Change entity to real char to avoid breaking type resolution

orklah:
Hi,

The ```&lt;``` entity after return type is breaking the phpstorm resolver:
Capture

Not sure if it's the right way to go as I'm not sure why there are entities and what'd happen if we remove them.

There are other functions where this is the case too (for example, strcoll), If this commit is validated, I'll modify the others

### DIFF
--- a/standard/standard_1.php
+++ b/standard/standard_1.php
@@ -452,8 +452,8 @@ function strpbrk ($haystack, $char_list) {}
  * If case_insensitivity is true, comparison is
  * case insensitive.
  * </p>
- * @return int &lt; 0 if main_str from position
- * offset is less than str, &gt;
+ * @return int < 0 if main_str from position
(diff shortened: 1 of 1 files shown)


b1rdex:
It looks like a bug for me. There is a space between type and description that distinguishes them. But PhpStorm somehow allows spaces in types declaration around `|` and `&`, even though that's not following PhpDoc standards and any popular PhpDoc parser would parse that differently.

isfedorov:
@orklah Thank you very much for pointing this out! This is definitely the bug and I've filled it in YT. https://youtrack.jetbrains.com/issue/WI-50310, please feel free to vote. So I'm closing this PR as it's not related actually with stubs but rather is a bug in phpdoc processing.

PR #725, opened 2019-12-28
Annotate PSR-14 Event Dispatcher

TimothyBJacobs:
The `EventDispatcherInterface` returns the same object as it was passed. https://github.com/php-fig/event-dispatcher/blob/master/src/EventDispatcherInterface.php#L18

This makes it possible to get type completion in scenarios like this.

```php
$event = $eventDispatcher->dispatch(new MyEvent());
$event->callSomeMethodOnMyEvent();
```

### DIFF
--- a/meta/.phpstorm.meta.php
+++ b/meta/.phpstorm.meta.php
@@ -87,6 +87,7 @@ namespace PHPSTORM_META {
   override(\DOMNode::insertBefore(0), type(0));
   override(\DOMNode::removeChild(0), type(0));
   override(\DOMNode::replaceChild(0), type(1));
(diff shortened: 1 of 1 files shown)


MaXal:
@TimothyBJacobs That is a great idea but there is the following problem. `EventDispatcherInterface` is not part of stubs rather a library. The addition of all libraries specifics into the stubs will make them unmanageable so I should decline the Pull Request.
But I would like to encourage you to provide it to the library itself.

TimothyBJacobs:
Hi @MaXal I would typically agree, but I think an exception should be considered for the various PSRs. They are about as close to official PHP user land libraries as we can get, and I doubt that the PSR itself would accept having an IDE specific configuration file bundled with its interfaces.

PR #728, opened 2020-01-02
Fix false for return.

dereuromark:
Refs https://youtrack.jetbrains.com/issue/WI-50517

I don't like this `<p>` stuff in the docblock lines - maybe the can be removed in follow up PRs.
But since they close with `</p>` already, they should also correctly open (so I added those, too).

### DIFF
--- a/Core/Core.php
+++ b/Core/Core.php
@@ -329,7 +329,7 @@ function defined ($name) {}
  * @param object $object [optional] <p>
  * The tested object. This parameter may be omitted when inside a class.
  * </p>
- * @return string the name of the class of which <i>object</i> is an
+ * @return string|false <p>The name of the class of which <i>object</i> is an
  * instance. Returns false if <i>object</i> is not an
  * object.
  * </p>
@@ -344,7 +344,7 @@ function get_class ($object = null) {}
 /**
  * the "Late Static Binding" class name
  * @link https://php.net/manual/en/function.get-called-class.php
- * @return string the class name. Returns false if called from outside a class.
+ * @return string|false The class name. Returns false if called from outside a class.
  * @since 5.3
  */
 function get_called_class () {}
@@ -352,10 +352,10 @@ function get_called_class () {}
 /**
  * Retrieves the parent class name for object or class
  * @link https://php.net/manual/en/function.get-parent-class.php
- * @param mixed $object [optional] <p>
+ * @param object|string|null $object [optional] <p>
  * The tested object or class name
  * </p>
- * @return string the name of the parent class of the class of which
+ * @return string|false <p>The name of the parent class of the class of which
  * <i>object</i> is an instance or the name.
  * </p>
  * <p>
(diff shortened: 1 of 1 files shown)


dereuromark:
Is there any feedback on this? OK to be merged?

Core/Core.php
```diff
@@ -371,7 +371,7 @@ function get_parent_class ($object = null) {}
 /**
  * Checks if the class method exists
  * @link https://php.net/manual/en/function.method-exists.php
- * @param mixed $object <p>
+ * @param object|string $object <p>
```
pestretsov: Looks like the `mixed` type is actually the correct one – [implementation just returns false](https://github.com/php/php-src/blob/master/Zend/zend_builtin_functions.c#L1124) for any other types. Although it doesn't make much sense to pass anything other than `object|string`, it would be more correct to follow the specification here

Core/Core.php
```diff
@@ -344,18 +344,18 @@ function get_class ($object = null) {}
...
+ * @return string|false The class name. Returns false if called from outside a class.
  * @since 5.3
  */
 function get_called_class () {}
 
 /**
  * Retrieves the parent class name for object or class
  * @link https://php.net/manual/en/function.get-parent-class.php
- * @param mixed $object [optional] <p>
+ * @param object|string|null $object [optional] <p>
```
pestretsov: Same here

Core/Core.php
dereuromark: I see. All right, reverting then.

dereuromark:
Done :)

orklah:
> it would be more correct to follow the specification here

Not sure I agree. The goals of stubs are detecting developer's errors and provide documentation in IDE, not being a perfect reflection of the implementation.

If a parameter type can't make sense, it should be flagged, not accepted because the implementation will accept it.

It's a different story for return types. They have to be the exact representation of the implementation because that may happen and we want to be able to handle it without raising an error in IDE.

dereuromark:
I also rather like to document params as to be the "expected" types handled, not all possible input types. This makes tools like PHPStan/Psalm find bugs/issues earlier :) `mixed` as type is the most evil thing you can have always. Kills all static analizer.
But for me right now the priority was the return types. Thus I didnt pursue it further.

pestretsov:
@dereuromark great, thanks!
@orklah I get your idea, but from the language, specification and implementation point of view, passing other types is not an error. This is kinda a philosophical question of finding the right balance, because in IDE it may prevent some users from making a mistake while at the same time distract other users with false-negative highlighting. Let it be like that for now, but we can discuss it later in separate PR or YouTrack ticket.

dereuromark:
@pestretsov Thanks for the merge.
How do we get this on our local machine? Does this ship itself with an update as php.jar? Or can we also rebuild ourselves?

pestretsov:
@dereuromark you can always change stubs locally as you like and [set the path inside IDE](https://github.com/JetBrains/phpstorm-stubs#updating-the-ide) ([more detailed info](https://blog.jetbrains.com/phpstorm/2018/03/how-to-provide-stubs-for-phpstorm/)). Otherwise, it will be bundled in one of the upcoming PhpStorm updates.

dereuromark:
I created a small follow up: https://github.com/JetBrains/phpstorm-stubs/pull/733

PR #733, opened 2020-01-16
Fix invalid defaulting.

dereuromark:
Follow https://github.com/JetBrains/phpstorm-stubs/pull/728

Using no argument is actually not valid

> Warning</b>:  implode() expects at least 1 parameter, 0 given

It is also already warning in yellow in PHPStorm IDE as code smell:
Tooltip_001

### DIFF
--- a/standard/standard_1.php
+++ b/standard/standard_1.php
@@ -260,7 +260,7 @@ function dirname ($path, $levels = 1) {}
  * extension (if any), and filename.
  * </p>
  * <p>
- * If options is used, this function will return a 
+ * If options is used, this function will return a
  * string if not all elements are requested.
  * @since 4.0.3
  * @since 5.0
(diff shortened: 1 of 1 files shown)


dereuromark:
They are currently noise and bad practice to keep (see other lines), but yeah, either way works.

KapitanOczywisty:
https://github.com/JetBrains/phpstorm-stubs/blob/1508824b38f726e4ca11b2985a5d22989936d770/CONTRIBUTING.md#L9

It *is* bad practice, but removing them messes with blame view for example.

dereuromark:
In this case it is probably acceptable. It is very minor cleanup.

bmewburn:
I think it would be incorrect to change it to `string[] $pieces`.

```
$ php -r "echo implode(',', [1,2,3]), PHP_EOL;"
1,2,3
```

pestretsov:
@dereuromark hi! As @bmewburn has mentioned, the param type change is not correct and the default value for `glue` was added intentionally in https://youtrack.jetbrains.com/issue/WI-30027

dereuromark:
This Ticket you linked was about the 2nd not the first argument. The first must not default. Check sandbox.

PR #741, opened 2020-01-28
Add stubs for ext-ds

enumag:
https://youtrack.jetbrains.com/issue/WI-38924


pestretsov:
@enumag hi! Thanks for the contribution. Please, fix those two points:
1. It seems like the min supported language version is PHP 7.0, but in those stubs there are language features that are not supported in PHP 7.0 like nullable types, return void types and others. This means, that users will get red code in IDE if they try to use it
2. Some data structure interfaces and classes are missing: Hashable, Stack, Queue, PriorityQueue

enumag:
> language features that are not supported in PHP 7.0 like nullable types, return void types and others

@pestretsov What "others" do you mean? I'm not using the iterable type (only in annotations) so I think this should be all, no? Or do you want me to change the `@param iterable` to `@param array|Traversable`?

pestretsov:
@enumag Yeah, seems you are right, let's move those to doc tags just to remove excessive red code)

MaXal:
@enumag Please fix the red code otherwise we can't merge your pull-request.

enumag:
Unfortunately I'm too busy to finish this at this time. I'll reopen this when I'm able to work on it and finish it.

imme-emosol:
For reference purposes, repository contains [/ds/ds.php](https://github.com/JetBrains/phpstorm-stubs/blob/master/ds/ds.php).

PR #743, opened 2020-01-29
Add stub for uuid PECL extension

adamkoehn:
This is the stub for the uuid PECL extension at version 1.1.0. I did not write the extension, I'm just using it. The documentation is a combination of comments from the extension source, and documentation from uuid.h. The PECL extension can be found here: https://pecl.php.net/package/uuid/1.1.0


pestretsov:
Thank you!

PR #746, opened 2020-02-12
Stubs for ext-decimal 1.3.0

aleksbrgt:
Stubs for the decimal  extension (https://php-decimal.io).

I started using this extension but there was no stub for it.

The PR is about version 1.3.0 of the extension.


aleksbrgt:
Just noticed it should resolve https://youtrack.jetbrains.com/issue/WI-50958

PR #754, opened 2020-03-10
fix: IntlChar::chr and IntlChar::ord returns null

sirian:
```php
null === IntlChar::chr(null)
null === IntlChar::chr(1.1)
null === IntlChar::ord(null)
null === IntlChar::ord('abc')
```


MaXal:
Please use `string|null` instead of `?string` in PHPDoc since your version is not well-supported.

AlexMovsesov:
Hi, thanks for your contribution! I noticed that many of the IntlChar methods have the same return type as chr or ord, so I extended this review with fixes for these methods. Also I added metadata for the most of the IntlChar methods.

PR #757, opened 2020-03-17
Moved getallheaders() and fastcgi_finish_request()

moliata:
Hello,

this pull request moves `getallheaders()` from ext-apache to PHP core as `getallheaders()` is no longer Apache-specific and is available in both PHP FastCGI Process Manager and CLI Development Server as well (for more information, please take a look [here](https://www.php.net/manual/en/function.getallheaders.php)).

`fastcgi_finish_request()` is also moved from PHP core to PHP-FPM as it's only available with the PHP FastCGI Process Manager SAPI (for more information, please take a look [here](https://www.php.net/manual/en/function.fastcgi-finish-request.php) and [here](https://www.php.net/manual/en/book.fpm.php)).

Best regards,
Ben.

### DIFF
--- a/apache/apache.php
+++ b/apache/apache.php
@@ -105,14 +105,6 @@ function apache_response_headers () {}
 function apache_setenv ( $variable, $value, $walk_to_top = false ) {}
 
 /**
- * Fetches all HTTP headers from the current request.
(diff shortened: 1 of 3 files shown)


moliata:
To be fixed.

PR #758, opened 2020-03-18
Fix getopt() return type

b1rdex:
I doubt if it can be specified as `string[]|false[]|false`? I'm not sure.


MaXal:
It can be added as `string[]|false[]|false` why not:).

b1rdex:
@MaXal I've updated the PR.

PR #760, opened 2020-03-18
Add doc comments to curl constants

sjlehn:
Added doc comments to a handful of constants based on PHP documentation


curl/curl_d.php
```diff
@@ -48,189 +60,820 @@
...
+ * @link https://www.php.net/manual/en/function.curl-setopt.php
+ * @since 5.2.3
+ */
 define ('CURLOPT_CONNECTTIMEOUT_MS', 156);
+/**
+ * <b>FALSE</b> to stop cURL from verifying the peer's certificate.
+ * Alternate certificates to verify against can be specified with the <b>CURLOPT_CAINFO</b> option or
+ * a certificate directory can be specified with the <b>CURLOPT_CAPATH</b> option.
+ * @link https://www.php.net/manual/en/function.curl-setopt.php
+ * @since 7.1
```
samuelfa, line 661: Hi!

This is incorrect. The option of "CURLOPT_SSL_VERIFYPEER" is available since at least PHP 5.
The documentation says:

https://www.php.net/manual/en/function.curl-setopt.php
"TRUE by default as of **cURL 7.10**. Default bundle installed as of **cURL 7.10**."

It's referring to "cURL" package, not PHP.
This error generate a fake error in PHPStorm.

image

FeBe95: Fixed in #826.

PR #761, opened 2020-03-18
Add stubs for ext-expect

sjlehn:
Added stubs for the expect extension (https://www.php.net/manual/en/intro.expect.php)


MaXal:
Please execute `php generate-stub-map` and add the resulting `PhpStormStubsMap.php` to the pull request.

PR #763, opened 2020-03-21
Fix event optional arguments

kelunik:
Used NAN to keep the type non-nullable.


AlexMovsesov:
Hi, thx for your contribution! As I can see from the source code of the https://pecl.php.net/event float: -1 is more suitable as default values for `$timeout` param for these methods. I fixed these values.

PR #764, opened 2020-03-25
Add stubs for BlackfireProbe

lolautruche:
This PR adds stubs for the `\BlackfireProbe` class, provided by the Blackfire extension.

### DIFF
--- a/blackfire/blackfire.php
+++ b/blackfire/blackfire.php
@@ -1,0 +1,107 @@
+<?php
+
+/**
+ * Blackfire extension stubs.
+ * @link https://blackfire.io
+ */
+final class BlackfireProbe
+{
+    /**
+     * Returns a global singleton and enables it by default.
+     *
+     * @return \BlackfireProbe
+     */
+    public static function getMainInstance() {}
+
+    /**
+     * Tells whether the probe is currently profiling or not.
+     *
+     * @return bool
+     */
+    public static function isEnabled() {}
+
+    /**
+     * Adds a marker for the Timeline View.
+     *
+     * @param string $markerName
+     */
+    public static function addMarker($markerName = '') {}
+
+    /**
+     * BlackfireProbe constructor. SHOULD NOT BE USED DIRECTLY!
+     * USE getMainInstance() INSTEAD!
+     *
+     * @param string $query An URL-encoded string that configures the probe. Part of the string is signed.
+     *                      Typically comes from X-Blackfire-Query HTTP header or BLACKRFIRE_QUERY environment variable.
+     * @param string|null $envId An id that is given to the agent for signature impersonation.
+     * @param string|null $envToken The token associated to $envId.
+     * @param string|null $agentSocket The URL where profiles will be written (directory, socket or TCP destination).
+     */
+    public function __construct($query, $envId = null, $envToken = null, $agentSocket = null) {}
+
+    /**
+     * Tells if the probe is cryptographically verified,
+     * i.e. if the signature in X-Blackfire-Query HTTP header or
+     * BLACKFIRE_QUERY environment variable is valid.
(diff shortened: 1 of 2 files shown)


AlexMovsesov:
Hi, thx for your contribution!
Why there is no constructor stub for this one:
https://github.com/blackfireio/php-sdk/blob/4188f61bad20b45b19b61c3b6225f9aec2c8c8e4/src/BlackfireProbe.php#L191 ?
It marked as API, is it internal API?

blackfire/blackfire.php
```diff
@@ -0,0 +1,95 @@
...
+     * BLACKFIRE_QUERY environment variable is valid.
+     *
+     * @return bool
+     */
+    public function isVerified() {}
+
+    /**
+     * @param string $configuration
+     */
+    public function setConfiguration($configuration) {}
```
AlexMovsesov: This param is nullable if this is the source: https://github.com/blackfireio/php-sdk/blob/4188f61bad20b45b19b61c3b6225f9aec2c8c8e4/src/BlackfireProbe.php#L291

lolautruche:
> Hi, thx for your contribution!
> Why there is no constructor stub for this one:
> https://github.com/blackfireio/php-sdk/blob/4188f61bad20b45b19b61c3b6225f9aec2c8c8e4/src/BlackfireProbe.php#L191 ?
> It marked as API, is it internal API?

Hey @AlexMovsesov !
I didn't document the constructor in order to avoid confusion with the singleton. The proper way for getting the probe object is by calling `\BlackfireProbe::getMainInstance()`.
Also, please note that the source you're pointing out is only a polyfill for PHP 5.2.
I could document the constructor however, but specifying that one should not use it...

blackfire/blackfire.php
lolautruche: It is not nullable in the probe itself:

```
Extension [ <persistent> extension #68 blackfire version 1.31.0~mac-x64-non_zts73 ] {
  - Classes [2] {
    Class [ <internal:blackfire> final class BlackfireProbe ] {

      - Methods [10] {
        Method [ <internal:blackfire, ctor> public method __construct ] {

          - Parameters [4] {
            Parameter #0 [ <required> $query ]
            Parameter #1 [ <optional> $probeId ]
            Parameter #2 [ <optional> $probeToken ]
            Parameter #3 [ <optional> $outputUrl ]
          }
        }

        Method [ <internal:blackfire> public method isVerified ] {
        }

        Method [ <internal:blackfire> public method getResponseLine ] {
        }

        Method [ <internal:blackfire> public method setConfiguration ] {

          - Parameters [1] {
            Parameter #0 [ <required> $str ]
          }
        }

        Method [ <internal:blackfire> public method enable ] {
        }

        Method [ <internal:blackfire> public method disable ] {
        }

        Method [ <internal:blackfire> public method close ] {
        }

        Method [ <internal:blackfire> public method discard ] {
        }

        Method [ <internal:blackfire> public method getHeapUsage ] {
        }

        Method [ <internal:blackfire> public method createSubProfileQuery ] {
        }
      }
    }
  }
}
```

lolautruche:
I just updated the PR :-)

blackfire/blackfire.php
AlexMovsesov: I mean that by that polyfill' source code - 
method setConfiguration accepts null argument: `$probe->setConfiguration(null);`,
it can be indicated in the doc: `@param string|null`
But I don't see the real source code, so it was an assumption.

lolautruche: Sure, no problem 😉 ! It is true that the source code is not public, that's why I pasted a snippet of `php --re blackfire` result.

lolautruche:
Thank you!

PR #767, opened 2020-04-02
WI-52079: Update proc_open function parameter

uzulla:
- `proc_open` function able be $cmd(first arg) passed as array in PHP 7.4
- update `$other_options` paragraph

### ref
https://www.php.net/manual/en/function.proc-open.php

### original issue
https://youtrack.jetbrains.com/issue/WI-52079


bzixilu:
Thank you for contribution!!!

PR #768, opened 2020-04-02
 Default value seems to be wrong

tuppabox:
The function also returns `null` if you pass `null` for `$options` instead of `[]`


OhNiice:
https://github.com/php/php-src/blob/master/ext/standard/password.c#L653
src seems to confirm :)

bzixilu:

PR #770, opened 2020-04-08
grpc: Fix ChannelCredentials::createInsecure @return hint

martin-helmich:
`ChannelCredentials::createInsecure` returns a `ChannelCredentials` instance, not `null`.

### DIFF
--- a/grpc/grpc.php
+++ b/grpc/grpc.php
@@ -535,7 +535,7 @@ namespace Grpc
         /**
          * Create insecure channel credentials
          *
-         * @return null
+         * @return ChannelCredentials The new composite credentials object
          */
         public static function createInsecure() {}
     }


bzixilu:
@martin-helmich I've run the following code with grpc extension installed:
`<? php
$ insecure = \ Grpc \ ChannelCredentials :: createInsecure ();
`
and got `null` as a result.

Also, I checked the implementation of ChannelCredentials::createInsecure here: https://github.com/grpc/grpc/blob/29bcc3f759dd06c18ac6fdc755f6436df21d56d2/src/php/ext/grpc/channel_credentials.c#L230 and found only null as a return value.

Could you please help me with a code sample, when the function doesn't not null?

martin-helmich:
Oh, you're absolutely right. 🤦 TBH, the usage of `createInsecure` in [the tutorial](https://grpc.io/docs/tutorials/basic/php/#constructing-a-client-object) -- and that it seemed only natural that a `SomeClass::create*` function should return an instance of said class -- led me to the assumption that `createInsecure` should return a `ChannelCredentials` instance.

But, the GRPC module's source code speaks for itself. Sorry for the fake news.

PR #771, opened 2020-04-09
WI-52193 Missing 'gc_status' return type

samuelnogueira:
Added missing documeation for gc_status core function
https://youtrack.jetbrains.com/issue/WI-52193


bzixilu:

PR #772, opened 2020-04-12
The value of mysqli::$sqlstate can be false

morozov:
Despite the [documentation](https://www.php.net/manual/en/class.mysqli.php), `mysqli::$sqlstate` can be actually false:
```php
$conn = mysqli_init();

var_dump(get_class($conn));
var_dump($conn->sqlstate);

// string(6) "mysqli"
// bool(false)
```

### DIFF
--- a/mysqli/mysqli.php
+++ b/mysqli/mysqli.php
@@ -103,7 +103,7 @@ class mysqli  {
 	 */
 	public $server_version;
 	/**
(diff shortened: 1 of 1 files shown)


morozov:
Actually, according to [bug #79469](https://bugs.php.net/bug.php?id=79469), this code is invalid.

PR #773, opened 2020-04-13
Fix default value

BenMorel:
According to [the doc](https://www.php.net/manual/en/function.fopen.php), `$use_include_path` is `false` by default.


bzixilu:

PR #774, opened 2020-04-15
Memcached: According to memcached 3.1.5 api, update some constant and function.

stevennight:
https://github.com/php-memcached-dev/php-memcached/blob/v3.1.5/memcached-api.php
https://github.com/php-memcached-dev/php-memcached/blob/v3.1.5/php_memcached.c
https://www.php.net/manual/en/book.memcached.php
http://docs.libmemcached.org/index.html
https://alo-framework.readthedocs.io/en/stable/class-Memcached.html

### DIFF
--- a/memcached/memcached.php
+++ b/memcached/memcached.php
@@ -1,6 +1,6 @@
 <?php
 
-// Start of memcached v.3.0.4
+// Start of memcached v.3.1.5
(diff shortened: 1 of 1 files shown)


stevennight:
the type double is wrong, close now.

PR #776, opened 2020-04-15
Implement Traversable

greg0ire:
Classes cannot implement Traversable directly, they have to implement
either Iterator or IteratorAggregate. This fixes a crash when using this
package with PHPStan:

> PHP Fatal error:  Class mysqli_result must implement interface
> Traversable as part of either Iterator or IteratorAggregate in Unknown
> on line 0

Source to back up my claim:

>   This is an internal engine interface which cannot be implemented in PHP scripts. Either IteratorAggregate or Iterator must be used instead. When implementing an interface which extends Traversable, make sure to list IteratorAggregate or Iterator before its name in the implements clause. 

From https://www.php.net/manual/en/class.traversable.php

### DIFF
--- a/mysqli/mysqli.php
+++ b/mysqli/mysqli.php
@@ -786,7 +786,7 @@ final class mysqli_warning  {
  * Implements Traversable since 5.4
  * @link https://php.net/manual/en/class.mysqli-result.php
  */
-class mysqli_result implements Traversable  {
+class mysqli_result implements IteratorAggregate  {
 	/**
 	 * @var int
 	 */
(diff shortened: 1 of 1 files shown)


bzixilu:
Unfortunately, I couldn't accept this pull-request since it added irrelevant information to the stubs.
You added mandatory for `IteratorAggregate` `getIterator` method, but based on: 
```
Reflection::export(new ReflectionClass('mysqli_result'));
```
`mysqli_result` doesn't contain this method. So if it appears in stubs, PhpStorm starts wrongly show this method in completion and other places.

greg0ire:
You're right I went for the lazy implementation, but the docs talks about `Iterator`, not `IteratorAggregate`:

https://www.php.net/manual/en/class.mysqli-result.php

Please reopen this, and I will push an implementation of `Iterator` instead.

bzixilu:
```
Reflection::export(new ReflectionClass('mysqli_result'));
```
also says that it implements `Traversable` but not `Iterator`. So I would avoid such changes in stubs as well. 
I would recommend you to check PhpStan behavior with the authors, probably it could be fixed/filtered on their side.

PR #777, opened 2020-04-15
FFI Improvements

SerafimArts:
Fixes/Improvements:

- `FFI::load` allows only string arguments
- `FFI::cdef` returns an instance of FFI in cases where declarations are more than one.
- Although it is noted in the documentation that some FFI functions/mettods passes objects by reference (that is, they can change them), but in reality this is not so. Fix it.

### DIFF
--- a/FFI/FFI.php
+++ b/FFI/FFI.php
@@ -58,7 +58,7 @@ namespace {
          * @param string $filename
          * @return FFI
          */
-        public static function load($filename): FFI {}
+        public static function load(string $filename): FFI {}
 
         /**
          * FFI definition parsing and shared library loading may take
--- a/FFI/FFI.php
+++ b/FFI/FFI.php
@@ -33,7 +33,7 @@ namespace {
          * @return CData|FFI|mixed
          * @throws ParserException
          */
-        public static function cdef(string $code, string $library = null) {}
+        public static function cdef(string $code, string $library = null): FFI {}
 
         /**
          * <p>Instead of embedding of a long C definition into PHP string,
@@ -276,4 +276,3 @@ namespace FFI {
     {
     }
 }
-
--- a/FFI/FFI.php
+++ b/FFI/FFI.php
@@ -114,7 +114,7 @@ namespace {
          * @param CData $pointer
          * @return void
          */
-        public static function free(CData &$pointer): void {}
+        public static function free(CData $pointer): void {}
 
         /**
          * Casts given $pointer to another C type, specified by C declaration
@@ -129,7 +129,7 @@ namespace {
          * @param CData $pointer
          * @return CData
          */
-        public static function cast($type, CData &$pointer): CData {}
+        public static function cast($type, CData $pointer): CData {}
 
         /**
          * This function creates and returns a FFI\CType object, representng
@@ -152,7 +152,7 @@ namespace {
          * @param CData $pointer
          * @return CType
          */
-        public static function typeof(CData &$pointer): CType {}
+        public static function typeof(CData $pointer): CType {}
 
         /**
          * Constructs a new C array type with elements of $type and
@@ -174,7 +174,7 @@ namespace {
          * @param CData $pointer
          * @return CData
          */
-        public static function addr(CData &$pointer): CData {}
+        public static function addr(CData $pointer): CData {}
 
         /**
          * Returns size of C data type of the given FFI\CData or FFI\CType.
@@ -200,7 +200,7 @@ namespace {
          * @param mixed $source
          * @param int $size
          */
-        public static function memcpy(CData &$target, &$source, int $size): void {}
+        public static function memcpy(CData $target, &$source, int $size): void {}
 
         /**
          * Compares $size bytes from memory area $a and $b.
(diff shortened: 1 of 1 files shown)


SerafimArts:
P.S. In addition, there are still problems:
1) `FFI::new(...)` can be called both statically and using the current FFI scope: `FFI::load('...')->new(...)` however, stubs do not allow this
2) Same with `FFI::cast`, `FFI::type` and `FFI::arrayType`

FFI/FFI.php
```diff
@@ -30,7 +30,7 @@ class FFI
          *
          * @param string $code The collection of C declarations.
          * @param string|null $library DSO library.
-         * @return CData|mixed
+         * @return CData|FFI|mixed
```
sj-i, line 33:
```suggestion
         * @return FFI
```

https://github.com/JetBrains/phpstorm-stubs/pull/777#issue-403991092
> FFI::cdef returns an instance of FFI in cases where declarations are more than one.

Actually it always returns \FFI even with empty $code.
https://github.com/php/php-src/blob/fee614f66bfb72384c7c6f316cad0c5c68fa9c0c/ext/ffi/ffi.c#L2932-L2940
https://github.com/php/php-src/blob/fee614f66bfb72384c7c6f316cad0c5c68fa9c0c/ext/ffi/ffi.c#L5008-L5009

```
 % php -r "var_dump(\FFI::cdef(''));"
object(FFI)#2 (0) {
}
```

FFI/FFI.php
```diff
@@ -30,7 +30,7 @@ class FFI
...
          * @throws ParserException
          */
         public static function cdef(string $code, string $library = null) {}
```
sj-i, line 36:
```suggestion
        public static function cdef(string $code, string $library = null): FFI {}
```

FFI/FFI.php
SerafimArts: Yep, its true.

bzixilu:
```
$function = new ReflectionMethod('FFI', 'free');
foreach ($function->getParameters() as $param) {
    if ($param->isPassedByReference()) {
        echo $param->getName() . ' is passed by reference';
    }
    echo PHP_EOL;
}
```
says that it's still by reference. 
@SerafimArts Could you please elaborate on how you checked that objects are not passed by reference.

sj-i:
> says that it's still by reference.

Aha, interesting.
Maybe `CData &` should be replaced with `CData|CData&` ...? But I'm not sure about using `&` in `@param` is a valid type notation or not on tools.
In case of any good solution couldn't be found, I suggest to revert changes only about pass by reference.

```
<?php // test1.php
$function = new ReflectionMethod('FFI', 'free');
foreach ($function->getParameters() as $param) {
    if ($param->isPassedByReference()) {
        echo $param->getName() . ' is passed by reference';
        if ($param->canBePassedByValue()) {
            echo ' and also can be passed by value';
        }
    }
    echo PHP_EOL;
}
```
```
% php test1.php 
ptr is passed by reference and also can be passed by value
```

```
<?php // test2.php
\FFI::cast('void *', \FFI::new('char *'));
echo 'accepts passed by value', PHP_EOL;
(function (\FFI\CData &$param) {
    echo 'outputs notice at passed by value', PHP_EOL;
})(\FFI::new('char *'));
```
```
 % php test2.php
accepts passed by value
PHP Notice:  Only variables should be passed by reference in /home/sji/work/test2.php on line 6
outputs notice at passed by value
```
2nd parameter of \FFI:cast() can be passed by value without notice.
Actually all parameters in methods of this class accept passing parameter by value, including \FFI::memcmp().

FYI:
They are defined as ZEND_SEND_PREFER_REF in its arginfo.
https://github.com/php/php-src/blame/fee614f66bfb72384c7c6f316cad0c5c68fa9c0c/ext/ffi/ffi.c#L4460-L4538

They accept both value and reference.
It seems that extensions can define such functions unlike userland functions.
array_multisort() is another example of such functions, though the manual says the first parameter is simply a reference here too. 
https://www.php.net/manual/ja/function.array-multisort.php

```
% php -r "var_dump(array_multisort([]));"
bool(true)
```

SerafimArts:
@bzixilu 
```
FFI::free(FFI::new('int8_t', false));
```

If the argument required passing by reference, this code would cause an error like this:
```
function my_free(\FFI\CData &$ref)
{
    FFI::free($ref);
}

my_free(FFI::new('uint8_t', false));
// Notice: Only variables should be passed by reference in xxxx
```

nikic:
Unfortunately the FFI extension contains quite a bit of "magic". A number of parameters are declared using the PREFER_REF send mode, which will try to pass by reference is possible and by value otherwise. In ReflectionParameter, such a case should be visible by both isPassedByReference() and canBePassedByValue() returning true (usually, only one of them can be true).

The reason for this behavior is this (example taken from 040.phpt):

```
$y = FFI::new("int[2]");
$y[0] = 6;
var_dump(FFI::cast("int8_t[4]", $y[0]));
```

Normally, writing `$y[0]` will return just `int(6)`. However, FFI::cast() wants to receive FFI\CData instead. By accepting the parameter by reference, it will be fetched "for writing", in which case `$y[0]` will return the FFI\CData value, not the scalar it contains. Other FFI parameters work similarly.

This is bad design in the FFI extension (it should have always provided FFI\CData, and an extra call or cast to get the scalar value), but well, this is the situation we have to deal with now...

For the purposes of this PR, I agree that it's best to remove the `&` in stubs. From the end-user perspective the fact that by-reference passing is used is not important (and it's really just an internal hack anyway). So this looks good to me (modulo the suggested change by @sj-i, to keep phpdoc in sync).

bzixilu:
@SerafimArts merged, thank you for your contribution!

PR #781, opened 2020-04-18
Added SoapHeader::__construct()

ondrejmirtes:
Hi, the stubs are missing [SoapHeader::__construct](https://www.php.net/manual/en/soapheader.construct.php) so I added it. Thanks.


bzixilu:

PR #782, opened 2020-04-22
Change PHPDoc for unserialize

n30fly:
According official docs first parameter has string type and `options` parameter can't be null.

https://www.php.net/manual/en/function.unserialize.php
https://github.com/php/php-src/blob/beff93f60ddb330cc5b04d716af3ba9cba108c8e/ext/standard/basic_functions.stub.php#L1528


bzixilu:
Great! Thank you for your contribution

PR #783, opened 2020-04-22
Add missed RdKafka\Producer::flush and purge

autowp:
Ref: https://arnaud.le-blanc.net/php-rdkafka-doc/phpdoc/class.rdkafka-producer.html


bzixilu:
Great! Thank you for your contribution

PR #785, opened 2020-04-23
WI-52602 Fix RFC reference for urlencode function.

ouvtk:
References:
https://youtrack.jetbrains.com/issue/WI-52602
https://www.php.net/manual/en/function.urlencode.php


bzixilu:

PR #789, opened 2020-04-28
fix wrong @return at \FFI::cdef()

sj-i:
Fixes the leftovers from https://github.com/JetBrains/phpstorm-stubs/pull/777#discussion_r410971024


bzixilu:
Great! Thank you

PR #793, opened 2020-05-09
Add stubs for sapi/phpdbg functions

sebastianbergmann:
Motivated by [this](https://github.com/sebastianbergmann/phpunit/issues/4211#issuecomment-625861828) issue in PHPUnit.

Based on https://github.com/php/php-src/blob/master/sapi/phpdbg/phpdbg.stub.php


sebastianbergmann:
Thank you for merging this. Can you share when a new release will be tagged that includes this addition?

andrey-sokolov:
@sebastianbergmann we are going to publish it at 2020.1.2 (approximately in a week).

PR #809, opened 2020-05-27
Fixed functions with variadic parameter

kukulich:


standard/standard_2.php
```diff
@@ -424,13 +421,13 @@ function sscanf ($str, $format, &...$_) {}
...
  * sprintf documentation.
  * </p>
- * @param mixed $_ [optional]
+ * @param mixed ...$vars [optional]
  * @return array|int If only two parameters were passed to this function, the values parsed will be
  * returned as an array. Otherwise, if optional parameters are passed, the
  * function will return the number of assigned values. The optional
  * parameters must be passed by reference.
  */
-function fscanf ($handle, $format, &$_ = null) {}
```
andrey-sokolov, line 433: it is passed by reference

kukulich: @andrey-sokolov Fixed.

PR #816, opened 2020-05-31
Added missing classes/functions/constants

kukulich:
Found with `get_defined_functions`, `get_defined_constants`, `get_declared_classes()` etc and generated by Reflection.

Some are even not documented in PHP so I'm not to able to improve the stubs.


gmp/gmp.php
```diff
@@ -677,6 +677,57 @@ function gmp_rootrem (GMP $a, $nth) {}
...
+ * @since 7.3
+ */
+function gmp_binomial($a, $b) {}
+
+/**
+ * Computes the Kronecker symbol
+ *
+ * @link https://www.php.net/manual/en/function.gmp-kronecker.php
+ *
+ * @param GMP|string|float|int $a
```
isfedorov, line 698: Seems `resource` param type is missed here

sockets/sockets.php
```diff
@@ -1332,6 +1332,44 @@ function socket_getopt ($socket, $level, $optname) {}
...
+ */
+function socket_wsaprotocol_info_export($socket, $target_pid) {}
+
+/**
+ * Imports a Socket from another Process
+ *
+ * @link https://www.php.net/manual/en/function.socket-wsaprotocol-info-import.php
+ *
+ * @param string $info_id
+ * @return string|false
```
isfedorov, line 1354: seems return should be `resource|false`

kukulich: Thanks, fixed.

gmp/gmp.php
kukulich: The PHP documentation is a little strange.

The function exists since 7.3 so I think the part "Either a GMP number resource in PHP 5.5 and earlier" should never apply.

isfedorov: Oh, right. I see. But in this case I think that `resource` should be removed from `gmp_binomial` function as it completely vise-versa case. Function is available since 7.3 but resource parameter for php 5.5 and earlier.

kukulich: You're right, fixed.

PR #820, opened 2020-06-08
mysqli correct link & params

stdex:


mysqli/mysqli.php
```diff
@@ -141,7 +142,7 @@ public function __construct (
 		$username = null,
 		$passwd = null,
 		$dbname = null,
-		$port = null,
+		$port = 0,
```
isfedorov, line 145: according to https://github.com/php/php-src/blob/master/ext/mysqli/mysqli.stub.php 
```
    public function __construct(
        ?string $host = null,
        ?string $user = null,
        ?string $password = null,
        ?string $database = null,
        ?int $port = null,
        ?string $socket = null
    ) {}
```
Could you please clarify why you have changed this value?

mysqli/mysqli.php
```diff
@@ -216,14 +217,15 @@ public function close () {}
...
+	 * @link https://php.net/manual/en/function.mysqli-connect.php
+	 * @param string $host [optional]
+	 * @param string $user [optional]
+	 * @param string $password [optional]
+	 * @param string $database [optional]
+	 * @param int $port [optional]
+	 * @param string $socket [optional]
 	 */
-	public function connect ($host, $user, $password, $database, $port, $socket) {}
+	public function connect ($host = '', $user = '', $password = '', $database = '', $port = 0, $socket = '') {}
```
isfedorov, line 228: According to https://github.com/php/php-src/blob/master/ext/mysqli/mysqli.stub.php#L61 
Stub is
```
public function connect(
        ?string $host = null,
        ?string $user = null,
        ?string $password = null,
        ?string $database = null,
        ?int $port = null,
        ?string $socket = null
    ) {}
```
so I think default value should be changed to nulls

mysqli/mysqli.php
```diff
@@ -322,14 +324,15 @@ public function kill ($processid) {}
...
+	 * @link https://php.net/manual/en/mysqli.construct.php
 	 * @param string $host [optional]
 	 * @param string $user [optional]
 	 * @param string $password [optional]
 	 * @param string $database [optional]
 	 * @param int $port [optional]
 	 * @param string $socket [optional]
 	 */
-	public function mysqli ($host = null, $user = null, $password = null, $database = null, $port = null, $socket = null) {}
+	public function mysqli ($host = null, $user = null, $password = null, $database = null, $port = 0, $socket = null) {}
```
isfedorov, line 335: The same as above. Could you please clarify why 0 instead of null?

mysqli/mysqli.php
```diff
@@ -549,7 +552,7 @@ public function query ($query, $resultmode = MYSQLI_STORE_RESULT) {}
 	 * </p>
 	 * @return bool true on success or false on failure.
 	 */
-	public function real_connect ($host = null, $username = null, $passwd = null, $dbname = null, $port = null, $socket = null, $flags = null) {}
+	public function real_connect ($host = null, $username = null, $passwd = null, $dbname = null, $port = 0, $socket = null, $flags = null) {}
```
isfedorov, line 555: according to https://github.com/php/php-src/blob/master/ext/mysqli/mysqli.stub.php#L174 stub is
```
public function real_connect(
        ?string $host = null,
        ?string $user = null,
        ?string $password = null,
        ?string $database = null,
        ?int $port = null,
        ?string $socket = null,
        int $flags = 0
    ) {}
```
So I think default values should be updated accordingly

mysqli/mysqli.php
```diff
@@ -1521,11 +1528,11 @@ function mysqli_commit ($link, $flags = 0, $name = null) {}
  * @param string $user The MySQL user name.
  * @param string $password If not provided or NULL, the MySQL server will attempt to authenticate the user against those user records which have no password only.
  * @param string $database If provided will specify the default database to be used when performing queries.
- * @param string $port Specifies the port number to attempt to connect to the MySQL server.
+ * @param int $port Specifies the port number to attempt to connect to the MySQL server.
  * @param string $socket Specifies the socket or named pipe that should be used.
  * @return mysqli|false object which represents the connection to a MySQL Server or false if an error occurred.
  */
-function mysqli_connect ($host = '', $user = '', $password = '', $database = '', $port = '', $socket = '') {}
+function mysqli_connect ($host = '', $user = '', $password = '', $database = '', $port = 0, $socket = '') {}
```
isfedorov, line 1535: This function is an alias of https://github.com/php/php-src/blob/master/ext/mysqli/mysqli.stub.php#L61 so I think default values should be the same

mysqli/mysqli.php
```diff
@@ -2070,16 +2081,16 @@ function mysqli_query ($link, $query, $resultmode = MYSQLI_STORE_RESULT) {}
...
+ * @param string $user [optional]
+ * @param string $password [optional]
+ * @param string $database [optional]
+ * @param int $port [optional]
+ * @param string $socket [optional]
+ * @param int $flags [optional]
  * @return bool
  */
-function mysqli_real_connect ($link, $host = '', $user = '', $password = '', $database = '', $port = '', $socket = '', $flags = null) {}
+function mysqli_real_connect ($link, $host = '', $user = '', $password = '', $database = '', $port = 0, $socket = '', $flags = null) {}
```
isfedorov, line 2093: According to https://github.com/php/php-src/blob/master/ext/mysqli/mysqli.stub.php#L651 default values for options parameters are nulls. So I think it makes sense to update them accordingly.

PR #822, opened 2020-06-09
Unprefix Generator::throw() method

ondrejmirtes:
I don't know what the `PS_UNRESERVE_PREFIX_` prefix is for. Let's see if it was added to the `throw()` error by mistake or not.


isfedorov:
Hi @ondrejmirtes ,
this prefix is required for some of internal PhpStorm mechanisms and that's why it have to be here.

PR #831, opened 2020-06-14
Is subclass of

stdex:


isfedorov:
Hi @stdex 
Thank you for the contribution! Just a small note. It makes sense to make commit messages a bit more descriptive to make git log more clear and precise.

PR #837, opened 2020-06-16
Added ZipArchive::OPSYS_* constants

staabm:
list generated using

```php
<?php
echo '<pre>';

$refl = new ReflectionClass(ZipArchive::class);
$const = $refl->getConstants();
foreach($const as $name => $val) {
    if (strpos($name, 'OPSYS') === 0) {
       echo "
	/**
	 * @link https://www.php.net/manual/en/zip.constants.php#ziparchive.constants.opsys.default
	 * @since 5.6
	 */
	 const ". $name . " = ". $val. ";
	 ";
    }
}
```

### DIFF
--- a/zip/zip.php
+++ b/zip/zip.php
@@ -306,7 +306,133 @@ class ZipArchive implements Countable {
 	 */
 	const EM_AES_256 = 259;
 
-    /**
+	/**
+	 * @link https://www.php.net/manual/en/zip.constants.php#ziparchive.constants.opsys.default
+	 * @since 5.6
+	 */
+	 const OPSYS_DOS = 0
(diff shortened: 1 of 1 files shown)


staabm:
can someone take care of the CS problems? I guess you can easily do it with some automation?

isfedorov:
Hi @staabm Thank you for the contribution. To fix cs problems could you please run locally `composer cs` to check the code style and `composer cs-fix` to fix it and then push fixed files. Please see https://github.com/JetBrains/phpstorm-stubs/blob/master/CONTRIBUTING.md#code-style

staabm:
here we go

PR #840, opened 2020-06-16
Change uppercase NULL into lowercase null for saveHTML default value

schlessera:
This PR changes an inconsistency in the stubs for the `dom` extension.

The signature for `DOMDocument::saveHTML($node = NULL)` has the uppercase `null` as default value for its first and only `$node` argument.

The PR changes this to the lowercase `null` instead.

At the very least, this will make the file more consistent. However, I suspect that this is the cause for the PHPStan issue at https://github.com/phpstan/phpstan/issues/3478, so this will also hopefully fix that issue.

### DIFF
--- a/dom/dom_c.php
+++ b/dom/dom_c.php
@@ -964,7 +964,7 @@ class DOMDocument extends DOMNode  {
      * @param DOMNode $node [optional] parameter to output a subset of the document.
      * @return string the HTML, or false if an error occurred.
      */
-    public function saveHTML (DOMNode $node = NULL) {}
+    public function saveHTML (DOMNode $node = null) {}
 
     /**
      * Dumps the internal document into a file using HTML formatting


isfedorov:
@schlessera to say the truth I'm not sure that difference in cases of `null` can be the reason of inspection warning but even if it is then I think this is possibly a bug in phpstan as it interprets them differently. Anyway I'll merge this change at least to get rid of inconsistency in stubs. Regarding https://phpstan.org/r/f2176152-5861-4451-a9b4-da32795c2b8f and `Method ExtendedDocument::saveHTML() should return string but returns string|false.` it seems to be an issue in the stubs as method's return type is declared as `@return string the HTML, or false if an error occurred.` but should be `@return string|false the HTML, or false if an error occurred.` Your PR is welcomed! :)

ondrejmirtes:
This PR isn't necessary as `NULL` and `null` are functionally equivalent, and it doesn't actually fix the PHPStan issue :) But I guess it's good for the consistency perspective.

The PHPStan issue was about something completely else - turns out the actual runtime PHP reflection is missing information about the saveHTML parameter.

schlessera:
@isfedorov I followed this up with a PR to adapt the return types for this extension: https://github.com/JetBrains/phpstorm-stubs/pull/841

PR #844, opened 2020-06-17
The PHP constants true, false, and null MUST be written using the lower case

b1rdex:
This is a follow-up for #840


isfedorov:
Hi @b1rdex 
Thank you for the suggestion. I've merged related changes in scope of PR #849 so I'm closing this PR now.

PR #850, opened 2020-06-19
Fix for http_build_query

jordikroon:
According to the PHP documentation. The additional parameters for `http_build_query` are defined as:

```
http_build_query ( mixed $query_data [, string $numeric_prefix [, string $arg_separator [, int $enc_type = PHP_QUERY_RFC1738 ]]] ) : string
```

We recently found out that these parameters are not nullable when strict_types enabled. This PR fixes the error below:
```
Exception 'TypeError' with message 'http_build_query() expects parameter 2 to be string, null given'
```

The second parameter is set to `arg_separator.output`. According to the default PHP.ini. That value is in almost any case `&`: arg_separator.output	"&"	PHP_INI_ALL

### DIFF
--- a/standard/standard_2.php
+++ b/standard/standard_2.php
@@ -549,7 +549,7 @@ function rawurldecode ($str) {}
  *  <p>If enc_type is PHP_QUERY_RFC3986, then encoding is performed according to » RFC 3986, and spaces will be percent encoded (%20).
  * @return string a URL-encoded string.
  */
-function http_build_query ($query_data, $numeric_prefix = "", $arg_separator = "&", $enc_type = PHP_QUERY_RFC1738){}
+function http_build_query ($query_data, string $numeric_prefix = "", string $arg_separator = "&", int $enc_type = PHP_QUERY_RFC1738){}
 
 /**
  * Returns the target of a symbolic link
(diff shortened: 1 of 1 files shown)


isfedorov:
Hi @jordikroon 
Hm, I can't reproduce such problem with strict_types. https://3v4l.org/6Fnv3 
Could you please show an example when exception occurs?

jordikroon:
Hi @isfedorov 

It occurs when you set the second and third parameter to null.
https://3v4l.org/sQFSg

isfedorov:
Then this exception is correct. According to https://www.php.net/manual/en/function.http-build-query.php parameters have type `string` and not `?string` and as far as you pass nulls, you get this runtime exception and setting default values to string isn't enough. You still will have runtime exception. But it's possible to highlight `null` in parameters with error to indicate that expected type is `string` but got `null`. To fix highlighting please fix stub to `function http_build_query ($query_data, string $numeric_prefix = "", string $arg_separator = "&", $enc_type = PHP_QUERY_RFC1738){}` and will merge PR after that

jordikroon:
I have added the types. Also included the enc_type `int` hint for consistency.

Thank you!

PR #851, opened 2020-06-22
WI-53902 Make SimpleXMLElement implement Iterator

wbars:
Reuse methods signatures from SimpleXMLIterator since it's safe to assume that these methods will be used in iterator access for this class


wbars:
Merged in intellij sources manually

wbars:
+ https://github.com/JetBrains/phpstorm-stubs/commit/898607c8dde9a73211ec4c680b319afa82c636d6

PR #852, opened 2020-06-23
Bring pthreads stub in line with pthreads v3.2.0

dktapps:
This doesn't include pthreads' new Socket class; my objective here is primarily to fix issues with static analysis tools which mis-report errors on code that works fine with v3.
pthreads v2 has been obsolete for 5 years, and cannot be used with anything newer than 5.6, so I don't consider dropping its classes problematic; however, if it is a problem, then phpstorm-stubs needs some way of versioning stubs because pthreads v2 and pthreads v3 have some changes as well as removals.


isfedorov:
Hi @dktapps,
Please execute `php generate-stub-map` and commit the resulting `PhpStormStubsMap.php`

dktapps:
done, thanks!

PR #855, opened 2020-06-25
added xdebug_is_debugger_active()

staabm:
see https://xdebug.org/docs/remote#xdebug_is_debugger_active

### DIFF
--- a/PhpStormStubsMap.php
+++ b/PhpStormStubsMap.php
@@ -5241,6 +5241,7 @@ const FUNCTIONS = array (
   'xdebug_get_profiler_filename' => 'xdebug/xdebug.php',
   'xdebug_get_stack_depth' => 'xdebug/xdebug.php',
   'xdebug_get_tracefile_name' => 'xdebug/xdebug.php',
+  'xdebug_is_debugger_active' => 'xdebug/xdebug.php',
   'xdebug_is_enabled' => 'xdebug/xdebug.php',
   'xdebug_memory_usage' => 'xdebug/xdebug.php',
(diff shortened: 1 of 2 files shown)


staabm:
> The commited stub map is not up to date. Please regenerate it using ./generate-stub-map
As far as I can tell, the CI error is unrelated to my changes

isfedorov:
Hi @staabm ,
As failure message noted you have to re-generate and push stub-map. Please execute `php generate-stub-map` and commit the resulting `PhpStormStubsMap.php` as described in https://github.com/JetBrains/phpstorm-stubs#how-to-update-stub-map

staabm:
just regenerated the map.

for a better contributor experience it would be great
- if it would not be required to have php 7.4 to install the projects dependencies. I had to install php 7.4 especially for this task
- the stub generation could even be automatically be achieved via github action, so no manual step would be required

isfedorov:
Hi @staabm 
Thank you for the contribution and for the comments.
Regarding 
>if it would not be required to have php 7.4 to install the projects dependencies
I've created PR #862 and after it will be merged you will be able to perform `composer install` using composer from docker container. As I understand `composer install` was the reason why you had to install php 7.4 locally, right?
Regarding
>the stub generation could even be automatically be achieved
Actually the point of CI here is to check and notify only and don't change files. Final decision should be still on reviewer/developer. It's done this way to avoid potential commit of wrong files. The same approach is used for cs-fixer script. CI only run cs-fixer dry-run and notifies if files are formatted wrong.

PR #856, opened 2020-06-25
Data Structures: 3rd Amendment

dguhl:
A constructor for Vector and PriorityQueue's push method had been missing. Thanks and shoutout to the PHPSTAN project, @enumag, @jkuchar and @ondrejmirtes for their patience.

### DIFF
--- a/ds/ds.php
+++ b/ds/ds.php
@@ -1878,12 +1878,12 @@ namespace Ds {
         /**
          * @var mixed
          */
-        public $key = null;
+        public $key;
 
         /**
          * @var mixed
          */
-        public $value = null;
+        public $value;
 
         /**
          * Creates a new instance using a given key and value.
--- a/ds/ds.php
+++ b/ds/ds.php
@@ -1876,6 +1876,16 @@ namespace Ds {
     class Pair implements JsonSerializable
     {
         /**
+         * @var mixed
+         */
+        public $key = null;
+
+        /**
+         * @var mixed
+         */
+        public $value = null;
+
+        /**
          * Creates a new instance using a given key and value.
          *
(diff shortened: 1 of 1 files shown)


enumag:
@dguhl Still missing the properties of the Pair class (from https://github.com/phpstan/phpstan/issues/3539):

```
Access to an undefined property Ds\Pair::$key.
Access to an undefined property Ds\Pair::$value.
```

dguhl:
They're not documented.

enumag:
I see, but unlike with Map::getIterator() which is not documented [correctly](https://github.com/phpstan/phpstan/issues/3513#issuecomment-650022983), this is in fact a bug in the documentation.

dguhl:
I really oppose incorporating undocumented behavior. 🙁

enumag:
@dguhl See https://github.com/php-ds/ext-ds/issues/119 and https://bugs.php.net/bug.php?id=76687. This really is just a documentation bug.

ds/ds.php
```diff
@@ -1875,6 +1875,16 @@ public function xor(Map $map): Map
      */
     class Pair implements JsonSerializable
     {
+        /**
+         * @var mixed
+         */
+        public $key = null;
```
enumag, line 1881: The `= null` is useless

dguhl: It follows [implemented behavior](https://github.com/php-ds/ext-ds/blob/master/src/php/classes/php_pair_ce.c#L66-L67)?

enumag: null is the default value anyway so it means exactly the same, adding = null to a property only makes sense when using typed properties which is not the case here

PR #857, opened 2020-06-25
PHP 8.0 support

SerafimArts:
### RFC https://wiki.php.net/rfc/stringable

- [x] `Stringable` interface

### RFC https://wiki.php.net/rfc/attributes_v2
- [x] `ReflectionClassConstant::getAttributes()`

- [x] `ReflectionFunctionAbstract::getAttributes()`
- [x] `ReflectionParameter::getAttributes()`
- [x] `ReflectionClass::getAttributes()`

### RFC https://wiki.php.net/rfc/attribute_amendments

- [x] `Attribute` class
- [x] `Attribute::TARGET_xxx` constants
- [x] `Attribute::IS_REPEATABLE` constant

### RFC https://wiki.php.net/rfc/constructor_promotion

- [x] `ReflectionParameter::isPromoted()`
- [x] `ReflectionProperty::isPromoted()`

### RFC https://wiki.php.net/rfc/str_contains

- [x] Add `str_contains` function

### RFC https://wiki.php.net/rfc/add_str_starts_with_and_ends_with_functions

- [x] Add `str_starts_with` function
- [x] Add `str_ends_with` function

### RFC https://wiki.php.net/rfc/get_debug_type

- [x] Add `get_debug_type` function

### PR https://github.com/php/php-src/pull/5427

- [x] Add `get_resource_id` function

### RFC https://wiki.php.net/rfc/weak_maps

- [x] Add `WeakMap` class

### RFC https://wiki.php.net/rfc/dom_living_standard_api

- [x] Implement new signatures (TBD)

### RFC https://wiki.php.net/rfc/token_as_object

- [x] Add `PhpToken` class

### ReflectionClassConstant visibility constants

I did not find links to these additions

- [x] `ReflectionClassConstant::IS_PUBLIC`
- [x] `ReflectionClassConstant::IS_PROTECTED`
- [x] `ReflectionClassConstant::IS_PRIVATE`

## Reflection API Changes

- Added `.phpstorm.meta.php` for Reflection autocomplete improvements
- Fixed errors
  - Removed nonexistent `ReflectionType::__clone`
  - Removed `final` keyword of `ReflectionReference`
  - and something else
- Other
  - All classes are divided by files (should I revert this changes?)
  - All methods and fields are marked with types (including new `mixed` and union types)
  - Added lost documentation descriptions and links
  - All type references replaced by `{@see xxx}` (adds navigation between types in PhpStorm)


Core/Core_c.php
```diff
@@ -660,3 +660,90 @@ public static function create(object $referent): WeakReference {}
      */
     public function get(): ?object {}
 }
+
+/**
+ * Stringable interface marks classes as available for serialization
+ * in a string.
+ *
+ * @since 8.0.0
```
isfedorov, line 668: Could you please change version to 8.0 here and below because 0 patch version doesn't make any sense and we've got rid of 0 patch version in stubs recently.

Core/Core_c.php
```diff
@@ -660,3 +660,90 @@ public static function create(object $referent): WeakReference {}
...
+/**
+ * @since 8.0.0
+ */
+// TODO Uncomment after PHP 8.0 release:
+// @@Attribute(Attribute::TARGET_CLASS)
+final class Attribute {
+    /**
+     * Marks that attribute declaration is allowed only in classes.
+     *
+     * @var int
```
isfedorov, line 689: Seems declaring type here doesn't make any sense as type can be easily inferred. `@var` tag here is just a noisy info. Could you please remove redundant `@var` tags in constants?

Core/Core_c.php
```diff
@@ -660,3 +660,90 @@ public static function create(object $referent): WeakReference {}
...
+    /**
+     * Notes that an attribute declaration in the same place is
+     * allowed multiple times.
+     *
+     * @var int
+     */
+    public const IS_REPEATABLE = 1 << 10;
+
+    /**
+     * @param int $flags
```
isfedorov, line 744: PhpDoc is redundant and can be safely removed

SerafimArts: In the PSR it is written that the code should be uniform.

Therefore, I put down docblocks for all types, even where it is not needed in order to maintain this uniformity.

Is removal definitely required?

Core/Core_c.php
SerafimArts: https://github.com/JetBrains/phpstorm-stubs/pull/857/commits/d5d6f1ec6b6fdc2f722fd8473d01f0dbbbc2b50c

Core/Core_c.php
SerafimArts: https://github.com/JetBrains/phpstorm-stubs/pull/857/commits/484f9fcc81fc9b3bc0afdd19b5878d0bb748769d

isfedorov:
Ok, currently next tests fail:
```
There were 3 failures:

1) StubTests\StubsTest::testClasses with data set "class ReflectionType" (StubTests\Model\PHPClass Object (...))

Missing method ReflectionType::__clone(){}

Failed asserting that an array has the key '__clone'.

/opt/project/phpstorm-stubs/tests/StubsTest.php:131

2) StubTests\StubsTest::testInterfaces with data set "interface Reflector" (StubTests\Model\PHPInterface Object (...))

Missing method Reflector::__toString(){}

Failed asserting that an array has the key '__toString'.

/opt/project/phpstorm-stubs/tests/StubsTest.php:253

3) StubTests\StubsTest::testClassesPHPDocs with data set "interface Stringable" (StubTests\Model\PHPInterface Object (...))

class Stringable has 

                    'since' version 8.0.0.'Since' version for PHP Core functionallity for style consistensy 

                    should have X.X format for the case when patch version is '0'.

Failed asserting that false is true.
```
And
```
The commited stub map is not up to date. Please regenerate it using ./generate-stub-map
```
And
```
Files that were not fixed due to errors reported during linting before fixing:

   1) /home/travis/build/JetBrains/phpstorm-stubs/Reflection/ReflectionMethod.php
   2) /home/travis/build/JetBrains/phpstorm-stubs/Reflection/ReflectionProperty.php
   3) /home/travis/build/JetBrains/phpstorm-stubs/Reflection/ReflectionParameter.php
   4) /home/travis/build/JetBrains/phpstorm-stubs/Reflection/ReflectionClassConstant.php
   5) /home/travis/build/JetBrains/phpstorm-stubs/Reflection/ReflectionFunction.php
   6) /home/travis/build/JetBrains/phpstorm-stubs/Reflection/ReflectionReference.php
   7) /home/travis/build/JetBrains/phpstorm-stubs/Reflection/ReflectionFunctionAbstract.php
   8) /home/travis/build/JetBrains/phpstorm-stubs/Reflection/ReflectionClass.php

Script php-cs-fixer fix -v --diff --dry-run handling the cs event returned with error code 4
```
So to fix the problems could you please 
1) Add method `ReflectionType::__clone()`
2) Add method `Reflector::__toString()` (Despite Reflector extends Stringable, tests currently use php 7.4 and thus don't have Stringable interface.)
3) Change `8.0.0` to `8.0` in `Stringable` interface
4) Execute `php generate-stub-map` and commit the resulting `PhpStormStubsMap.php`
5) Run `composer cs-fix` and commit fixed by cs-fixer files

Thank you in advance!

Reflection/ReflectionExtension.php
```diff
@@ -0,0 +1,187 @@
...
+ * The <b>ReflectionExtension</b> class reports
+ * information about an extension.
+ *
+ * @link https://php.net/manual/en/class.reflectionextension.php
+ */
+class ReflectionExtension implements Reflector
+{
+    /**
+     * Name of the extension, same as calling the {@see ReflectionExtension::getName()}
+     * method. Read-only, throws {@see ReflectionException} in attempt to write.
```
isfedorov, line 13: Could you please clarify why you have added doc that property is read-only (provide point in src)? Can't find such information in docs or src.

SerafimArts:
> Add method ReflectionType::__clone()

I'm sorry, I focused on the documentation: https://www.php.net/manual/en/class.reflectiontype.php

There is no `__clone` method. However, introspection shows the existence of this method. Fixed: https://github.com/JetBrains/phpstorm-stubs/pull/857/commits/b1e61a70fe7fbaeb5aeb866ad9f254c1398e72bf

At the same time I discovered some new functionality that was changed, but not reflected in the documentation. Method `ReflectionType::isBuiltin` was removed: https://github.com/JetBrains/phpstorm-stubs/pull/857/commits/fce6473a0eb8a8174ffe973093dffaadd0e8d531

Reflection/ReflectionExtension.php
SerafimArts: https://3v4l.org/7fTgK

This information is not in the documentation. But I checked that all fields of Reflection classes is readonly (like `name` or `class`).

SerafimArts:
> Add method Reflector::__toString() (Despite Reflector extends Stringable, tests currently use php 7.4 and thus don't have Stringable interface.)

Fixed: https://github.com/JetBrains/phpstorm-stubs/pull/857/commits/0ac7d7a82be1d12a96434af214a7f71d522402e7

SerafimArts:
This error will always occur: https://travis-ci.org/github/JetBrains/phpstorm-stubs/builds/702999388#L1078
Until the version of PHP is updated to 8.0 (https://travis-ci.org/github/JetBrains/phpstorm-stubs/builds/702999388#L208)

However, updating the version is currently not possible, as the cs-fixer is not compatible with more recent versions of the language: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.16/composer.json#L17

However, we can use phpcbf, which supports any modern version of PHP: https://github.com/squizlabs/PHP_CodeSniffer/blob/master/composer.json#L28 

But when using this fixer, it will fix the entire code)))
image

isfedorov:
@SerafimArts Oh, I see.. The reason is union/mixed type hints. I suggest to get rid of them for now then and leave union/mixed types only in PhpDoc for now. Thus we will make tests green and more over as far as these stubs are used by some other tools we will avoid potential problems caused by these type hints in those tools as well.

Reflection/ReflectionAttribute.php
```diff
@@ -0,0 +1,63 @@
+<?php
+
+/**
+ * @since 8.0
+ */
+class ReflectionAttribute
```
isfedorov, line 6: Please correct me if I'm wrong but seems several methods are missed in this class . For example https://github.com/php/php-src/blob/master/ext/reflection/php_reflection.c#L6460 . Or please confirm if it was intended to skip those methods for now

SerafimArts: This RFC was merged recently: https://wiki.php.net/rfc/attribute_amendments

At the time of preparing this PR, these methods were not yet in the master branch. So yes, several methods have been skipped. Some I added marked "TODO" comment.

Since they are already merged, it makes sense to add them already.

Reflection/ReflectionExtension.php
isfedorov: Hm, may be in this case it make sense to define property via `@property-read` in class' PhpDoc? Thus it will be strictly defined as read-only.

Reflection/ReflectionAttribute.php
SerafimArts: Added https://github.com/JetBrains/phpstorm-stubs/pull/857/commits/bda34ef1fd712f5f8bf2f3cde0eb29889aaa1d5a

SerafimArts:
@isfedorov Does it make sense to remove the errors of stat analysis using annotations? Like:
```php
    /**
     * ....
     * @noinspection PhpHierarchyChecksInspection
     * @noinspection PhpSignatureMismatchDuringInheritanceInspection
     */
    public static function export($argument, bool $return = false): ?string {}
```

Reflection/ReflectionClass.php
```diff
@@ -0,0 +1,669 @@
...
+    /**
+     * Gets an array of methods for the class.
+     *
+     * @link https://php.net/manual/en/reflectionclass.getmethods.php
+     * @param int $filter Filter the results to include only methods
+     * with certain attributes. Defaults to no filtering.
+     * @return ReflectionMethod[] An array of {@see ReflectionMethod} objects
+     * reflecting each method.
+     */
+    public function getMethods(int $filter = 119): array
```
isfedorov, line 215: What does this magic number mean? Also doc says that "Defaults to no filtering." so seems it should be null

SerafimArts: > What does this magic number mean

This is a bitmask of all available values: `RefelctionMethod::IS_STATIC | RefelctionMethod::IS_PUBLIC | etc...`

Reflection/ReflectionClass.php
```diff
@@ -0,0 +1,669 @@
...
+    /**
+     * Gets properties
+     *
+     * @link https://php.net/manual/en/reflectionclass.getproperties.php
+     * @param int $filter The optional filter, for filtering desired
+     * property types. It's configured using the {@see ReflectionProperty} constants,
+     * and defaults to all property types.
+     * @return ReflectionProperty[]
+     */
+    public function getProperties(int $filter = 23): array
```
isfedorov, line 251: The same questions as above about magic number

Reflection/ReflectionClass.php
SerafimArts: Yes, indeed, the argument may be null. My bad)

SerafimArts: Fixed: https://github.com/JetBrains/phpstorm-stubs/pull/857/commits/0e9a2fcdfa3dfe65f18a4cb69d42475245a1d05a

SerafimArts:
I ended up with the implementation of all the functionality that is currently available in PHP 8.0. Since the last review have been added:
- WeakMap https://github.com/JetBrains/phpstorm-stubs/pull/857/commits/e7fa6531bcf058fa458da795e7acbc2148b607e0
- PhpToken https://github.com/JetBrains/phpstorm-stubs/pull/857/commits/3f18d715e161e797eb3b4382a63ebe04c4c8045c
- DOM Living Standard https://github.com/JetBrains/phpstorm-stubs/pull/857/commits/957a52acd02c67dd2f72996412b18a983b145003

Reflection/ReflectionExtension.php
SerafimArts: In addition to the existing?

isfedorov: No, instead of existing. Property still will be available in completion and resolve and in the same time it will be resolved as read-only so seems this is better fit for the case.

SerafimArts: These fields cannot contain an extended description, like`@link https://www.php.net/manual/en/class.reflectionclass.php#reflectionclass.props.name`. Only in this format:

```php
/**
 * @property-read string $name Name of the class.
 */
class ReflectionClass implements Reflector
```

isfedorov: So instead of 
```
class ReflectionClass{
public string $name;
}
```
It will be
```
/**
* @property-read $name
*/
class ReflectionClass{}
```
As Example see `EventBuffer::length`

isfedorov: In this case read-only property will be highlighted on write attempt 
Screenshot 2020-06-30 at 00 44 16

SerafimArts: I know, however, I repeat that for such fields it is impossible to add any descriptions except a short one. That is, I should delete all the links to the documentation that were there before.

isfedorov: Well, it seems that read-only fields ("name" and "class" only?) don't have really large extended description and we still can add the links as inline tags though. Probably highlighting of potentially failing code is more important then showing several additional lines of docs in this case. And of course I'm talking here only about read-only fields in changed Reflection classes. All other properties should still stay as is.

SerafimArts: Ok that's quite reasonable: https://github.com/JetBrains/phpstorm-stubs/pull/857/commits/1cc29782e47637cf9b8fee85da35a4f266a43cea

Reflection/ReflectionParameter.php
```diff
@@ -0,0 +1,283 @@
...
+     * Exports
+     *
+     * @link https://php.net/manual/en/reflectionparameter.export.php
+     * @param string $function The function name.
+     * @param string $parameter The parameter name.
+     * @param bool $return Setting to {@see true} will return the export,
+     * as opposed to emitting it. Setting to {@see false} (the default) will do the
+     * opposite.
+     * @return string|null The exported reflection.
+     * @deprecated Since PHP 7.4 and removed in 8.0
```
isfedorov, line 36: Could you please remove  `Since PHP` words before version and split this into two separate tags
```
@deprecated 7.4
@removed 8.0
```
Otherwise deprecation will not be highlighted properly by PhpStorm

SerafimArts: `@removed` is a nonexistent tag. 

I mean that this is not in the specification: https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc.md and implementations: https://docs.phpdoc.org/latest/references/phpdoc/tags/index.html

Reflection/ReflectionType.php
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    /**
+     * Checks if it is a built-in type
+     *
+     * @link https://php.net/manual/en/reflectiontype.isbuiltin.php
+     * @return bool Returns {@see true} if it's a built-in type, otherwise {@see false}
+     * @since 7.0
+     *
+     * @deprecated Since 8.0 this method has been removed from the ReflectionType
```
isfedorov, line 29: Could you please remove `Since` word before version
`@deprecated 8.0`
Otherwise deprecation will not be highlighted properly by PhpStorm due to wrong parsing of this comment

Reflection/ReflectionType.php
```diff
@@ -0,0 +1,56 @@
...
+    {
+    }
+
+    /**
+     * To string
+     *
+     * @link https://php.net/manual/en/reflectiontype.tostring.php
+     * @return string Returns the type of the parameter.
+     * @since 7.0
+     * @deprecated Since PHP 7.1. Please use {@see ReflectionType::getName()} instead.
```
isfedorov, line 42: The same as above, could you please remove `Since PHP` in deprecation

SerafimArts: Done.

At the same time, I did it for LDAP too.

SerafimArts:
I did not break the tests! Honestly! They fell themselves :3

Reflection/Reflection.php
```diff
@@ -27,7 +27,7 @@ public static function getModifierNames(int $modifiers): array
      * opposed to emitting it. Setting to {@see false} (the default) will do the opposite.
      * @return string|null If the return parameter is set to {@see true}, then the
      * export is returned as a string, otherwise {@see null} is returned.
-     * @deprecated Since PHP 7.4 and removed in 8.0
+     * @deprecated 7.4 and removed in 8.0
```
isfedorov, line 30: Still..please split to 
```
@deprecated 7.4
@removed 8.0
```
PhpStorm has dedicated support to properly highlight entities with `@removed` tag based on language level

Reflection/ReflectionType.php
```diff
@@ -39,7 +39,7 @@ public function isBuiltin(): bool
      * @link https://php.net/manual/en/reflectiontype.tostring.php
      * @return string Returns the type of the parameter.
      * @since 7.0
-     * @deprecated Since PHP 7.1. Please use {@see ReflectionType::getName()} instead.
+     * @deprecated 7.1. Please use {@see ReflectionType::getName()} instead.
```
isfedorov, line 42: Could you please remove trailing dot in `7.1.` as it's currently parsed by tests as wrong version. We will update tests to properly process such cases but for now let's just remove the dot.

Reflection/Reflection.php
SerafimArts: As u wish https://github.com/JetBrains/phpstorm-stubs/pull/857/commits/5ecd612d071a18bf6ff435f87de4626d835538af

Reflection/ReflectionType.php
SerafimArts: Done https://github.com/JetBrains/phpstorm-stubs/pull/857/commits/d4ecbfd63e3ab81974a8a5ee35cfb04ec0ea1815

isfedorov:
@SerafimArts 
>@isfedorov Does it make sense to remove the errors of stat analysis using @noinspection annotations? Like:

We have discussed this internally and that's what we've got: base method `Reflector::export()` is removed in php-7.4. Probably based method isn't used really often in the existing code. So probably we can just remove base method `Reflector::export()` from stubs as redundant and this way we also will resolve PhpHierarchyChecksInspection and PhpSignatureMismatchDuringInheritanceInspection warnings in child classes.
What do you think about this way?

SerafimArts:
We can also transfer it to the docblock section:
image
However, the syntax of dockblocks is limited in applicability. 

Perhaps it’s worth considering some sort of advanced, based on attributes, allowing to describe any cases? For example, cannot mark this docblock method as deprecated. It is now impossible to describe several methods with different signatures (method overloading). Or at the same time both a static and an instance method, etc...

But it is, thoughts out loud)

isfedorov:
Well, we have performed some additional testing and seems we will have to leave this method as is to avoid breaks in old php versions. More over those tests showed that it's need to get rid of changes with return typehints, scalar type hints in parameters and constant visibilities in old classes and methods due to such changes will break code with language level < 7. Statistics shows that 57% of users use php 5.6 language level.
image (1)
@SerafimArts that's pity, but could you please revert all the changes related to return typehints, scalar typehints in parameters and const visibilities in old classes and methods and leave such features only for entities marked with `since > 7`?

SerafimArts:
> Statistics shows that 57% of users use php 5.6 language level.

Its default language level of PhpStorm, lol))) Just change to the latest one and there will be a completely different picture.

isfedorov:
@SerafimArts My screenshot above shows results of PhpStorm internal statistic that we gather based on value set by users in PhpStorm.

SerafimArts:
I repeat once again that 5.6 is the default language level version for all new projects. So, which is set immediately. These users simply did not change it ;)

isfedorov:
@SerafimArts you are right. This is default value but tricky part here is that if people don't change default language level then 1)They use php 5.6 or 2) They don't use features of php 7. (otherwise code would be red). Generally based on stats a lot of people don't have local interpreter at all as the run code on the server. For example there is stat like https://w3techs.com/technologies/details/pl-php that shows that approximately a half of usages is php 5. Even based on stats that you have referenced summary php usage of versions < 7 is quite large and based on definition that by adding features of php >7 we make code red for those ~30-40% of users and not adding new features we loose mostly nothing it seems that the latter is better.

SerafimArts:
Sounds logical. I'll do. But you still set the max version in the next release, otherwise I're already tired of constantly switching)

isfedorov:
@SerafimArts 
>otherwise I're already tired of constantly switching

You can set default language level in `File->New Project Settings->Preferences For New Projects` and this language level will be set as default in all new projects.
Regarding PR seems it's good to go now and I'll squash and merge it.

SerafimArts:
> You can set default language level in File->New Project Settings->Preferences For New Projects and this language level will be set as default in all new projects.

lol

atehnix:
I think that in PhpStorm there should not be an outdated version of PHP by default.
At the moment there should be PHP 7.3 (Active support) or PHP 7.2 (Security support).
This is at least. Ideally, of course, the latest released version.

It's about the PHP version used when creating a new project. Therefore, if someone creates new projects for a very outdated version of PHP, then it will not be a surprise for him that the language has changed a lot over the years.
Using the current version is the norm. Using an outdated version is an exceptional case.

PR #862, opened 2020-06-28
Add composer to Dockerfile

isfedorov:
This PR is follow-up of comments in #855
To avoid local installation of required in composer php version, I've added composer to Dockerfile so now php isn't required to be installed locally at all.
To run `composer install` and run tests, remote interpreter created from Dockerfile can be used.
Generally I've copied official Dockerfile for composer from https://github.com/composer/docker/tree/master/1.9 to our Dockerfile and updated it to install required for tests extensions


b1rdex:
You could've used multi-stage build to simplify it: `COPY --from=composer /usr/bin/composer /usr/bin/composer`

isfedorov:
Thank you @b1rdex ! I've updated Dockerfile

PR #864, opened 2020-07-07
Parameter $message is optional.

umairkhan-dev:
See first example in [examples](https://github.com/phpredis/phpredis#example-8) of [PING](https://github.com/phpredis/phpredis#ping) section.
```
/* When called without an argument, PING returns `TRUE` */
$redis->ping();
```

### DIFF
--- a/redis/Redis.php
+++ b/redis/Redis.php
@@ -385,7 +385,7 @@ class Redis
      * @throws RedisException
      * @link    https://redis.io/commands/ping
      */
-    public function ping($message)
+    public function ping($message=null)
     {
     }
 
--- a/redis/Redis.php
+++ b/redis/Redis.php
@@ -378,7 +378,7 @@ class Redis
     /**
      * Check the current connection status
      *
-     * @param string $message
+     * @param string $message [optional]
      *
      * @return bool|string TRUE if the command is successful or returns message
      * Throws a RedisException object on connectivity error, as described above.


redis/Redis.php
```diff
@@ -378,7 +378,7 @@ public function getOption($option)
     /**
      * Check the current connection status
      *
-     * @param string $message
+     * @param string $message [optional]
```
staabm, line 381: I guess the parameter in the actual function below needs to be defined with a null-default value, or similar?

umairkhan-dev: I think you are right, but as `php-redis` is `pecl` package, so I could not check its function defination.
However I have performed a small test by making a child class of `\Redis`.
```
var_dump($this->ping(), $this->ping(null));
```
and they both output same result.
```
bool(true)
bool(true)
```

staabm: Ping is a `redis_opt_str_cmd`

https://github.com/phpredis/phpredis/blob/08f202e775037ccf849d7b933dddb467c9c2ee5f/redis.c#L1228

And those seem to accept null: https://github.com/phpredis/phpredis/blob/a311cc4ec3cecdbaf83ba66985efa82137e37cc0/redis_commands.c#L170

umairkhan-dev: I have added `null` as default value for `$message`. Updated PR.

pestretsov:
Thank you!

PR #865, opened 2020-07-09
Fix Closure::bind return type

Jean85:


b1rdex:
Could you please do the same for `bindTo`?

b1rdex:
Also would be great to change the `newthis` type to `object|null` for both. Sorry :)

Jean85:
No problem, done!

PR #868, opened 2020-07-10
Add stubs for phpseclib constants

kesselb:
https://github.com/phpseclib/phpseclib 

PHPStorm (and static analysis tools like psalm) are unable to resolve constants like `NET_SFTP_HANDLE` used by phpseclib because they are defined at runtime: https://github.com/phpseclib/phpseclib/search?q=define_array&unscoped_q=define_array


pestretsov:
Hi @kesselb!
Thanks for your time! Unfortunately we don't accept stubs for libraries for now. I can suggest two options:
1) Create PR to the lib's repo and define constants in the library itself
2) Use your changes to stubs and provide PHPStorm with custom stubs path (Here is how: https://blog.jetbrains.com/phpstorm/2018/03/how-to-provide-stubs-for-phpstorm/)

kesselb:
Hi @pestretsov,

thanks for your review. I will suggest it to phpseclib/phpseclib.

PR #871, opened 2020-07-16
Clarify unserialize() spec

oldpec:
Second parameter can only be an array, the true/false/array values allowed are on an option inside the array
https://www.php.net/manual/en/function.unserialize.php

### DIFF
--- a/standard/standard_4.php
+++ b/standard/standard_4.php
@@ -151,7 +151,7 @@ function serialize ($value) {}
  * @param array $options [optional]
  * <p>Any options to be provided to unserialize(), as an associative array.</p>
  * <p>
- * The 'allowed_classes' option key may be set to a value that is 
+ * The 'allowed_classes' option key may be set to a value that is
  * either an array of class names which should be accepted, FALSE to
  * accept no classes, or TRUE to accept all classes. If this option is defined
  * and unserialize() encounters an object of a class that isn't to be accepted,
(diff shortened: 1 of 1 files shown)


felixdorn:
Have you ever used `unserialize` ? This is alreafy correct.

oldpec:
> Have you ever used `unserialize` ?

@felixdorn I am glad you are a student because here is a lesson for you.
Being wrong is ok, sometimes there's something you just don't understand, it happens to us all from time to time.
Being rude is not ok. It doesn't help anything. 
And when you are both wrong and rude, like you are here, is just embarrassing.

> This is already correct.
The current stub doc isn't even internally consistent ("@param mixed" for the second parameter in the phpdoc and "array" in the function spec) so it can't be "correct".

```
% php72 -a
Interactive shell
php > $temp=serialize('hello');
php > unserialize($temp,false);
Warning: unserialize() expects parameter 2 to be array, boolean given in php shell code on line 1
```

```
% php74 -a
Interactive shell

php > $temp=serialize('hello');
php > unserialize($temp,false);

Warning: unserialize() expects parameter 2 to be array, bool given in php shell code on line 1
```

```
php > $temp=serialize('hello');
php > unserialize($temp);
php > unserialize($temp,['allowed_classes'=>false]);
php > unserialize($temp,['allowed_classes'=>true]);
//Try this one, it might help
php > unserialize($temp,['allowed_classes'=>['rudeCommentFilter']]);
```


```
php > $obj=new \Exception();
php > $temp=serialize($obj);
php > $des=unserialize($temp,['rudeCommentFilter']); //If you have been doing this, you have been doing it wrong
php > echo get_class($des);
Exception
php > $des=unserialize($temp,['allowed_classes'=>['rudeCommentFilter']]);//This is how it works
php > echo get_class($des);
__PHP_Incomplete_Class
php > 

```

PR #877, opened 2020-07-29
Document argument as nullable

greg0ire:
This was overlooked in #698 and causes some issues for static analysis,
see https://github.com/doctrine/persistence/pull/122


pestretsov:
Thanks!

PR #883, opened 2020-08-06
Fix incorrect parameter name signatures for dba_open and dba_popen

TysonAndre:
These are variadic functions in php 7 and 8

https://wiki.php.net/rfc/named_params will make parameter names of
internal functions part of the API, so having placeholder names instead
of matching names for methods/functions is something that
should probably be changed in many, many places.
(e.g. standard/standard_9.php)

Related to https://youtrack.jetbrains.com/issue/WI-54703 - this process of supporting named arguments should probably be automated in a script that renames parameters in signatures and doc comments


slapoguzov:
Thanks!

PR #884, opened 2020-08-07
Fix @return for class_parents, class_implements and class_uses

InvisibleSmiley:


SPL/SPL_f.php
```diff
@@ -102,7 +102,7 @@ function spl_autoload_call ($class_name) {}
  * the __autoload magic
  * method.
  * </p>
- * @return array An array on success, or false on error.
+ * @return array|false An array on success, or false on error.
```
b1rdex, line 105:
```suggestion
 * @return string[]|false An array on success, or false on error.
```

SPL/SPL_f.php
```diff
@@ -117,7 +117,7 @@ function class_parents ($class, $autoload = null) {}
...
```
b1rdex, line 120:
```suggestion
 * @return string[]|false An array on success, or false on error.
```

SPL/SPL_f.php
```diff
@@ -177,7 +177,7 @@ function iterator_apply (Traversable $iterator, $function, array $args = null) {
  * Return the traits used by the given class
  * @param mixed $class An object (class instance) or a string (class name).
  * @param bool $autoload Whether to allow this function to load the class automatically through the __autoload() magic method.
- * @return array Array on success, or FALSE on error.
+ * @return array|false Array on success, or FALSE on error.
```
b1rdex, line 180:
```suggestion
 * @return string[]|false An array on success, or false on error.
```

slapoguzov:
Thank you!

PR #885, opened 2020-08-07
Fix `false on error` returns

b1rdex:


mcrypt/mcrypt.php
```diff
@@ -363,7 +363,7 @@ function mcrypt_generic_deinit ($td) {}
  * @param resource $td <p>
  * The encryption descriptor.
  * </p>
- * @return int|bool If the self test succeeds it returns false. In case of an error, it
+ * @return bool If the self test succeeds it returns false. In case of an error, it
```
slapoguzov, line 366: According to the documentation and source code (https://www.php.net/manual/en/function.mcrypt-enc-self-test.php https://github.com/php/pecl-encryption-mcrypt/blob/master/mcrypt.c#L743) this function returns `int`

mcrypt/mcrypt.php
```diff
@@ -363,7 +363,7 @@ function mcrypt_generic_deinit ($td) {}
...
  * returns true.
```
b1rdex, line 367:
```suggestion
 * @return int Returns 0 on success and a negative integer on failure
```

PR #886, opened 2020-08-07
WI-46268 Fix getallheaders()

radiat-r:
This tries to solve https://youtrack.jetbrains.com/issue/WI-46268

Despite having apache in it's name the function apache_request_headers() is not only available with apache.

It's documentation mentions FastCGI, FPM and the CLI server:
See https://www.php.net/manual/en/function.apache-request-headers.php

By looking at the php sourcecode, it is defined in:
sapi/apache2handler/php_functions.c
sapi/cgi/cgi_main.c
sapi/cli/php_cli_server.c
sapi/fpm/fpm/fpm_main.c

Also, it is defined as an alias in:
sapi/litespeed/lsapi_main.c

getallheaders() is an alias to apache_request_headers() and defined as such in all of the above mentioned places as well.

I've moved apache_request_headers() and getallheaders() from apache/apache.php to standard/standard_4.php, because that's where the other header related functions are.

As an additional note, standard/_standard_manual.php contains the stub for fastcgi_finish_request() which in php-src is only defined in sapi/fpm/fpm/fpm_main.c. It would probably make sense to move it into fpm/fpm.php then? I did this in the second commit, but that one is completely optional. Feel free to omit that one if you disagree.


slapoguzov:
Thank you!

yoosefi:
thorough. thanks!

PR #889, opened 2020-08-13
hash() can return false

baukevdw:
When passing an invalid hashing algorithm to `hash()` an error will be triggered and hash will return `false`

```php
<?php
// Passing intentionally an invalid algorithm
var_dump(hash('sha-256', 'Hello world!')); // bool(false)


slapoguzov:
`hash()` will throw an exception in such cases in PHP 8.0 : https://github.com/php/php-src/commit/d5871e26ad95a8022aa68d0ca95d05398bf78a41 , https://3v4l.org/lsfHb

PR #891, opened 2020-08-15
Wrong usage of @ since

BenMorel:
`@since` should be reserved for the case when the function has been *added* in a given version, not modified.

Because of this, PHPStorm considers `bcmod()` to be available since PHP 7.2, when it was actually available before then:

image

### DIFF
--- a/bcmath/bcmath.php
+++ b/bcmath/bcmath.php
@@ -81,16 +81,16 @@ function bcdiv ($dividend, $divisor, $scale = 0) {}
  * Get modulus of an arbitrary precision number
  * @link https://php.net/manual/en/function.bcmod.php
  * @param string $dividend <p>
- * The dividend, as a string.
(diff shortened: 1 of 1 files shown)


bcmath/bcmath.php
```diff
@@ -94,7 +94,6 @@ function bcdiv ($dividend, $divisor, $scale = 0) {}
  * </p>
  * @return string|null the modulus as a string, or <b>NULL</b> if
  * <i>divisor</i> is 0.
- * @since 7.2 scale param added, dividend/divisor no longer truncated
```
slapoguzov, line 97: I suggest moving this information to the corresponding parameter blocks, but without `@`

BenMorel: @slapoguzov Done!

slapoguzov:
Thank you!

PR #892, opened 2020-08-15
add / fix some phpdocs

voku:


interbase/interbase.php
```diff
@@ -656,15 +655,11 @@ function ibase_server_info ($service_handle, $action) {}
...
- * </p>
- * @param string $event_name2 [optional] <p>
- * </p>
- * @param string $_ [optional]
+ * @param string ...$event_names <p>
+ * The event names.
  * @return string the name of the event that was posted.
  */
-function ibase_wait_event ($event_name1, $event_name2 = null, $_ = null) {}
+function ibase_wait_event (...$event_names) {}
```
slapoguzov, line 662: I can't find any proofs of this. Have you checked this?

voku: Yes, but it's not 100% correct in both cases. :-/

https://github.com/php/pecl-database-interbase/blob/2044c06d94126115b38de5487cc2119a45aad6bb/ibase_events.c#L125

```
ibase_wait_event ( string ...$event_names) : string
ibase_wait_event ( resource $link_identifier, string ...$event_names) : string
```

standard/standard_4.php
```diff
@@ -172,13 +171,12 @@ function unserialize (string $str, array $options = []) {}
...
  * @link https://php.net/manual/en/function.var-dump.php
- * @param mixed $expression <p>
+ * @param mixed ...$expressions <p>
  * The variable you want to export.
  * </p>
- * @param mixed $_ [optional]
  * @return void
  */
-function var_dump ($expression, $_ = null) {}
+function var_dump (...$expressions) {}
```
slapoguzov, line 179: it's not the same because it also means that I can use this function without any arguments but I can't. Could you also check other similar changes?

voku: You are right, checked and reverted. :+1:

slapoguzov:
@voku please, take a look at tests, seems some tests are broken after adding `&` to phpdocs

voku:
@slapoguzov I tried to fix this problem upstream (https://github.com/phpDocumentor/ReflectionDocBlock/pull/254) or is this behavior expected?

slapoguzov:
@voku I think it's a bug in phpDocumentor. I suggest excluding changes that break tests from this MR and merge it. And you can open a new MR with these changes when the problem in phpDocumentor is resolved.

voku:
@slapoguzov the bug in phpDocumentor is fixed :)

PR #893, opened 2020-08-18
IntlBreakIterator::createCharacterInstance - optional argument

MorrisJobke:
The argument of IntlBreakIterator::createCharacterInstance is optional. See https://www.php.net/manual/en/intlbreakiterator.createcharacterinstance.php and https://github.com/php/php-src/blob/f78a09101496d05b4fb50fe37b7dbf5d038452ec/ext/intl/breakiterator/breakiterator_arginfo.h#L5


slapoguzov:
Thank you!

PR #897, opened 2020-08-18
WI-54952: fix rdkafka stubs, generator

nick-zh:
Moved functions where needed, to reflect the the state of the extension:
https://github.com/arnaud-lb/php-rdkafka/blob/master/rdkafka.c#L726-L743

As per the below discussion, also adjusted the stub generator to recurse into sub-folders, since this caused some issues without it


nick-zh:
Is it possible that nested directories are not recognized?
I updated `PhpStormStubsMap.php` as per readme, but the result does not seem to differ (which in my case makes sense probably), but i was only able to find the files in the root of rdkafka

ondrejmirtes:
Well you haven't added or removed any files so I don't see why `PhpStormStubsMap.php` should change.

ondrejmirtes:
But you're right that classes from RdKafka are missing in the map.

nick-zh:
So i moved all the classes to the root folder, so they are present in the stubs map, i hope that is ok

ondrejmirtes:
Might be worth fixing the generator script instead, shouldn't be hard. Maybe it finds other namespaces with the same problem.

nick-zh:
Agreed, i fixed it accordingly :v:

ondrejmirtes:
Awesome, more stubs available :)

nick-zh:
@slapoguzov @andrey-sokolov please let me know if any more changes / adjustments are needed :v:

slapoguzov:
thank you!

PR #899, opened 2020-08-18
add model for inferring the type of function callback arguments

i582:
This model allows to describe the types of arguments for callback functions depending on the arguments of the calling function.
[WI-54960](https://youtrack.jetbrains.com/issue/WI-54960)

### DIFF
--- a/meta/.phpstorm.meta.php
+++ b/meta/.phpstorm.meta.php
@@ -653,6 +653,43 @@ namespace PHPSTORM_META {
     exitPoint(\wp_die());
     exitPoint(\dd());
 
+    /**
+     * @param callable $callable Class, Method or function call
+     * @see callback()
+     * @see arg()
+     * @return mixed
+     */
+    function registerCallbackTypes($callable) {
+      return "registerCallbackTypes $callable";
+    }
+
+    /**
+     * description of the callback arguments
+     * @param mixed $args
+     * @see arg()
+     * @return mixed
+     */
+    function callback(...$args) {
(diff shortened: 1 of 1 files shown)


meta/.phpstorm.meta.php
```diff
@@ -653,6 +653,43 @@ function exitPoint($functionReference) {
...
+     * @param mixed $args
+     * @see arg()
+     * @return mixed
+     */
+    function callback(...$args) {
+      return "callback ${$args}";
+    }
+
+    /**
+     * argument type #$argNum of the caller callback
```
YuriyNasretdinov, line 677: This needs to be a type of an _element_, not the type of the argument (argument to array_map is array, and the callbacks accept an element of an array)

YuriyNasretdinov (changes requested):

meta/.phpstorm.meta.php
i582: Yes, it is, thanks.

b1rdex:
I've already added a comment on the issue page and saying that IMO it's better to get support for generics (https://youtrack.jetbrains.com/issue/WI-47158) rather than create a new proprietary thing.

slapoguzov:
Thanks for your contribution. However, this is a new syntax for `.phpstorm.meta.php` and we have to consider these changes more detailed. Please,[ create a new ticket at our tracker](https://youtrack.jetbrains.com/newIssue?project=WI) and describe your suggestions there. You also can attach this mr to the ticket.

PR #907, opened 2020-09-17
Update urls to HTTPS

J0WI:
Created with https://github.com/EFForg/https-everywhere/blob/master/utils/rewriter/rewriter.js


MaXal:
Sorry but what is the main point of adding `www`?

J0WI:
> Sorry but what is the main point of adding `www`?

Consistency. `php.net` also redirects `www.php.net`.

MaXal:
Unfortunately, I don't see enough value in such a massive patch so we decline the PR. In case, there will be a user-visible or tools difference in behavior, we will reconsider this!

PR #910, opened 2020-09-28
ReflectionProperty::getType() returns ReflectionType

BenMorel:
In PHP 8, `ReflectionProperty::getType()` returns either a `ReflectionNamedType` or a `ReflectionUnionType`:

https://3v4l.org/3X0Rp

So the correct return value is `ReflectionType`, just like [ReflectionParameter::getType()](https://github.com/JetBrains/phpstorm-stubs/blob/874def36f96dbb7db2eb2ff5e47ff54ac3ea34fc/Reflection/ReflectionParameter.php#L129-L139).

### DIFF
--- a/Reflection/ReflectionProperty.php
+++ b/Reflection/ReflectionProperty.php
@@ -213,7 +213,7 @@ class ReflectionProperty implements Reflector
      * Gets property type
      *
      * @link https://php.net/manual/en/reflectionproperty.gettype.php
-     * @return ReflectionNamedType|null Returns a {@see ReflectionNamedType} if the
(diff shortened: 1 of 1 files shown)


isfedorov:
Please see https://youtrack.jetbrains.com/issue/WI-54056

BenMorel:
While it is safe to say that it will always be a `ReflectionNamedType` in PHP <= 7.4, it can also be a `ReflectionUnionType` in PHP 8 if union types are used. Please see the 3v4l link above.

So the correct stub type is the common parent, `ReflectionType`.

isfedorov:
@BenMorel fair enough!

PR #911, opened 2020-09-28
WI-55725: make second argument to imagexbm optional

MasterOdin:
Closes https://youtrack.jetbrains.com/issue/WI-55725

Per the [PHP Doc](https://www.php.net/manual/en/function.imagexbm.php), the second argument is defined as `[$filename = NULL]` and where if not set, will cause the function to output the stream versus sending it to the set filename.

### DIFF
--- a/gd/gd.php
+++ b/gd/gd.php
@@ -2062,7 +2062,7 @@ function imagecolormatch ($image1, $image2) {}
  * Output XBM image to browser or file
  * @link https://php.net/manual/en/function.imagexbm.php
  * @param resource $image
- * @param string $filename <p>
+ * @param string $filename [optional] <p>
  * The path to save the file to. If not set or null, the raw image stream
  * will be outputted directly.
  * </p>
@@ -2073,7 +2073,7 @@ function imagecolormatch ($image1, $image2) {}
  * </p>
  * @return bool true on success or false on failure.
  */
-function imagexbm ($image, $filename, $foreground = null) {}
+function imagexbm ($image, $filename = null, $foreground = null) {}
 
 /**
  * Applies a filter to an image


AlexMovsesov:
Hi, thx for your contribution! 
I agree with that by the [PHP Doc](https://www.php.net/manual/en/function.imagexbm.php) the second argument for the `imagexbm` is optional, but I see warnings while running this function without the `$filename` argument.
Also, there are the following problems:
By PHP's stubs: 
[GD stubs](https://github.com/php/php-src/blob/0a84fba0deb1c1b75770a436c4236dc56e6d0463/ext/gd/gd.stub.php#L103) 
there is a nullable nonoptional `$filename` parameter, 
By PHP's sources:
[ARG INFO](https://github.com/php/php-src/blob/0a84fba0deb1c1b75770a436c4236dc56e6d0463/ext/gd/gd_arginfo.h#L200-L204)
[Check](https://github.com/php/php-src/blob/0a84fba0deb1c1b75770a436c4236dc56e6d0463/ext/gd/gd.c#L1879) of the arguments.
For this syntax you can find the doc [here](https://github.com/php/php-src/blob/0a84fba0deb1c1b75770a436c4236dc56e6d0463/docs/parameter-parsing-api.md)

From what PHP's version you're not receiving warnings?

MasterOdin:
I was not running the code, rather I was getting the warning through tooling that I use that's based on this library. I was not running this function directly, nor do I think the library I inherited that uses this function has ever had to deal with an xbm image now.

You are right that running the function without the second argument throws a PHP Warning about the missing argument. I've filed an upstream bug report (https://bugs.php.net/bug.php?id=80168) for this.

MasterOdin:
Based on response on that ticket, it looks to be a documentation issue for 7.x (and then 8.0 will make it truly optional). Closing this PR as it is not valid then.

PR #912, opened 2020-10-01
WI-54641 Add type declarations to magic method signatures

pestretsov:
https://wiki.php.net/rfc/magic-methods-signature

### DIFF
--- a/standard/_types.php
+++ b/standard/_types.php
@@ -350,7 +350,6 @@ class object {
     /**
      * This method is called by var_dump() when dumping an object to get the properties that should be shown.
      * If the method isn't defined on an object, then all public, protected and private properties will be shown.
-     * @since 5.6
      *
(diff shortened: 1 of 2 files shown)


b1rdex:
Why no `mixed` type for arguments and return values?

pestretsov:
@b1rdex it's actually not real `mixed`. Real `mixed` doesn't allow void return types (and this is how PHPStorm's treats declared `mixed`, search for `mixed|void` in https://wiki.php.net/rfc/mixed_type_v2). However, it's allowed to have `void` return type declaration: https://3v4l.org/IqCnR

pestretsov:
The real checks performed by PHP Engine are here: https://github.com/php/php-src/pull/4177/files#diff-246862e485897a77cdcef1c52d473a52

PR #913, opened 2020-10-05
Update to Php8rc1 and align parameter names

isfedorov:
Please pay attention to added muted problems and added test that compares parameter names in reflection and stubs.

### DIFF
--- a/tests/TestData/mutedProblems.json
+++ b/tests/TestData/mutedProblems.json
@@ -1532,6 +1532,41 @@
           ]
         }
       ]
+    },
+    {
+      "name": "Directory",
+      "methods": [
+        {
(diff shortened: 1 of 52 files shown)


tests/TestData/mutedProblems.json
```diff
@@ -1532,6 +1532,41 @@
...
+        },
+        {
+          "name": "read",
+          "problems": [
+            "parameter mismatch"
+          ]
+        }
+      ]
+    },
+    {
```
MaXal: This should be removed and the stub should be changed to `private __construct()`.

MaXal (changes requested):

PR #918, opened 2020-10-12
Provide #[ArrayShape] for some functions

wbars:


orklah:
Oh, that's cool!

cc @voku

Core/Core.php
```diff
@@ -1,6 +1,7 @@
 <?php
 
 // Start of Core v.5.3.6-13ubuntu3.2
+use JetBrains\PhpStorm\ArrayShape;
```
mfn, line 4: 👀  where does that come from 😅

b1rdex: https://github.com/JetBrains/phpstorm-stubs/blob/master/meta/attributes/ArrayShape.php

b1rdex:
Wouldn't it break php < 8.0 runtime?

cc @ondrejmirtes

wbars:
@b1rdex phpstorm-stubs are not meant to be executed by PHP, it serves mostly static analysis purpose.

PR #919, opened 2020-10-13
ob_implicit_flush php 8 signature

b1rdex:
https://github.com/php/php-src/commit/46d22e435f43529aa55b54460dd265c4bd22409f


isfedorov:
Hi @b1rdex 
Tests failed with message 
```
1) StubTests\StubsTest::testFunctions with data set "function ob_implicit_flush" (StubTests\Model\PHPFunction Object (...))

Function ob_implicit_flush has signature ob_implicit_flush($enable)but stub function has signature ob_implicit_flush($flag)
```
This test checks that parameter names in stubs are equal to parameter names in PHP (based on reflection) to avoid runtime errors with named parameters. Please see https://youtrack.jetbrains.com/issue/WI-54703 for details

b1rdex:
The test failed because it ran on php 7.4. It was renamed to $enable in 8.0-rc1: https://github.com/php/php-src/blob/php-8.0.0RC2/ext/standard/basic_functions.stub.php#L40

MaXal:
The test is run on `php:8.0.0rc1-alpine`, maybe the change didn't get in?

isfedorov:
>It was renamed to $enable in 8.0-rc1:
That's right. But in your change you changed parameter name from `$enable` to `$flag` and as far as tests run with php8-rc1 test fails.

b1rdex:
@isfedorov haha you're right. IDK how it happened. Will fix soon.

PR #920, opened 2020-10-14
Add \GdImage class

b1rdex:
https://php.watch/versions/8.0/gdimage

### DIFF
--- a/gd/gd.php
+++ b/gd/gd.php
@@ -83,7 +83,7 @@ function gd_info () {}
 /**
  * Draws an arc
  * @link https://php.net/manual/en/function.imagearc.php
- * @param resource $image
+ * @param resource|GdImage $image
  * @param int $cx <p>
  * x-coordinate of the center.
  * </p>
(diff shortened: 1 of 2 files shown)


gd/gd.php
```diff
@@ -3013,3 +3013,8 @@ function imagescale ($image, $new_width, $new_height = -1, $mode = IMG_BILINEAR_
  * @since 5.5
  */
 function imagesetinterpolation ($image, $method = IMG_BILINEAR_FIXED) {}
+
+/**
+ * @since 8.0
+ */
+final class GdImage{}
```
MaXal, line 3020: Should we add a private constructor to enforce: `Cannot be instantiated with new \GdImage().` ?

MaXal (changes requested):

gd/gd.php
b1rdex: Decide yourself. I did it the same as `CurlHandler` – just a class definition.

b1rdex: FYI: they aren't cloneable too: https://github.com/php/php-src/blob/master/ext/gd/tests/gdimage_prevent_cloning.phpt

b1rdex: But CurlHandle *is* cloneable: https://github.com/php/php-src/blob/master/ext/curl/tests/curl_handle_clone.phpt (sure why not 🤣)

PR #924, opened 2020-10-19
Add Socket and AddressInfo resource objects

b1rdex:
https://php.watch/versions/8.0/sockets-sockets-addressinfo


MaXal:
Looks like a safe change from the backward compatibility point of view so I merge it. In case, some issues arise, we will have to change `Socket` to `Socket|resource`.

PR #927, opened 2020-10-21
Migrate @property-read to #[Immutable]

MaXal:


Ev/Ev.php
```diff
@@ -387,14 +389,20 @@ final class Ev
...
- * @property-read bool $is_pending TRUE if the watcher is pending, i.e. it has outstanding events, but its callback
- *      has not yet been invoked. FALSE otherwise. As long, as a watcher is pending (but not active), one must not
- *      change its priority.
  */
 abstract class EvWatcher
 {
+    /**
+     * @var bool TRUE if the watcher is active. FALSE otherwise.
+     */
+    #[Immutable] public $is_active;
```
b1rdex, line 398: PHP 7 will skip the property definition because `#` means comment. Shouldn't the attribute definition and the property definition be split to two lines?

wbars: @b1rdex phpstorm-stubs are not meant to be executed by PHP, it serves mostly static analysis purpose.

Reflection/ReflectionClassConstant.php
```diff
@@ -1,16 +1,25 @@
...
  *
- * @property-read string $name Constant name, same as calling the {@see ReflectionClassConstant::getName()} method
- * @property-read string $class Fully qualified class name where this constant was defined
- *
  * @link https://www.php.net/manual/en/class.reflectionclassconstant.php
  * @since 7.1
  */
 class ReflectionClassConstant implements Reflector
 {
+    /**
```
wbars, line 13: Let's provide real type signatures for all properties where applicable

PR #932, opened 2020-10-23
Update Reflector.php

WinterSilence:
mistake in comment to `__toString()`

### DIFF
--- a/Reflection/Reflector.php
+++ b/Reflection/Reflector.php
@@ -9,10 +9,10 @@
 interface Reflector extends Stringable
 {
     /**
-     * Exports a class
+     * Exports a class.
      *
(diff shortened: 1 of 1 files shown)


Reflection/Reflector.php
```diff
@@ -21,9 +21,6 @@ public static function export();
     /**
      * Returns the string representation of any Reflection object.
      *
-     * Please note that since PHP 8.0 this method is absent in this interface
```
MaXal, line 24: Could you please leave this part as is? It's correct and clarifies the existence of the method.

MaXal (changes requested):

Reflection/Reflector.php
WinterSilence: @MaXal it's not correct

PR #933, opened 2020-10-25
Why there is no releases in a year?

vladyslavstartsev:
Hello. 
Thank you for making this repository.
I was wondering, why there is no releases in a year?

I mean, this repo is relatively active but doesn't make releases.

The reason why I'm asking is because I don't understand why do you delay the next one.


MaXal:
@vladyslavstartsev Thank you for the request! We've just forgotten to publish 2020.1 and 2020.2. I've added `v2020.2` tag which will mark stub's state up until the attributes introduction.

PR #934, opened 2020-10-26
Set correct return type for SimpleXMLElement::__get()

AJenbo:
If SimpleXMLElement is extended, it's children are also instances of the extending class.


MaXal:
@AJenbo Unfortunately, I can't accept your request since there is a bug in PhpStorm: https://youtrack.jetbrains.com/issue/WI-56660 which will completely break type inference on the dynamic properties of SimpleXML.
I will reopen and merge your request as soon as it's fixed.

PR #935, opened 2020-10-28
Update yaf.php add response method

ljyf5593:
Update yaf.php add response method

### DIFF
--- a/yaf/yaf.php
+++ b/yaf/yaf.php
@@ -1722,14 +1722,16 @@ abstract class Yaf_Response_Abstract {
 
     /**
      * @link https://secure.php.net/manual/en/yaf-response-abstract.response.php
+     * 
+     * @return void
      */
     public function response(){ }
 
     /**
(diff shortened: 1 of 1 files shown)


yaf/yaf.php
```diff
@@ -1720,6 +1720,22 @@ private function __clone(){ }
      */
     public function __toString(){ }
 
+    /**
+     * @link https://secure.php.net/manual/en/yaf-response-abstract.response.php
```
MaXal, line 1724: Please add `Send response` to the description and `@return void` for clarity.

yaf/yaf.php
```diff
@@ -1720,6 +1720,22 @@ private function __clone(){ }
...
+     */
+    public function response(){ }
+
+    /**
+     * @link https://secure.php.net/manual/en/yaf-response-abstract.setheader.php
```
MaXal, line 1729: Please add `Set reponse header` to the description.

MaXal (changes requested):

PR #940, opened 2020-10-29
Added Direct IO Functions stubs 

elcreator:
https://www.php.net/manual/en/ref.dio.php


elcreator:
While documentation says this is for php5 it was just moved to PECL but continues work with the same API
https://github.com/php/pecl-system-dio/blob/master/php7/php_dio.h

MaXal:
Please run `generate-stub-map` and add the update of `PhpStormStubsMap.php` to the PR.

PR #944, opened 2020-11-01
Fix parse error in Core.php

ondrejmirtes:


ondrejmirtes:
Sorry, realized I'm parsing this with an old parser.

PR #945, opened 2020-11-01
Fixed parse error in Core_c.php

ondrejmirtes:


ondrejmirtes:
Sorry, realized I'm parsing this with an old parser.

PR #948, opened 2020-11-04
Fix phpDoc for setEncryptionIndex

xel1045:
The `$method` property should be `int` since all `ZipArchive::EM_` constants are `int`.

Possible `EM_` values are:

```php
	/**
	 * No encryption
	 * @link https://secure.php.net/manual/en/zip.constants.php
	 * @since 7.2
	 */
	const EM_NONE = 0;

	/**
	 * AES 128 encryption
	 * @link https://secure.php.net/manual/en/zip.constants.php
	 * @since 7.2
	 */
	const EM_AES_128 = 257;

	/**
	 * AES 192 encryption
	 * @link https://secure.php.net/manual/en/zip.constants.php
	 * @since 7.2
	 */
	const EM_AES_192 = 258;

	/**
	 * AES 256 encryption
	 * @link https://secure.php.net/manual/en/zip.constants.php
	 * @since 7.2
	 */
	const EM_AES_256 = 259;
```

### DIFF
--- a/zip/zip.php
+++ b/zip/zip.php
@@ -881,7 +881,7 @@ class ZipArchive implements Countable {
      * Set the encryption method of an entry defined by its index
      * @link https://php.net/manual/en/ziparchive.setencryptionindex.php
      * @param int $index Index of the entry.
-     * @param string $method The encryption method defined by one of the ZipArchive::EM_ constants.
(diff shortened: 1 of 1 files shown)


xel1045:
Oops, the problem seems to be related to PHP.

PR #949, opened 2020-11-07
some more phpdoc fixes

voku:
This change is [Reviewable](https://reviewable.io/reviews/jetbrains/phpstorm-stubs/949)

### DIFF
--- a/Core/Core.php
+++ b/Core/Core.php
@@ -317,7 +317,7 @@ function error_reporting ($error_level = null) {}
  * @param string $constant_name <p>
  * The name of the constant.
  * </p>
- * @param mixed $value <p>
+ * @param null|array|bool|int|float|string $value <p>
  * The value of the constant.
  * In PHP 5, value must be a scalar value (integer, float, string, boolean, or null).
  * In PHP 7, array values are also accepted.
@@ -378,7 +378,7 @@ function get_called_class () {}
 /**
  * Retrieves the parent class name for object or class
  * @link https://php.net/manual/en/function.get-parent-class.php
- * @param mixed $object_or_class [optional] <p>
+ * @param object|string $object_or_class [optional] <p>
  * The tested object or class name
  * </p>
  * @return string|false <p>The name of the parent class of the class of which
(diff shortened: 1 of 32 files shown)


Core/Core.php
```diff
@@ -762,14 +761,14 @@ function create_function ($args, $code) {}
  * @param resource $resource <p>
  * The evaluated resource handle.
  * </p>
- * @return string If the given <i>handle</i> is a resource, this function
+ * @return string|false If the given <i>handle</i> is a resource, this function
```
isfedorov, line 764: Hi @voku ,
Could you please clarify about this change? As I see this function always return string https://github.com/php/php-src/blob/master/Zend/zend_builtin_functions.c#L1404

b1rdex:
AFAIK `array<key-type, value-type>` syntax is already supported in the latest EAP. So I guess it might be used in the stubs too.
Here is what I found to be working in my code:
```
list<type>
array<vtype>
array<ktype, vtype>
non-empty-list<vtype>
non-empty-array<vtype>
non-empty-array<ktype, vtype>
```

Would be nice to see some comment from the team.

wbars:
> AFAIK `array<key-type, value-type>` syntax is already supported in the latest EAP. So I guess it might be used in the stubs too.
> Here is what I found to be working in my code:
> 
> ```
> list<type>
> array<vtype>
> array<ktype, vtype>
> non-empty-list<vtype>
> non-empty-array<vtype>
> non-empty-array<ktype, vtype>
> ```
> 
> Would be nice to see some comment from the team.

@b1rdex This syntax is provided by 3rd party tools, while phpstorm-stubs project depends only on PHP itself and PhpStorm internals.

orklah:
> @b1rdex This syntax is provided by 3rd party tools, while phpstorm-stubs project depends only on PHP itself and PhpStorm internals.

Can we imagine psalm and phpstan plugin parsing the stubs too? The basic syntax would stay in @param/return for everyone and those with the plugins installed could enjoy more detailed stubs in @psalm/phpstan tags ?

b1rdex:
https://blog.jetbrains.com/phpstorm/2020/10/phpstorm-2020-3-eap-2/
Paragraph "Types support" says PhpStorm supports part of the mentioned syntax.

PR #951, opened 2020-11-11
Class constants were moved from DateTime to DateTimeInterface in PHP 7.2

ondrejmirtes:
This means that the constants do not exist on DateTimeImmutable in PHP 7.1 for example.

### DIFF
--- a/date/date_c.php
+++ b/date/date_c.php
@@ -4,18 +4,69 @@
  * @since 5.5
  */
 interface DateTimeInterface {
+    /**
+     * @since 7.2
+     */
     const ATOM = 'Y-m-d\TH:i:sP';
+    
+    /**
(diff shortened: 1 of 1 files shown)


wbars:
https://youtrack.jetbrains.com/issue/WI-56949

orklah:
FYI, this has the side effect of displaying `DateTime::ATOM` as removed in PhpStorm.

It still works like a charm:
https://3v4l.org/lMs6J

ondrejmirtes:
@orklah Looks like a bug in PhpStorm. Since the constant is on DateTimeInterface on PHP 7.2 and newer, PhpStorm should pick that up.

orklah:
Yep, I agree, that's why I mentionned that was just a side effect

PR #954, opened 2020-11-13
floor() and ceil() never return false in PHP 8

jlherren:


orklah:
Seems like they could before PHP8:
https://3v4l.org/XWcND

jlherren:
Oh well... Can this be considered an unlikely edge case, which is reported by any decent IDE and static analyzer? Otherwise discard this PR.

orklah:
I guess it's up to the maintainers too decide but it's a delicate subject. By removing this in the stubs, it's like saying that it can never happen. Which means every code that handle this false value, even for edge cases, will be flagged as an error in IDE.

MaXal:
@jlherren Please use `#[PhpStormStubsElementAvailable]` attribute. See: https://github.com/JetBrains/phpstorm-stubs/blob/20d254690b577ed20677e9ec9aa4a81367295811/enchant/enchant.php#L7 for example.

MaXal:
We've addressed this situation using `#[LanguageLevelTypeAware(["8.0" => "float"], default: "float|false")]`

PR #957, opened 2020-11-18
Remove wrong @since annotation that confuses the engine

ondrejmirtes:
Originally reported at: https://github.com/phpstan/phpstan/issues/4095


MaXal:
`@since` for stubs outside of core can be freely used since they don't necessarily indicate language version but rather a version of library.

PR #962, opened 2020-11-22
Small fixes

voku:


Core/Core_c.php
```diff
@@ -273,7 +273,7 @@ class Exception implements Throwable {
      * @link https://php.net/manual/en/exception.construct.php
      * @param string $message [optional] The Exception message to throw.
      * @param int $code [optional] The Exception code.
-     * @param Throwable $previous [optional] The previous throwable used for the exception chaining.
+     * @param null|Throwable $previous [optional] The previous throwable used for the exception chaining.
```
isfedorov, line 276: Please also update the type to nullable in the method signature

Core/Core_c.php
```diff
@@ -369,7 +369,7 @@ class Error implements Throwable {
      * @link https://php.net/manual/en/error.construct.php
      * @param string $message [optional] The Error message to throw.
      * @param int $code [optional] The Error code.
-     * @param Throwable $previous [optional] The previous throwable used for the exception chaining.
+     * @param null|Throwable $previous [optional] The previous throwable used for the exception chaining.
```
isfedorov, line 372: The same as above

Ev/Ev.php
```diff
@@ -806,7 +805,7 @@ final class EvPeriodic extends EvWatcher
      *      numerical stability).
      * @param float $interval The current interval value. Can be modified any time, but changes only take effect when
      *      the periodic timer fires or EvPeriodic::again() is being called.
-     * @param callable $reschedule_cb If set, tt must return the next time to trigger, based on the passed time value
+     * @param null|callable $reschedule_cb If set, tt must return the next time to trigger, based on the passed time value
```
isfedorov, line 808: Please also update the type to nullable in the method signature

Ev/Ev.php
```diff
@@ -845,7 +844,7 @@ public function at() {}
...
```
isfedorov, line 847: The same as above

PDO/PDO.php
```diff
@@ -884,7 +884,7 @@ public function __construct ($dsn, $username = null, $password = null, $options
 	 * Emulated prepared statements does not communicate with the database server
 	 * so <b>PDO::prepare</b> does not check the statement.
 	 */
-	public function prepare ($query, array $options = array()) {}
+	public function prepare ($query, array $options = []) {}
```
isfedorov, line 887: Please rollback such changes because they are still code-style related.

Phar/Phar.php
```diff
@@ -729,18 +729,18 @@ public function stopBuffering () {}
 	 * (PHP &gt;= 5.3.0, PECL phar &gt;= 2.0.0)<br/>
 	 * mapPhar for web-based phars. front controller for web applications
 	 * @link https://php.net/manual/en/phar.webphar.php
-	 * @param string $alias [optional] <p>
+	 * @param null|string $alias [optional] <p>
```
isfedorov, line 732: Seems these types also can be updated in the method signature

SPL/SPL.php
```diff
@@ -1097,8 +1097,8 @@ class RegexIterator extends FilterIterator {
      * Create a new RegexIterator
      * @link https://php.net/manual/en/regexiterator.construct.php
      * @param Iterator $iterator The iterator to apply this regex filter to.
-     * @param string $regex The regular expression to match.
-     * @param int $pattern [optional] Operation mode, see RegexIterator::setMode() for a list of modes.
+     * @param string $pattern [optional] The regular expression to match.
```
isfedorov, line 1100: Could you please clarify why this parameter is marked as optional?

voku: The parameter "$regex" is not available in the __constructor, it was renamed from "$regex" to "$pattern"

Phar/Phar.php
voku: I am not sure if I should use ```?string $alias = null```or ```string $alias = null``` here?

PDO/PDO.php
voku: ok

SPL/SPL.php
isfedorov: The question is why `$pattern` is marked as `[optional]`? As I can see in docs and PHP src this parameter isn't optional. `$regex` previously also was not optional

Phar/Phar.php
isfedorov: `?string $alias = null`. Please note that a similar change should be done for other parameters in the method

SPL/SPL.php
voku: Sorry, you are right... was a copy&past error.

Phar/Phar.php
voku: ok

SQLite/SQLite.php
```diff
@@ -748,9 +748,9 @@ function sqlite_fetch_array ($result, $result_type = SQLITE_BOTH, $decode_binary
  * Fetches the next row from a result set as an object
  * @link https://php.net/manual/en/function.sqlite-fetch-object.php
  * @param resource $result
- * @param string $class_name [optional]
- * @param array $ctor_params [optional]
- * @param bool $decode_binary [optional]
+ * @param null|string $class_name [optional]
```
isfedorov, line 751: Seems these types also can be added as type hints to the function signature

bcmath/bcmath.php
```diff
@@ -150,13 +151,22 @@ function bcpow (string $num, string $exponent, ?int $scale = 0): string
 function bcsqrt (string $num, ?int $scale)
 {}
 
+#[PhpStormStubsElementAvailable(to: '7.2')]
```
MaXal, line 154: This should be `to: '7.3'`

bcmath/bcmath.php
```diff
@@ -150,13 +151,22 @@ function bcpow (string $num, string $exponent, ?int $scale = 0): string
...
- * The scale factor. Since 7.3.0 can be omitted.
- * </p>
- * @return int|true <b>INT</b> since 7.3.0 and <b>TRUE</b> before.
+ * @param int $scale
+ * @return true
+ */
+function bcscale (int $scale): bool
+{}
+
+#[PhpStormStubsElementAvailable('7.3')]
```
MaXal, line 164: This should be `8.0`.

standard/standard_3.php
```diff
@@ -739,6 +739,9 @@ function fmod(float $num1, float $num2): float
  * Performs a floating-point division under
  * IEEE 754 semantics. Division by zero is considered well-defined and
  * will return one of Inf, -Inf or NaN.
+ * @param float $num1
```
MaXal, line 742: Empty `@param`s tag can be avoided in case there is a parameter type in the signature.

xml/xml.php
```diff
@@ -382,7 +383,20 @@ function xml_parse_into_struct (#[LanguageLevelTypeAware(["8.0" => "XmlParser"],
...
+ * Get XML parser error code
+ * @link https://php.net/manual/en/function.xml-get-error-code.php
+ * @param XmlParser $parser <p>
+ * A reference to the XML parser to get error code from.
+ * </p>
+ * @return int Returns one of the error codes listed in the error codes
+ * section.
+ */
+#[Pure]
+function xml_get_error_code (XmlParser $parser): int {}
```
MaXal, line 399: What do you think about reusing `LanguageLevelTypeAware` here on argument and function to avoid duplication?

MaXal:
Please also rebase on the master so there are conflicts.

bcmath/bcmath.php
isfedorov: According to docs - yes, but according to https://3v4l.org/Ptm7Q seems that null can be passed successfully even with versions < 8 so I think it makes sense just to make type nullable and apply LanguageLevelTypeAware for return type only according to https://3v4l.org/o0UXs

isfedorov: @voku @MaXal what do you think about such solution? Leave just one function with nullable parameter and return type that will have appropriate LanguageLevelTypeAware attribute.

Ev/Ev.php
isfedorov: `callable $reschedule_cb = null` still isn't nullable in `createStopped` function

dom/dom.php
```diff
@@ -9,7 +9,7 @@
  * @param SimpleXMLElement $node <p>
  * The <b>SimpleXMLElement</b> node.
  * </p>
- * @return DOMElement|null|false The <b>DOMElement</b> node added or <b>FALSE</b> if any errors occur.
+ * @return DOMElement|null The added <b>DOMElement</b> node.
```
isfedorov, line 12: This change seems to be incorrect. `simplexml_load_string('</>');` will return false so seems the correct way would be to change return type-hint instead of phpdoc

dom/dom_c.php
```diff
@@ -128,13 +128,13 @@ class DOMNode  {
...
      * </p>
-     * @param DOMNode $child [optional] <p>
+     * @param null|DOMNode $child [optional] <p>
      * The reference node. If not supplied, newnode is
      * appended to the children.
      * </p>
      * @return DOMNode The inserted node.
      */
-    public function insertBefore (DOMNode $node, DOMNode $child = null) {}
+    public function insertBefore (DOMNode $node, ?DOMNode $child = null) {}
```
isfedorov, line 137: There is a tricky case here. Class DOMNode is available since PHP 5 but nullable types are available only since PHP 7.1 so if one will try to overwrite this method in PHP < 7.1 then one will get a warning `Declaration should be compatible with DOMNode->insertBefore` and thus such changes should be avoided. Nullable type hints can be added only in final classes and methods that can't be overwritten.

bcmath/bcmath.php
voku: @isfedorov https://3v4l.org/HP28H I think PHP 8.0 will be correct here

voku: I added three different versions. I hope "PhpStormStubsElementAvailable" is used correctly  now.

standard/standard_3.php
voku: Should I revert it?

xml/xml.php
voku: The parameter, the return and the description are different, so it feels like something different. But I don't know how you want to handle that in general, is there a general rule for this?

zmq/zmq.php
```diff
@@ -702,9 +702,9 @@ class ZMQDevice
      *
      * @param ZMQSocket $frontend Frontend parameter for the devices. Usually where there messages are coming.
      * @param ZMQSocket $backend Backend parameter for the devices. Usually where there messages going to.
-     * @param ZMQSocket $listener Listener socket, which receives a copy of all messages going both directions. The type of this socket should be SUB, PULL or DEALER.
+     * @param null|ZMQSocket $listener Listener socket, which receives a copy of all messages going both directions. The type of this socket should be SUB, PULL or DEALER.
      */
-    public function __construct(ZMQSocket $frontend, ZMQSocket $backend, ZMQSocket $listener = null)
+    public function __construct(ZMQSocket $frontend, ZMQSocket $backend, ?ZMQSocket $listener = null)
```
isfedorov, line 707: As I wrote in one of the comments Nullable type hints can be added only in final classes and methods that can't be overridden. Otherwise, users who use PHP < 7.1 and override such methods will get warnings.  Class ZMQDevice isn't final so here should be no `?` in the type hint

xml/xml.php
isfedorov: It's possible to add two `LanguageLevelTypeAware` attributes: for return and for parameter, like `
#[LanguageLevelTypeAware(["8.0" => "int"], default: "int|false")]
function xml_get_error_code (#[LanguageLevelTypeAware(["8.0" => "XmlParser"], default: "resource")] $parser): int|false {}
`
and update description a bit with a note regarding PHP 8

xdebug/xdebug.php
```diff
@@ -351,7 +352,7 @@ function xdebug_get_gc_total_collected_roots(): int {}
  * @param int   $listType
  * @param array $configuration
```
isfedorov, line 353: I think it makes sense to update the return tag in phpdoc according to function signature

xdebug/xdebug.php
```diff
@@ -188,30 +189,30 @@ function xdebug_break (): bool {}
  * Use the PHP function getcwd() to figure out what the current working directory is.
  * @param $trace_file
```
isfedorov, line 190: Let's add types of parameters to phpdoc as well

standard/standard_1.php
```diff
@@ -960,16 +961,37 @@ function trim (string $string, string $characters = " \t\n\r\0\x0B"): string
...
+ * You can use the optional second parameter to specify tags which should
+ * not be stripped.
+ * </p>
+ * <p>
+ * HTML comments and PHP tags are also stripped. This is hardcoded and
+ * can not be changed with allowable_tags.
+ * </p>
+ * @return string the stripped string.
+ */
+#[Pure]
```
isfedorov, line 981: Seems here also `LanguageLevelTypeAware` can be used instead of function duplication

sodium/sodium.php
```diff
@@ -1123,41 +1150,56 @@ function sodium_crypto_secretstream_xchacha20poly1305_init_push(string $key): ar
...
+ * @param string $key
+ * @return string
  * @throws SodiumException
  * @since 7.2
  * @see https://www.php.net/manual/en/function.sodium-crypto-secretstream-xchacha20poly1305-init-pull.php
  */
 function sodium_crypto_secretstream_xchacha20poly1305_init_pull(string $header, string $key): string{}
 
 /**
+ * @param string &$state ,
```
isfedorov, line 1171: Extra comma

sodium/sodium.php
```diff
@@ -1080,18 +1081,38 @@ function sodium_crypto_aead_chacha20poly1305_keygen(): string {}
...
+ * @param string $nonce
+ * @param string $key
+ * @return string
  * @throws SodiumException
  * @since 7.2
  * https://www.php.net/manual/en/function.sodium-crypto-aead-xchacha20poly1305-ietf-encrypt.php
  */
-function sodium_crypto_aead_xchacha20poly1305_ietf_encrypt(string $message, string $additional_data, string $nonce, string $key): string{}
+function sodium_crypto_aead_xchacha20poly1305_ietf_encrypt(
+    string $message,
```
isfedorov, line 1111: Not sure that such formatting changes required. Seems there are no too many benefits with them

sodium/sodium.php
```diff
@@ -1080,18 +1081,38 @@ function sodium_crypto_aead_chacha20poly1305_keygen(): string {}
...
-function sodium_crypto_aead_xchacha20poly1305_ietf_decrypt(string $ciphertext, string $additional_data, string $nonce, string $key): string|false {}
+function sodium_crypto_aead_xchacha20poly1305_ietf_decrypt(
+    string $ciphertext,
+    string $additional_data,
+    string $nonce,
+    string $key
+): string|false {}
 
 /**
+ * @param string$message
```
isfedorov, line 1101: space missed between type and param name

redis/RedisCluster.php
```diff
@@ -2506,7 +2506,7 @@ public function zUnionStore($Output, $ZSetKeys, array $Weights = null, $aggregat
      * $redisCluster->zInterStore('ko4', array('k1', 'k2'), array(1, 5), 'max'); // 2, 'ko4' => array('val3', 'val1')
      * </pre>
      */
-    public function zInterStore($Output, $ZSetKeys, array $Weights = null, $aggregateFunction = 'SUM') { }
+    public function zInterStore($Output, $ZSetKeys, ?array $Weights = null, $aggregateFunction = 'SUM') { }
```
isfedorov, line 2509: As far as parent class isn't final I suggest leaving parameter type hint non-nullable to avoid extra warning with PHP < 7.1 in child classes

http/http.php
```diff
@@ -2615,15 +2614,15 @@ function http_get ($url, array $options = null ,  array &$info = null ) {}
  * @param string $url [optional] <p>
  * URL
  * </p>
- * @param array $options [optional] <p>
- * &link.http.request.options;
+ * @param null|array $options [optional] <p>
+():
```
isfedorov, line 2618: Probably this (): is extra symbols

http/http.php
```diff
@@ -2187,22 +2187,22 @@ function http_date ($timestamp = null) {}
  * (PECL pecl_http &gt;= 0.21.0)<br/>
  * Build an URL
  * @link https://php.net/manual/en/function.http-build-url.php
- * @param mixed $url [optional] <p>
+ * @param null|mixed $url [optional] <p>
  * (part(s) of) an URL in form of a string or associative array like parse_url returns
  * </p>
- * @param mixed $parts [optional] <p>
+ * @param null|mixed $parts [optional] <p>
```
isfedorov, line 2193: This actually doesn't make sense as mixed includes `null`

http/http.php
```diff
@@ -2735,21 +2728,20 @@ function http_put_stream ($url, $stream = null, array $options = null ,  array &
...
+ * @param null|string $body [optional] <p>
  * Request body
  * </p>
- * @param array $options [optional] <p>
- * &link.http.request.options;
+ * @param null|array $options [optional] <p>
  * </p>
- * @param array &$info [optional] <p>
+ * @param null|array &$info [optional] <p>
  * &link.http.request.info;
```
isfedorov, line 2740: Broken link probably can be also removed

http/http.php
```diff
@@ -2715,18 +2709,17 @@ function http_put_file ($url, $file = null, array $options = null ,  array &$inf
...
+ * @param null|resource $stream [optional] <p>
  * The stream to read the PUT request body from
  * </p>
- * @param array $options [optional] <p>
- * &link.http.request.options;
+ * @param null|array $options [optional] <p>
  * </p>
- * @param array &$info [optional] <p>
+ * @param null|array &$info [optional] <p>
  * &link.http.request.info;
```
isfedorov, line 2718: This broken link can be removed

http/http.php
```diff
@@ -2695,18 +2690,17 @@ function http_put_data ($url, $data = null, array $options = null ,  array &$inf
...
+ * @param null|string $file [optional] <p>
  * The file to put
  * </p>
- * @param array $options [optional] <p>
- * &link.http.request.options;
+ * @param null|array $options [optional] <p>
  * </p>
- * @param array &$info [optional] <p>
+ * @param null|array &$info [optional] <p>
  * &link.http.request.info;
```
isfedorov, line 2699: One more broken link that can be removed

dom/dom.php
```diff
@@ -11,7 +11,7 @@
  * </p>
  * @return DOMElement|null|false The <b>DOMElement</b> node added or <b>FALSE</b> if any errors occur.
  */
-function dom_import_simplexml (object $node): ?DOMElement {}
+function dom_import_simplexml ($node) {}
```
isfedorov, line 14: Could you please clarify why `object` type hint should be removed here?

bcmath/bcmath.php
```diff
@@ -150,15 +151,34 @@ function bcpow (string $num, string $exponent, ?int $scale = 0): string
...
  * Set default scale parameter for all bc math functions
  * @link https://php.net/manual/en/function.bcscale.php
- * @param int|null $scale <p>
- * The scale factor. Since 7.3.0 can be omitted.
- * </p>
- * @return int|true <b>INT</b> since 7.3.0 and <b>TRUE</b> before.
+ * @param int $scale
+ * @return true
+ */
+function bcscale (int $scale): bool
```
isfedorov, line 161: Let's change duplicated functions to `LanguageLevelAware` attribute and leave just one function

isfedorov:
@voku Could you please review your changes in methods and check that nullable type hints are added only in final classes and final methods otherwise if someone with PHP < 7.1 will override those methods with nullable type hints, one will get extra warnings that "Declaration should be compatible" but will not be able to fix this warning with that language level. Also seems there are some conflicting files so you will have to rebase and commit the rebased version one more time. Thank you for your contribution!

dom/dom.php
voku: PhpStorm itself reported "Argument type does not match the declared" because the phpdoc is "SimpleXMLElement" != "object", but I will revert the change, because I am not sure what is correct here.

bcmath/bcmath.php
voku: I don't know how to handle new default value NULL via ```LanguageLevelAware``` :-/

sodium/sodium.php
voku: These formatting changes have been rolled back.

voku:
> @voku Could you please review your changes in methods and check that nullable type hints are added only in final classes and final methods otherwise if someone with PHP < 7.1 will override those methods with nullable type hints, one will get extra warnings that "Declaration should be compatible" but will not be able to fix this warning with that language level. Also seems there are some conflicting files so you will have to rebase and commit the rebased version one more time (You can try to run tests locally before push to avoid extra commits). Thank you for your contribution!

I fixed the nullable stuff, but I think I still see a false positive test result for "array" vs "string[]"

isfedorov:
@voku I think that problem is that you are declaring parameter type via `LanguageLevelTypeAware` but this parameter doesn't have such type based on reflection.

voku:
> @voku I think that problem is that you are declaring parameter type via `LanguageLevelTypeAware` but this parameter doesn't have such type based on reflection.

Ok, I added extra stubs for "to: 7.4" and "8.0". :+1:

.idea/inspectionProfiles/profiles_settings.xml
```diff
@@ -1,7 +1,7 @@
 <component name="InspectionProjectProfileManager">
   <settings>
-    <option name="PROJECT_PROFILE" value="Project Default" />
-    <option name="USE_PROJECT_PROFILE" value="true" />
+    <option name="PROJECT_PROFILE" value="vdmg" />
```
isfedorov, line 3: Please rollback all the changes of .idea folder

voku: :-/ done: ```git checkout upstream/master -- .idea/```                                                                                                                                        

I need to add it into the global .gitignore

isfedorov:
@voku could you please properly rebase your branch on upstream/master and resolve all conflicts? Currently, PR cannot be merged as GitHub says `This branch cannot be rebased due to conflicts
Rebasing the commits of this branch on top of the base branch cannot be performed automatically due to conflicts encountered while reapplying the individual commits from the head branch. `

voku:
> @voku could you please properly rebase your branch on upstream/master and resolve all conflicts? Currently, PR cannot be merged as GitHub says `This branch cannot be rebased due to conflicts Rebasing the commits of this branch on top of the base branch cannot be performed automatically due to conflicts encountered while reapplying the individual commits from the head branch. ` Also please squash all not meaningful commits (like ... v2, merge commits, duplicated commits) so the history would be clean

@isfedorov I have clean-up the git history and rebased with upstream.

bcmath/bcmath.php
isfedorov: That's right, unfortunately, it's not possible to handle default value via that attribute, but at least return types in two functions `function bcscale (int $scale): bool` and `function bcscale (int $scale): int` could be handled and leave only function `function bcscale (?int $scale = null): int` marked with `#[PhpStormStubsElementAvailable('8.0')]`

standard/standard_9.php
```diff
@@ -786,6 +789,22 @@ function array_combine(array $keys, array $values): array { }
  * @return bool true on success or false on failure.
  */
 #[Pure]
+function array_key_exists($key, $array): bool
+{ }
+
+#[PhpStormStubsElementAvailable('8.0')]
```
isfedorov, line 795: I think `#[LanguageLevelTypeAware]` can be used here for the type of $array as well.

voku: done

openssl/openssl.php
```diff
@@ -542,18 +544,114 @@ function openssl_pkcs12_read(string $pkcs12, &$certificates, string $passphrase)
...
+ * @param array|null $extra_attributes [optional] <p>
+ * <i>extraattribs</i> is used to specify additional
+ * configuration options for the CSR. Both <i>dn</i> and
+ * <i>extraattribs</i> are associative arrays whose keys are
+ * converted to OIDs and applied to the relevant part of the request.
+ * </p>
+ * @return OpenSSLCertificateSigningRequest|false the CSR.
+ */
+function openssl_csr_new(array $distinguished_names, &$private_key, ?array $options, ?array $extra_attributes): OpenSSLCertificateSigningRequest|false
+{ }
```
isfedorov, line 647: LanguageLevelTypeAware also can be used here for return type as this function is completely equal to above one except return type. Do I miss something?

voku: Yep, fixed. :+1:

PR #963, opened 2020-11-23
Fix DateTime/DateTimeImmutable::getLastErrors() return type

b1rdex:


b1rdex:
This is wrong

b1rdex:
Actually it's possible: https://3v4l.org/HBq0q

isfedorov:
Hi @b1rdex ,
Thank you for the PR and for the example from 3v4l

PR #969, opened 2020-11-26
iterator_to_array is not pure

ondrejmirtes:
See example: https://github.com/phpstan/phpstan/issues/4126

### DIFF
--- a/SPL/SPL_f.php
+++ b/SPL/SPL_f.php
@@ -155,7 +155,6 @@ function spl_object_hash (object $object): string
  * </p>
  * @return array An array containing the elements of the iterator.
  */
-#[Pure]
 function iterator_to_array (Traversable $iterator, bool $preserve_keys = true): array
 {}
 


isfedorov:
@ondrejmirtes sorry, but could you please clarify why the function isn't pure? What is a side-effect of this function? From your example `https://phpstan.org/r/32b5f205-57f0-475c-8bfb-47feca2ed3aa` I see the error message `Call to function iterator_to_array() on a separate line has no effect.` that means (if I understand phpstan properly) that result of the function call isn't used and if I change that code to `$a = iterator_to_array($i);` then I see "No errors" message

ondrejmirtes:
@isfedorov The problem is that we have no control over what happens in the iterator. The iterator might throw an exception. Or it might print something which is what the linked example is about:

```php
function i(array $a): Traversable
{
    foreach ($a as $e) {
        echo $e, PHP_EOL;
        yield $e;
    } 
}
```

So `iterator_to_array` doesn't have a guarantee of being pure.

isfedorov:
@ondrejmirtes true! Thank you for your contribution!

PR #983, opened 2020-12-08
[Sodium] Update parameter name

Nyholm:
`sodium_crypto_box_seal` takes a public key. 

```php
<?php

$keypair = sodium_crypto_box_keypair();
$publicKey = sodium_crypto_box_publickey($keypair);
$privateKey = sodium_crypto_box_secretkey($keypair);

$message = 'input';
$ciphertext = sodium_crypto_box_seal($message, $publicKey);
var_dump($publicKey);

$output = sodium_crypto_box_seal_open($ciphertext, $keypair);
var_dump($output);
````

### DIFF
--- a/sodium/sodium.php
+++ b/sodium/sodium.php
@@ -456,7 +456,7 @@ function sodium_crypto_box_publickey_from_secretkey(string $secret_key): string 
  * @throws SodiumException
  * @since 7.2
  */
-function sodium_crypto_box_seal(string $message, string $key_pair): string {}
+function sodium_crypto_box_seal(string $message, string $key): string {}
 
 /**
  * Anonymous public-key encryption (decrypt)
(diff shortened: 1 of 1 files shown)


isfedorov:
@Nyholm the problem here is that parameter name is `$key_pair` according to reflection. Try to invoke `var_dump((new ReflectionFunction('sodium_crypto_box_seal'))->getParameters());`
Screenshot 2020-12-17 at 18 33 47
and we check parameters names according to reflection that's why test fails.
Also stubs from php-src https://github.com/php/php-src/blob/PHP-8.0/ext/sodium/libsodium.stub.php#L55

Nyholm:
Interesting. 
you are correct closing this pr, but my pr is valid :)

I’ll see if I’ll update the php source. 
thank you.

PR #991, opened 2020-12-15
More type fixes

voku:
found via https://github.com/voku/php-doc-fixer

---
This change is [Reviewable](https://reviewable.io/reviews/jetbrains/phpstorm-stubs/991)

### DIFF
--- a/igbinary/igbinary.php
+++ b/igbinary/igbinary.php
@@ -18,7 +18,7 @@
  * Likewise, when the object is restored using unserialize() the __wakeup member function is called.
  *
  * @param mixed $value The value to be serialized.
- * @return string Returns a string containing a byte-stream representation of value that can be stored anywhere.
+ * @return string|null Returns a string containing a byte-stream representation of value that can be stored anywhere or <b>NULL</b> on error.
  * @link https://secure.php.net/serialize PHP default serialize
  */
 function igbinary_serialize($value) {}
@@ -31,7 +31,7 @@ function igbinary_serialize($value) {}
  * In case the passed string is not unserializeable, NULL is returned and E_WARNING is issued.
  *
  * @param string $str The serialized string.
- * @return mixed The converted value is returned, and can be a boolean, integer, float, string, array or object.
+ * @return mixed|false The converted value is returned, and can be a boolean, integer, float, string, array, object or <b>false</b> by empty string input.
  * @link https://secure.php.net/manual/en/function.unserialize.php PHP default unserialize
  * @link https://secure.php.net/~helly/php/ext/spl/interfaceSerializable.html Serializable
  */
--- a/standard/standard_7.php
+++ b/standard/standard_7.php
@@ -221,7 +221,7 @@ function unpack (string $format, string $string, int $offset = 0): array|false
 function get_browser (?string $user_agent, bool $return_array): object|array|false
 {}
 
-
+#[PhpStormStubsElementAvailable(to: '7.4')]
 /**
  * One-way string encryption (hashing)
  * @link https://php.net/manual/en/function.crypt.php
@@ -241,7 +241,6 @@ function get_browser (?string $user_agent, bool $return_array): object|array|fal
  * @return string|null the encrypted string or <b>NULL</b> if an error occurs
  */
 #[Pure]
-#[PhpStormStubsElementAvailable(to: '7.4')]
 function crypt ($string, $salt = null): ?string
 {}
 
--- a/standard/standard_7.php
+++ b/standard/standard_7.php
@@ -221,7 +221,6 @@ function unpack (string $format, string $string, int $offset = 0): array|false
 function get_browser (?string $user_agent, bool $return_array): object|array|false
 {}
 
-#[PhpStormStubsElementAvailable(to: '7.4')]
 /**
  * One-way string encryption (hashing)
  * @link https://php.net/manual/en/function.crypt.php
@@ -241,6 +240,7 @@ function get_browser (?string $user_agent, bool $return_array): object|array|fal
  * @return string|null the encrypted string or <b>NULL</b> if an error occurs
  */
 #[Pure]
+#[PhpStormStubsElementAvailable(to: '7.4')]
 function crypt ($string, $salt = null): ?string
 {}
 
--- a/xlswriter/xlswriter.php
+++ b/xlswriter/xlswriter.php
@@ -498,7 +498,7 @@ namespace Vtiful\Kernel {
         /**
          * Wrap
          *
-         * @return Format
+         * @return $this
          */
         public function wrap(): self
         {
@@ -508,7 +508,7 @@ namespace Vtiful\Kernel {
         /**
          * Bold
          *
-         * @return Format
+         * @return $this
          */
         public function bold(): self
         {
@@ -518,7 +518,7 @@ namespace Vtiful\Kernel {
         /**
          * Italic
          *
-         * @return Format
+         * @return $this
          */
         public function italic(): self
         {
@@ -530,7 +530,7 @@ namespace Vtiful\Kernel {
          *
          * @param int $style const BORDER_***
          *
-         * @return Format
+         * @return $this
          */
         public function border(int $style): self
         {
@@ -542,7 +542,7 @@ namespace Vtiful\Kernel {
          *
          * @param int ...$style const FORMAT_ALIGN_****
          *
-         * @return Format
+         * @return $this
          */
         public function align(...$style): self
         {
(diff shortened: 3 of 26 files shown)


ds/ds.php
```diff
@@ -35,7 +35,7 @@ public function clear(): void;
         /**
          * Returns a shallow copy of the collection.
          * @link https://www.php.net/manual/en/ds-collection.copy.php
-         * @return Collection
+         * @return \Ds\Collection
```
isfedorov, line 38: What is the purpose of such changes? type is already being properly inferred as `\Ds\Collection` 
Screenshot 2020-12-16 at 01 32 56

voku: Yes via PhpStorm, but it's easier for everyone if there are unambiguous phpDocs, and so a simple script can check the phpDocs, without the need to resolve the namespaces.

MaXal: @voku I agree with @isfedorov that the change is unneeded and should be reverted. It makes PHPDoc more verbose and any script will have to resolve namespaces to be correct anyway since it could be easily be broken by some other change. 
Let's not simplify the creation of some abstract scripts by complicating existing stubs and forcing addition rules (only FQN in PHPDoc).

MaXal (changes requested):
Please revert changes to FQN in PHPDoc.

voku:
> Please revert changes to FQN in PHPDoc.

done

voku:
@isfedorov the tests fail because of ```ob_implicit_flush(int $enable = 1 | bool $enable = true)``` but there are different default values, do I need to ignore the problem or how should I handle that?

isfedorov:
> @isfedorov the tests fail because of `ob_implicit_flush(int $enable = 1 | bool $enable = true)` but there are different default values, do I need to ignore the problem or how should I handle that?

@voku I think it's better to use `#[LanguageLevelTypeAware] to set different parameter types and remove default value completely. As far as parameter has `[optional]` in phpdoc there will be no warning in PhpStorm that parameter is missed. If you still want to add some default value you can set 1 or true,  in any case on the call of `ob_implicit_flush()` default value from extension will be used in runtime and default value from stubs actually will not be used.

isfedorov:
@voku please also don't forget to add classes `PhpStormStubsElementAvailable` or `LanguageLevelTypeAware` to imports where required

igbinary/igbinary.php
```diff
@@ -31,7 +31,7 @@ function igbinary_serialize($value) {}
  * In case the passed string is not unserializeable, NULL is returned and E_WARNING is issued.
  *
  * @param string $str The serialized string.
- * @return mixed The converted value is returned, and can be a boolean, integer, float, string, array or object.
+ * @return mixed|false The converted value is returned, and can be a boolean, integer, float, string, array, object or <b>false</b> by empty string input.
```
isfedorov, line 34: `Mixed` already includes `false` so `false` here can be removed from typehint

standard/standard_7.php
```diff
@@ -221,7 +221,7 @@ function unpack (string $format, string $string, int $offset = 0): array|false
 function get_browser (?string $user_agent, bool $return_array): object|array|false
 {}
 
-
+#[PhpStormStubsElementAvailable(to: '7.4')]
```
isfedorov, line 224: Could you please rollback this change as it's code style related? Overall in stubs attributes are placed between phpdoc and signature

xlswriter/xlswriter.php
```diff
@@ -498,7 +498,7 @@ public function __construct($fileHandle)
         /**
          * Wrap
          *
-         * @return Format
+         * @return $this
```
isfedorov, line 501: Please rollback these changes as they don't make any sense

xmlwriter/xmlwriter.php
```diff
@@ -519,7 +519,7 @@ public function flush ($empty = true) {}
  * @param string $uri <p>
  * The URI of the resource for the output.
  * </p>
- * @return bool|resource Object oriented style: Returns <b>TRUE</b> on success or <b>FALSE</b> on failure.
+ * @return false|resource|XMLWriter Object oriented style: Returns <b>TRUE</b> on success or <b>FALSE</b> on failure.
```
isfedorov, line 522: https://github.com/php/php-src/blob/master/ext/xmlwriter/php_xmlwriter.c#L924 seems it can return true as well so reducing type from `bool` to `false` seems to be incorrect

yar/yar.php
```diff
@@ -98,7 +98,7 @@ public function __call ( $method , $parameters ){}
      * - YAR_OPT_CONNECT_TIMEOUT
      * - YAR_OPT_HEADER (Since 2.0.4)
      * @param $value
-     * @return object|false Returns $this on success or FALSE on failure.
+     * @return $this|false Returns $this on success or FALSE on failure.
```
isfedorov, line 101: Please change $this to class name

igbinary/igbinary.php
voku: We already have many stubs that use "mixed|false" and I think it can help some developers to see that there is a kind of error handling that are needed to check, if you will properly use this. But if you think that it's not helpful, should we revert all "x|mixed|x" PHPDocs or only this?

standard/standard_7.php
voku: Currently, I see the "PhpStormStubsElementAvailable" attributes only over the PHPDoc, and this was the only place where we use it between PHPDoc and signature. Should we move all in between or only this?

xlswriter/xlswriter.php
voku: Please take a look at this working example: 

```php
<?php

use Vtiful\Kernel\Excel;
use Vtiful\Kernel\Format;

class Format2 extends Format 
{
        /**
         * @return $this
         */
        public function lall(): self
        {
	    // ...

            return $this;
        }
}

$excel = new \Vtiful\Kernel\Excel([
    'path' => './'
]);

$excelObject = $excel->fileName('tutorial.xlsx');
$excelHandle = $excelObject->getHandle();

// new format object
$format = new Format2($excelHandle);

// set format
$formatResource = $format->lall()->bold();

var_dump($formatResource); // output: object(Format2)#2 (0) { }
```

xmlwriter/xmlwriter.php
voku: The ```if (self) {``` check will prevent this behavior for the procedural style. 

⇾ https://www.php.net/manual/en/function.xmlwriter-open-uri.php

yar/yar.php
voku: But ```$this``` or ```static``` is more correct here, or? 

```php
<?php

/* // server.php
<?php

class API {
    public function test(int $i) {
        return $i + 1;
    }
}

$service = new Yar_Server(new API());
$service->handle();
*/

class Yar_Client2 extends Yar_Client
{
    /**
     * @return $this
     */
    public function lall(): self {
        return $this;
    }
}

$client = new Yar_Client2("http://localhost/server.php");

echo $client->setOpt(YAR_OPT_CONNECT_TIMEOUT, 10)->lall()->test(99); // output: 100
```

xlswriter/xlswriter.php
isfedorov: So `static` return type should be used in this case, isn't it? My point is that `$this` isn't a type but `@return` should have a type declaration. Yes, PhpStorm can resolve type based on `$this` but it seems to me that it's better to follow PhpDoc convention.

igbinary/igbinary.php
isfedorov: Fair enough. Let's leave it.

standard/standard_7.php
isfedorov: Thank you for the note! Well, this is rather the issue with other `PhpStormStubsElementAvailable` and if look at other attributes they are placed between phpdoc and signature. So yes, I think it makes sense to move all attributes between phpdoc and signature. Despite all those changes would be style-related but taking into account possible issues like https://youtrack.jetbrains.com/issue/WI-55651 I would move attributes just in case.

yar/yar.php
isfedorov: The same answer as above actually. If a class can be extended it makes sense to use `static` as a return type as `$this` isn't actually a type

voku: There are already many stubs with "$this" in phpdocs, should we convert all to "static"? Should I create a separate PR for that?

isfedorov: Yes, it makes sense and yes, please create a separate PR for this. Thank you for your contribution!

standard/standard_7.php
voku: I will create a new PR for this code style fixes. :+1:

PR #1001, opened 2020-12-19
Fix DateTimeZone signatures

BenMorel:
Sources:

- https://www.php.net/manual/en/datetimezone.getlocation.php
- https://www.php.net/manual/en/datetimezone.getoffset.php
- https://www.php.net/manual/en/datetimezone.gettransitions.php
- https://www.php.net/manual/en/datetimezone.listabbreviations.php
- https://www.php.net/manual/en/datetimezone.listidentifiers.php

### DIFF
--- a/date/date_c.php
+++ b/date/date_c.php
@@ -8,62 +8,62 @@ interface DateTimeInterface {
      * @since 7.2
      */
     const ATOM = 'Y-m-d\TH:i:sP';
-    
+
     /**
      * @since 7.2
      */
     const COOKIE = 'l, d-M-Y H:i:s T';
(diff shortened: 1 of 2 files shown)


date/date_c.php
```diff
@@ -8,62 +8,62 @@ interface DateTimeInterface {
      * @since 7.2
      */
     const ATOM = 'Y-m-d\TH:i:sP';
-    
+
```
BenMorel, line 11: Oops, PHPStorm fixed some empty lines when saving the file :)
Leaving as is though, I think this deserves to be fixed, too.

PR #1002, opened 2020-12-20
Improve assertions

peter279k:
# Changed log

- Using the `assertCount` to assert expected is same as result count.
- Using the `assertStringNotContainsString` to assert expected string is not contains specific string.


isfedorov:
@peter279k Thank you for the contribution! 👍

PR #1009, opened 2021-01-08
Proposal: Throws attribute

navarr:


wbars:
Hello. While your suggestion surely make sense, I don't see what additional opportunities migrating this to attributes will open to users. The only argument of the attribute is exception FQN (string) and there is no additional profit in specifying this FQN in attribute argument rather than phpdoc tag value.

I will decline this proposal for now, but these may be reconsidered in future. Thank you for the suggestion.

PR #1015, opened 2021-01-12
minor cleanup

isfedorov:


qodana.yaml
```diff
@@ -0,0 +1,181 @@
+version: 1.0
+failThreshold: 0
+profile:
+  name: qodana.recommended
+exclude:
+  - name: All
+    paths:
```
MaXal, line 7: Maybe we should keep the check for stubs as well, just use a different profile when it will be possible.

MaXal (changes requested):

qodana.yaml
isfedorov: Yes, we will configure it later. It is just initial version to check if config will work at all :) I think we will just disable some inspections for stubs folders and leave all inspections for tests folder

PR #1019, opened 2021-01-18
Typehint mismatch in signature and phpdoc

isfedorov:
Typehints are not being stored in arrays instead of a single string. Related tests are updated to use arrays.

### DIFF
--- a/tests/Model/BasePHPElement.php
+++ b/tests/Model/BasePHPElement.php
@@ -38,11 +38,17 @@ abstract class BasePHPElement
     abstract public function readMutedProblems(stdClass|array $jsonData): void;
 
     #[Pure]
-    protected function getFQN(Node $node): string
+    public static function getFQN(Node $node): string
     {
         $fqn = '';
-        if ($node->namespacedName === null) {
-            $fqn = $node->name->parts[0];
+        if (!property_exists($node, 'namespacedName') || $node->namespacedName === null) {
+            if (property_exists($node, 'name')) {
+                $fqn = $node->name->parts[0];
+            } else {
+                foreach ($node->parts as $part) {
+                    $fqn .= "$part\\";
+                }
+            }
         } else {
             /**@var string $part */
             foreach ($node->namespacedName->parts as $part) {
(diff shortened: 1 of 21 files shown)


tests/Model/BasePHPElement.php
```diff
@@ -52,35 +58,34 @@ protected function getFQN(Node $node): string
         return rtrim($fqn, "\\");
     }
 
-    protected static function convertReflectionTypeToString(?ReflectionType $type): string
+    protected static function getReflectionTypeAsArray(?ReflectionType $type): array
     {
+        $reflectionTypes = [];
         if ($type instanceof ReflectionNamedType) {
-            return (string)$type;
+            $type->allowsNull() && $type->getName() !== 'mixed' ?
```
MaXal, line 65: Doesn't this work `array_push($reflectionTypes, (string) $type)`?

tests/Model/BasePHPElement.php
```diff
@@ -52,35 +58,34 @@ protected function getFQN(Node $node): string
...
+                array_push($reflectionTypes, '?' . $type->getName()) : array_push($reflectionTypes, $type->getName());
         }
         if ($type instanceof ReflectionUnionType) {
-            $reflectionType = '';
             foreach ($type->getTypes() as $type) {
-                $reflectionType .= (string)$type . '|';
+                array_push($reflectionTypes, $type->getName());
```
MaXal, line 70: Why not `array_push($reflectionTypes, (string) $type)`?

tests/Model/PHPMethod.php
```diff
@@ -44,10 +44,16 @@ public function readObjectFromReflection($reflectionObject): static
...
     public function readObjectFromStubNode($node): static
     {
-        $this->parentName = $this->getFQN($node->getAttribute('parent'));
+        $this->parentName = self::getFQN($node->getAttribute('parent'));
         $this->name = $node->name->name;
+        $typeFromAttribute = self::findTypeFromAttribute($node->attrGroups);
         $this->availableVersionsRangeFromAttribute = self::findAvailableVersionsRangeFromAttribute($node->attrGroups);
-        $this->returnType = self::convertParsedTypeToString($node->getReturnType());
+        if ($typeFromAttribute != null) {
+            array_push($this->returnTypesFromSignature, ...array_map(fn(string $type) => preg_replace('/\w+\[]/', 'array', $type),
```
MaXal, line 52: This code is duplicated a couple of times. Should we change `findTypeFromAttribute` to return array as well?

tests/Model/PHPFunction.php
```diff
@@ -24,9 +24,9 @@ class PHPFunction extends BasePHPElement
      */
     public array $parameters = [];
 
-    public ?Type $returnTag = null;
+    public array $returnTypesFromPhpDoc = [];
```
MaXal, line 27: Let's add `string[]` as a field type.

tests/Model/PHPParameter.php
```diff
@@ -10,7 +10,7 @@
 
 class PHPParameter extends BasePHPElement
 {
-    public string $type = '';
+    public array $types = [];
```
MaXal, line 13: Let's add type here (`string[]`).

tests/Model/BasePHPElement.php
isfedorov: Expected this question :) `__toString` is deprecated 
```
#[Deprecated(replacement: "%class$->getName()", since: "7.1")]
    public function __toString()
    {
    }
```
that's why decided to get rid of it now, to avoid future problems

tests/Model/BasePHPElement.php
isfedorov: The same as above

tests/Model/PHPMethod.php
isfedorov: Yes, thought about it. You confirmed my doubts. Will update this. Thanks!

tests/Model/PHPFunction.php
isfedorov: Will do!

tests/Model/PHPParameter.php
isfedorov: Yep!

PR #1028, opened 2021-01-26
Add thrown exceptions in imagick

dryabov:
It's made by semi-automatical analysis of the imagick's source code.
Unfortunately, there is no official docs on this topic, and this patch is better than nothing.


dryabov:
I see the tests are failed, but I don't know the reason. Is there a way to get some details in addition to "Process completed with exit code 1"?

isfedorov:
@dryabov I think the problem is that you manually added empty line at the end of `PhpStormStubsMap.php`. Please just invoke `docker-compose -f docker-compose.yml run php /usr/local/bin/php /opt/project/phpstorm-stubs/generate-stub-map` locally and just commit `PhpStormStubsMap.php` without any other changes in that file

dryabov:
@isfedorov Yes. The problem was that I worked on PC w/o my GitHub's private key and has to send patches via its web interface that modifies file in such a way.

PR #1029, opened 2021-01-27
add test for optional parameters

isfedorov:
Optional parameters in stubs should be defined by default value or by `[optional]` note in related phpdoc. Vararg parameters are parsed as optional as well

### DIFF
--- a/standard/standard_1.php
+++ b/standard/standard_1.php
@@ -1069,7 +1069,7 @@ function implode (array|string $separator = "", ?array $array): string
  * elements in the same order, with the glue string between each element.
  */
 #[Pure]
-function join (array|string  $separator = "", ?array $array = null): string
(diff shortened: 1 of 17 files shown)


standard/standard_1.php
```diff
@@ -1069,7 +1069,7 @@ function implode (array|string $separator = "", ?array $array): string
  * elements in the same order, with the glue string between each element.
  */
 #[Pure]
-function join (array|string  $separator = "", ?array $array): string
+function join (array|string  $separator = "", ?array $array = null): string
```
MaXal, line 1072: This is not correct: https://3v4l.org/n3SIL

MaXal (changes requested):

standard/standard_1.php
isfedorov: Reverted change and added to ignore

PR #1031, opened 2021-01-30
Add stubs for php-ast

TysonAndre:
These are the stubs for https://github.com/nikic/php-ast using documentation I wrote for https://github.com/phan/phan 's use of php-ast (and function descriptions from php-ast's ast.stub.php)

Having descriptions for the constants including what to expect in $node->children is useful when hovering over code using those constants.

The build failure seems unrelated (pecl was down at the time of the build?)


isfedorov:
@TysonAndre could you please change return type in signature in functions `ast\parse_code` and `ast\parse_file` from `ast\Node` to `Node`? There is a separate test that compares return types in signature and phpdoc and it fails now (unfortunately it's not so smart yet to compare fqn and non fqn) and actually FQN is redundant in these cases.

isfedorov:
@TysonAndre that's strange, I see your commit and changes but PR still uses old version of code. I've tried to download your patch and apply it locally but got next result
Screenshot 2021-02-01 at 16 45 12
So seems something went wrong with the changes. Could you please try to rebase and force push your changes one more time? Also please don't forget to invoke `docker-compose -f docker-compose.yml run php /usr/local/bin/php /opt/project/phpstorm-stubs/generate-stub-map` and commit the resulting `PhpStormStubsMap.php`

PR #1039, opened 2021-02-12
Use a modern return type for ReflectionFunctionAbstract::getReturnType()

lptn:
ReflectionFunctionAbstract::getReturnType() added to PHP7.0

As for PHP7.0, it returns ReflectionType
As for PHP7.1+, it returns ReflectionNamedType

Technically, existing stub is valid, because class ReflectionNamedType extends ReflectionType, but when you use `ReflectionFunctionAbstract::getReturnType()->getName()` it’s wrong on PHP7.0 and leads to Fatal error: `Uncaught Error: Call to undefined method ReflectionType::getName()`.

As alternative, we can use `@return ReflectionType|ReflectionNamedType|null`

@see https://3v4l.org/Q8aDT

related PR on psalm repo: https://github.com/vimeo/psalm/pull/5204

### DIFF
--- a/Reflection/ReflectionProperty.php
+++ b/Reflection/ReflectionProperty.php
@@ -236,7 +236,7 @@ class ReflectionProperty implements Reflector
      * Gets property type
      *
      * @link https://php.net/manual/en/reflectionproperty.gettype.php
-     * @return ReflectionType|null Returns a {@see ReflectionType} if the
+     * @return ReflectionNamedType|null Returns a {@see ReflectionNamedType} if the
      * property has a type, and {@see null} otherwise.
      * @since 7.4
      */
--- a/Reflection/ReflectionProperty.php
+++ b/Reflection/ReflectionProperty.php
@@ -236,11 +236,12 @@ class ReflectionProperty implements Reflector
      * Gets property type
      *
      * @link https://php.net/manual/en/reflectionproperty.gettype.php
-     * @return ReflectionNamedType|null Returns a {@see ReflectionNamedType} if the
+     * @return ReflectionNamedType|ReflectionUnionType|null Returns a {@see ReflectionType} if the
      * property has a type, and {@see null} otherwise.
      * @since 7.4
      */
     #[Pure]
+    #[LanguageLevelTypeAware(['8.0' => 'ReflectionNamedType|ReflectionUnionType|null'], default: 'ReflectionNamedType|null')]
 	public function getType()
     {
     }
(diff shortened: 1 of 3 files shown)


isfedorov:
@lptn I suggest to use `LanguageLevelTypeAware` attribute for this function. Using this attribute PhpStorm will use `ReflectionType` or `ReflectionNamedType` depending on language level set for a project. See `spl_autoload_functions` as example of attribute usage

Reflection/ReflectionProperty.php
```diff
@@ -236,7 +236,7 @@ public function setAccessible($accessible)
      * Gets property type
      *
      * @link https://php.net/manual/en/reflectionproperty.gettype.php
-     * @return ReflectionType|null Returns a {@see ReflectionType} if the
+     * @return ReflectionNamedType|null Returns a {@see ReflectionNamedType} if the
```
lptn, line 239: in real life, it always returns ReflectionNamedType: https://3v4l.org/Y2Jqo

lptn:
@isfedorov
thanks a lot, done ;)

Reflection/ReflectionProperty.php
isfedorov: While it is safe to say that it will always be a ReflectionNamedType in PHP 7.4, it can also be a ReflectionUnionType in PHP 8 if union types are used. Please see https://3v4l.org/3X0Rp.

So the correct stub type is the common parent, ReflectionType or LanguageLevelTypeAware should be used again. Please update this change

lptn: @isfedorov 
great note, thanks a lot! Fixed by https://github.com/JetBrains/phpstorm-stubs/pull/1039/commits/4807a64f80bdb9b0539dc55e9c74702aa1225682

It's still not clear to me on should I update PHPDoc blocks for these methods: they uses `@return ReflectionType|null Returns a {@see ReflectionType}`. But since we can't differentiate PHP dock blocks for different PHP versions, it’s not 100% valid to use union types like `@return ReflectionType|ReflectionNamedType|ReflectionUnionType|null`. Alternatively, we can re-declare these methods and use `PhpStormStubsElementAvailable` attribute (as we do, for example, with `bcscale` function). What do you think about it?

isfedorov: Yes, I think `PhpStormStubsElementAvailable` can be used for this.

isfedorov:
@lptn PR looks good but unfortunately, it can't be merged until WI-58641 is fixed. As soon as WI-58641 will be fixed we will merge the PR

isfedorov:
Fixed in #1095

PR #1051, opened 2021-02-19
Fix var name in sodium.php

tolbon:
Hello,
I fix var name in body function with parameter name

but i think it's useless cause lot of function here doesn't have body.
Do you want to delete all body for this file ?


isfedorov:
@tolbon you are absolutely right! Function bodies are useless as these functions are just stubs, so please remove the bodies.

PR #1052, opened 2021-02-19
Soap stubs PHP 7 versions

sakarikl:
Soap stubs PHP 7 versions also included

### DIFF
--- a/soap/soap.php
+++ b/soap/soap.php
@@ -2,6 +2,7 @@
 
 // Start of soap v.
 use JetBrains\PhpStorm\Deprecated;
+use JetBrains\PhpStorm\Internal\LanguageLevelTypeAware;
 
 /**
  * The SoapClient class provides a client for SOAP 1.1, SOAP 1.2 servers. It can be used in WSDL
@@ -261,7 +262,7 @@ class SoapClient  {
      * @throws SoapFault A SoapFault exception will be thrown if the wsdl URI cannot be loaded.
      * @since 5.0.1
      */
-    public function __construct ($wsdl, array $options = null) {}
+	public function __construct (#[LanguageLevelTypeAware(["7.0" => "string", "7.1" => "?string"], default: "")] $wsdl, array $options = null) {}
 
 	/**
 	 * @link https://php.net/manual/en/soapclient.call.php
@@ -271,7 +272,7 @@ class SoapClient  {
 	 * @since 5.0.1
 	 */
 	#[Deprecated]
-	public function __call ($name, $args) {}
+	public function __call (#[LanguageLevelTypeAware(["7.0" => "string"], default: "")] $name, $args) {}
 
 	/**
 	 * Calls a SOAP function
@@ -314,8 +315,7 @@ class SoapClient  {
 	 * option set to <b>FALSE</b>, a SoapFault object will be returned.
 	 * @since 5.0.1
 	 */
-	public function __soapCall ($name, $args, $options = null, $inputHeaders = null, &$outputHeaders = null) {}
-
+	public function __soapCall (#[LanguageLevelTypeAware(["7.0" => "string"], default: "")] $name, #[LanguageLevelTypeAware(["7.0" => "array"], default: "")] $args, #[LanguageLevelTypeAware(["7.0" => "array", "7.1" => "?array"], default: "")] $options = null, $inputHeaders = null, #[LanguageLevelTypeAware(["7.0" => "array"], default: "")] &$outputHeaders = null) {}
 	/**
 	 * Returns last SOAP request
 	 * @link https://php.net/manual/en/soapclient.getlastrequest.php
@@ -395,7 +395,7 @@ class SoapClient  {
 	 * @return string The XML SOAP response.
 	 * @since 5.0.1
 	 */
-	public function __doRequest ($request, $location, $action, $version, $oneWay = false) {}
+	public function __doRequest (#[LanguageLevelTypeAware(["7.0" => "string"], default: "")] $request, #[LanguageLevelTypeAware(["7.0" => "string"], default: "")] $location, #[LanguageLevelTypeAware(["7.0" => "string"], default: "")] $action, #[LanguageLevelTypeAware(["7.0" => "int"], default: "")] $version, #[LanguageLevelTypeAware(["7.0" => "bool"], default: "")] $oneWay = false) {}
 
 	/**
 	 * The __setCookie purpose
@@ -409,7 +409,7 @@ class SoapClient  {
 	 * @return void No value is returned.
 	 * @since 5.0.4
 	 */
-	public function __setCookie ($name, $value = null) {}
+	public function __setCookie (#[LanguageLevelTypeAware(["7.0" => "string"], default: "")] $name, #[LanguageLevelTypeAware(["7.0" => "string", "7.1" => "?string"], default: "")] $value = null) {}
 
 	/**
 	 * Sets the location of the Web service to use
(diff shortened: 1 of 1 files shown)


sakarikl:
https://youtrack.jetbrains.com/issue/WI-58728

soap/soap.php
```diff
@@ -261,7 +262,135 @@ public function SoapClient ($wsdl, array $options = null) {}
      * @throws SoapFault A SoapFault exception will be thrown if the wsdl URI cannot be loaded.
      * @since 5.0.1
      */
-    public function __construct ($wsdl, array $options = null) {}
+	#[PhpStormStubsElementAvailable(to: '5.6')]
```
isfedorov, line 265: Could you please clarify why you choose `PhpStormStubsElementAvailable` but not `LanguageLevelTypeAware` here? Type hints in parameters could be customized using the latter one.

soap/soap.php
```diff
@@ -261,7 +262,135 @@ public function SoapClient ($wsdl, array $options = null) {}
...
+	 * <b>SOAP_SSL_METHOD_TLS</b>,
+	 * <b>SOAP_SSL_METHOD_SSLv2</b>,
+	 * <b>SOAP_SSL_METHOD_SSLv3</b> or
+	 * <b>SOAP_SSL_METHOD_SSLv23</b>.
+	 * </p>
+	 * @throws SoapFault A SoapFault exception will be thrown if the wsdl URI cannot be loaded.
+	 * @since 5.0.1
+	 */
+	#[PhpStormStubsElementAvailable(from: '7.0')]
+	public function __construct (?string $wsdl, array $options = []) {}
```
isfedorov, line 393: The method has `@since` 5.0.1 but uses nullable type-hint. This can cause errors in the case of adding a child class with language level < 7.1

soap/soap.php
```diff
@@ -271,8 +400,20 @@ public function __construct ($wsdl, array $options = null) {}
...
 
+	/**
+	 * @link https://php.net/manual/en/soapclient.call.php
+	 * @param string $name
+	 * @param array $args
+	 * @return mixed
+	 * @since 5.0.1
+	 */
+	#[Deprecated]
+	#[PhpStormStubsElementAvailable(from: '7.0')]
```
isfedorov, line 414: Let's use `LanguageLevelTypeAware` instead of `PhpStormStubsElementAvailable` to get rid of duplicated entities

isfedorov (changes requested):
Overall: Let's use `LanguageLevelTypeAware` instead of `PhpStormStubsElementAvailable` to get rid of duplicated entities.

soap/soap.php
sakarikl: can LanguageLevelTypeAware use multiple different versions? as 7.0 and 7.1 have differencies also. My commit did not yet take those into account.

sakarikl: meaning whole file changes.

soap/soap.php
sakarikl: Should that @since be 7.1? but the function existed already. signature only changed

soap/soap.php
isfedorov: Yes, sure. Multiple versions can be declared like `#[LanguageLevelTypeAware(['8.0' => 'bool', "7.0" => 'int', '7.1' => 'string'], default: 'float')]`

sakarikl: how about default changing? it has changed from null to '' in some cases. Sorry if asking basics, but did not found any documentation for that.

isfedorov: The default value doesn't matter too much actually as this is just a stub and if a parameter will not be provided in the code, the default value from the extension will be used in runtime. So if the extension has default value `''` for some parameter then in runtime `''` will be used despite stub will be declared with default value `null`. But it worth mentioning that we still have a separate test that compares default values of parameters with reflection for sake of stubs cleanliness

orklah: I'd object that PHP EA has an inspection that suggest dropping a literal parameter that has the same value as the default.

This means it could make a wrong suggestion if the default value is not the right one

isfedorov: >This means it could make a wrong suggestion if the default value is not the right one
Yes, that's right. From the inspection's point of view, a suggestion can be wrong depending on the language version. But unfortunately, we don't support LanguageLevelAwareness for default values yet to properly process such cases.

sakarikl: Would using  PhpStormStubsElementAvailable solve that?

sakarikl: Does LanguageLevelTypeAware support return types?

isfedorov: Can't say about EA but regarding PhpStorm the answer is "No". At least not yet. If you think it would be useful could you please create a ticket in YouTrack to highlight the redundant optional argument depending on `PhpStormStubsElementAvailable` attribute?

sakarikl: changed the PR to use LanguageLevelTypeAware

soap/soap.php
```diff
@@ -261,7 +262,7 @@ public function SoapClient ($wsdl, array $options = null) {}
      * @throws SoapFault A SoapFault exception will be thrown if the wsdl URI cannot be loaded.
      * @since 5.0.1
      */
-    public function __construct ($wsdl, array $options = null) {}
+	public function __construct (#[LanguageLevelTypeAware(["7.0" => "string"], ["7.1" => "?string"], default: "")] $wsdl, array $options = null) {}
```
isfedorov, line 265: Please note that not a set of arrays but array of key => value should be used instead. `#[LanguageLevelTypeAware(["7.0" => "string", "7.1" => "?string"], default: "")`.

soap/soap.php
```diff
@@ -261,7 +262,7 @@ public function SoapClient ($wsdl, array $options = null) {}
      * @throws SoapFault A SoapFault exception will be thrown if the wsdl URI cannot be loaded.
      * @since 5.0.1
      */
-    public function __construct ($wsdl, array $options = null) {}
+	public function __construct (#[LanguageLevelTypeAware(["7.0" => "string", "7.1" => "?string"], default: "")] $wsdl, array $options = null) {}
```
isfedorov, line 265: default type set to `""` doesn't make any sense as in this case type will be taken from phpdoc. So as far as $wsdl can be `string` or `?string` I think it makes sense to set attribute like `#[LanguageLevelTypeAware(["7.1" => "?string"], default: "string")].

sakarikl: But before that there was no type in signature. The stub would be wrong for PHP 5. If I understood correctly.

isfedorov: That's right. there was no type in signature and currently, also there is no type in signature. Type in the attribute is just a way to clarify a type depending on language version but it is not a replacement for type in signature. So declaring `#[LanguageLevelTypeAware(["7.1" => "?string"], default: "string")]` we will define that by default parameter should be `string` (we can't pass some object here, right?), but since PHP 7.1 it also can be null (or nullable string). Hope this shade some light to this attribute.

sakarikl: But then that stub is still wrong for PHP7 So the correct way to do this would be to make duplicate stubs. and in some cases you need at least 3 different stubs to get all possible combinations.

isfedorov: >But then that stub is still wrong for PHP7

Why? The stub will provide `string` as parameter type for language level <= 7 and `?string` for language level >=7.1. In case of stubs, it's not required for a type to be declared exactly in the signature. It also can be declared in PhpDoc for example. The difference only is that type declared in PhpDoc is applied for all language levels but a type declared in an attribute is applied for the language level defined in the attribute.

sakarikl: Fatal error: Uncaught TypeError: SoapClient::__construct(): Argument #1 ($wsdl) must be of type ?string, int given

It does have types

isfedorov:
Fixed in #1096

PR #1057, opened 2021-02-25
Support for multiple equal classes/interfaces in stubs

isfedorov:


tests/Model/StubsContainer.php
```diff
@@ -69,13 +70,24 @@ public function addFunction(PHPFunction $function): void
...
+     * @param string $name
+     * @param string|null $sourceFilePath
+     * @return PHPClass|null
+     * @throws RuntimeException
+     */
+    public function getClass(string $name, ?string $sourceFilePath = null): ?PHPClass
     {
-        if (array_key_exists($name, $this->classes) && isset($this->classes[$name])) {
-            return $this->classes[$name];
+        $classes = array_filter($this->classes, fn(PHPClass $class): bool => $class->name === $name);
```
MaXal, line 81: Let's preserve backward compatibility and return an element if it's the only one. The `sourceFilePath` based resolve will kick in only if we find duplicates.

tests/Model/StubsContainer.php
```diff
@@ -92,30 +104,43 @@ public function getClasses(): array
...
+     * @param string $name
+     * @param string|null $sourceFilePath
+     * @return PHPInterface|null
+     * @throws RuntimeException
+     */
+    public function getInterface(string $name, ?string $sourceFilePath = null): ?PHPInterface
     {
-        if (array_key_exists($name, $this->interfaces) && isset($this->interfaces[$name])) {
-            return $this->interfaces[$name];
+        $interfaces = array_filter($this->interfaces, fn(PHPInterface $interface): bool => $interface->name === $name);
```
MaXal, line 134: Same as above.

MaXal (changes requested):

PR #1068, opened 2021-03-10
Fix codestyle

SerafimArts:
When adding attributes to Reflection (at 03.11.2020, I could not find the original PR), the PSR codestyle was broken. 

This PR fixes this problem.

### DIFF
--- a/Reflection/ReflectionAttribute.php
+++ b/Reflection/ReflectionAttribute.php
@@ -32,7 +32,7 @@ class ReflectionAttribute
      * @since 8.0
      */
     #[Pure]
-	public function getName()
+    public function getName()
     {
     }
 
@@ -43,7 +43,7 @@ class ReflectionAttribute
      * @since 8.0
      */
     #[Pure]
-	public function getTarget()
+    public function getTarget()
     {
     }
 
@@ -54,7 +54,7 @@ class ReflectionAttribute
      * @since 8.0
      */
     #[Pure]
-	public function isRepeated()
+    public function isRepeated()
     {
     }
 
@@ -65,7 +65,7 @@ class ReflectionAttribute
      * @since 8.0
      */
     #[Pure]
-	public function getArguments()
+    public function getArguments()
     {
     }
 
--- a/Reflection/ReflectionClass.php
+++ b/Reflection/ReflectionClass.php
@@ -83,7 +83,7 @@ class ReflectionClass implements Reflector
      * @return string The class name.
      */
     #[Pure]
-	public function getName()
+    public function getName()
     {
     }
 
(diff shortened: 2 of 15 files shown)


isfedorov:
@SerafimArts what is the profit of this PR? Please see https://github.com/JetBrains/phpstorm-stubs/blob/master/CONTRIBUTING.md#code-style
1) These changes will not bring any benefits to end-users.
2) They are not strengthened with any tests so the next commit could revert them back without any warning

SerafimArts:
@isfedorov That is why I had to rewrite the entire reflection package: https://github.com/JetBrains/phpstorm-stubs/pull/857 =) Are you suggesting that I do it again?

isfedorov:
@SerafimArts ok, let's fix code styles. But I suggest fixing the problem generally then. Please add appropriate code-style setting ('indentation_type' => true) to php_cs config, fix all the files and commit them.

SerafimArts:
@isfedorov it may be easier to solve this problem once and for all with `'@PSR12' => true`?

isfedorov:
@SerafimArts I thought about PSR12 but I have a concern that with this style all stubs like 
`public function getContentEncoding() {}` 
will be changed to
```
public function getContentEncoding()
{
}
```
And it seems that placing braces to separate lines is redundant in case of stubs. Probably it will make sense to add some custom cs fixer rule for this case.

SerafimArts:
There is no such option in the fixer at all =\

There was a similar problem in vimeo psalm. They solved it with a special file format: `*.phpstub`, where semicolons can be used for methods in any structures, like: `public function getContentEncoding();`.

isfedorov:
>There is no such braces option modification in the fixer at all
That's why I wrote that `Probably it will make sense to add some custom cs fixer rule for this case.` :) Will think what else can be done to get the expected code style and if there will be no other thoughts will add a custom rule. If you would like to participate you are welcome to try to add such cs fixer rule by yourself.

SerafimArts:
On the other hand, maybe this is not so critical? All PHP developers are already accustomed to this format and they will perceive it adequately.

What's more, empty curly braces don't get in the way of anyone creating their own exception classes in "real" production code, like:
```php
class MyException extends \LogicException
{
}
```

This is a similar case, which is perceived by everyone as normal.

> Will think what else can be done to get the expected code style and if there will be no other thoughts will add a custom rule.

I think that this may not be custom rule, but a PR to the phpcs fixer repository here: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.18/src/Fixer/Basic/BracesFixer.php

By adding an option there, for example: `allow_single_line_methods_with_empty_body`

isfedorov:
Fixed in #1075

PR #1069, opened 2021-03-10
Add expected arguments for phpinfo()

pronskiy:


standard/standard_0.php
```diff
@@ -1091,7 +1092,7 @@ function image_type_to_extension (int $image_type, $include_dot = true): string|
  * </p>
  * @return bool true on success or false on failure.
  */
-function phpinfo (int $flags = INFO_ALL): bool
+function phpinfo (#[ExpectedValues(values: [INFO_GENERAL, INFO_CREDITS, INFO_CONFIGURATION, INFO_MODULES, INFO_ENVIRONMENT, INFO_VARIABLES, INFO_LICENSE, INFO_ALL])] int $flags = INFO_ALL): bool
```
isfedorov, line 1095: as far as several flags can be passed via bitwise `or` it makes sense to use `flags:` instead of `values:`

PR #1072, opened 2021-03-11
fix typo in standard_6.php

successgo:


standard/standard_6.php
```diff
@@ -338,7 +338,7 @@ function stream_socket_client (string $address, &$error_code, &$error_message, ?
  * Depending on the environment, Unix domain sockets may not be available.
  * A list of available transports can be retrieved using
  * stream_get_transports. See
- * for a list of bulitin transports.
+ * for a list of builtin transports.
```
isfedorov, line 341: @successgo I suggest to fix it even more and change to "built-in" :)

isfedorov (changes requested):

standard/standard_6.php
successgo: Good

PR #1074, opened 2021-03-17
[WI-59284] Fix SoapVar constructor typehints

InvisibleSmiley:


isfedorov:
Fixed in https://github.com/JetBrains/phpstorm-stubs/commit/56ca769287d44b799fea9396e279bfec1097ccb3

PR #1081, opened 2021-03-19
[WI-59335] Document getPrevious result nullability

InvisibleSmiley:


InvisibleSmiley:
Code Inspection seemed to have failed internally. Please help.

PR #1101, opened 2021-04-01
Remove Pure attribute on realpath

ancarda:
This function depends on side-effects, as can be demonstrated with this minimal test case:

```php
var_dump(realpath('foo')); // bool(false)
mkdir('foo');
var_dump(realpath('foo')); // string(8) "/tmp/dir"
```

https://youtrack.jetbrains.com/issue/WI-59567

### DIFF
--- a/standard/standard_6.php
+++ b/standard/standard_6.php
@@ -1137,7 +1137,7 @@ function socket_get_status($stream): array {}
  * the file does not exist.
  * </p>
  */
-#[Pure]
+#[Pure(true)]
 function realpath(string $path): string|false {}
 
 /**


isfedorov (changes requested):
@ancarda `#[Pure]` in context of phpstorm attributes means that function doesn't have side effects (like print something, write data etc) but your case isn't exactly about side effect. Your case is about different output with the same input that caused by the different environment (created folder in the example). `#[Pure]` attribute has a separate key for this
```
class Pure
{
    /**
     * @param bool $mayDependOnGlobalScope Whether the function result may be dependendent on anything except passed variables
     */
    public function __construct(bool $mayDependOnGlobalScope = false) {}
}
```
So the correct way would be just to change attribute to `#[Pure(true)]` for the function.

isfedorov (approved):

PR #1106, opened 2021-04-08
Delete removed annotations from DateTime constants

Tobion:
PhpStorm marks code using these constants as broken as they have been removed. But they can still be used as they are inherited from the DateTimeInterface.
In theory we could remove the constants from DateTime completely. But the interface constants have `since 7.2` which would also be wrong for DateTime constants.
So I think this is the best solution.

Ref. #951

### DIFF
--- a/date/date_c.php
+++ b/date/date_c.php
@@ -377,69 +377,18 @@ class DateTimeImmutable implements DateTimeInterface
  */
 class DateTime implements DateTimeInterface
 {
-    /**
-     * @removed 7.2
-     */
     public const ATOM = 'Y-m-d\TH:i:sP';
-
-    /**
-     * @removed 7.2
-     */
     public const COOKIE = 'l, d-M-Y H:i:s T';
-
-    /**
-     * @removed 7.2
-     */
     public const ISO8601 = 'Y-m-d\TH:i:sO';
-
-    /**
-     * @removed 7.2
-     */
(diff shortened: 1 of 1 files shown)


isfedorov:
Hi @Tobion, this is intentional tag to avoid extra errors about inheriting of constant from interface with language level > 7.2
Screenshot 2021-04-11 at 00 56 43
Also referenced in https://youtrack.jetbrains.com/issue/WI-56949

Tobion:
Well, if it solves a warning but introduces another warning, I don't see the advantage. Or how do you suggest to solve the new warning?

isfedorov:
@Tobion could you please create a YT ticket for the case when you see this another warning? Probably it's covered by https://youtrack.jetbrains.com/issue/WI-57005 or https://youtrack.jetbrains.com/issue/WI-14529

Tobion:
Yeah WI-57005 should cover this.

guilliamxavier:
@isfedorov: I'm disappointed that you don't accept this PR, at least while waiting for the bug to be fixed (the "extra errors" you want to avoid are only in PhpStorm own stubs, while this is everywhere in user code...)

MaXal:
The [issue](https://youtrack.jetbrains.com/issue/WI-57005) has been fixed and the fix will be available in 2021.2.

PR #1112, opened 2021-04-15
Two implementations of merging strtr

yanlong-li:
Avoid the repeated definition of strtr, and solve the problem of no syntax prompt in array mode

[php manual strtr](https://www.php.net/manual/en/function.strtr.php#refsect1-function.strtr-description)

### DIFF
--- a/standard/standard_1.php
+++ b/standard/standard_1.php
@@ -645,10 +645,13 @@ function ucwords(string $string, string $separators = " \t\r\n\f\v"): string {}
  * @param string $string <p>
  * The string being translated.
  * </p>
- * @param string $from <p>
- * The string replacing from.
+ * @param array|string $replace_pairs <p>
+ * The replace_pairs parameter may be used as a substitute for to and from in which case it's an array in the form array('from_str' => 'to_str', ...).
+ * Parameter to needs to be null.
+ * <br/>
+ * Or string for from in the form string. Parameter to is required and is a string.
  * </p>
- * @param string $to <p>
+ * @param ?string $to <p>
  * The string being translated to to.
  * </p>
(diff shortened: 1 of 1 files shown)


standard/standard_1.php
```diff
@@ -645,10 +645,13 @@ function ucwords(string $string, string $separators = " \t\r\n\f\v"): string {}
...
- * @param string $from <p>
- * The string replacing from.
+ * @param array|string $replace_pairs <p>
+ * The replace_pairs parameter may be used as a substitute for to and from in which case it's an array in the form array('from_str' => 'to_str', ...).
+ * Parameter to needs to be null.
+ * <br/>
+ * Or string for from in the form string. Parameter to is required and is a string.
  * </p>
- * @param string $to <p>
+ * @param ?string $to <p>
```
williamdes, line 654:
```suggestion
 * @param string|null $to <p>
```

DmitryTronin:
Hello @yanlong-li 
Unfortunately we have to decline the PR. There's a couple of reasons for this, one of them is that this is a historical design to have multiple definitions for functions with alternative signatures: https://youtrack.jetbrains.com/issue/WI-9048. We've hardcoded a fix for a similar case: `implode()` but, currently, that is not something we'd like to keep implementing.
Another reason: accepting this PR would cause false negative for a second parameter when only 2 params are used:


Next, you were mentioning an issue with missing syntax prompt in array mode. Could you please illustrate this? This is working well on a default installation:


Generally, we're planning to address the problem with duplicates in stubs in some other way. Here's a link to a corresponding report that you can follow: https://youtrack.jetbrains.com/issue/WI-56936.

PR #1114, opened 2021-04-15
Fixed php-cs-fixer crashing when run directly

dktapps:


williamdes (approved):
🤔

isfedorov:
Due to conflicting files fixed it in a separate PR #1123

PR #1115, opened 2021-04-18
add boolean values to sqlite3 returns

fentie:
Fixes https://youtrack.jetbrains.com/issue/WI-60012

### DIFF
--- a/sqlite3/sqlite3.php
+++ b/sqlite3/sqlite3.php
@@ -178,7 +178,7 @@ class SQLite3
      * @param string $query <p>
      * The SQL query to execute.
      * </p>
-     * @return SQLite3Result an <b>SQLite3Result</b> object if the query returns results. Otherwise,
+     * @return SQLite3Result|bool an <b>SQLite3Result</b> object if the query returns results. Otherwise,
      * returns <b>TRUE</b> if the query succeeded, <b>FALSE</b> on failure.
      */
     public function query($query) {}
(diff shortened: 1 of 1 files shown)


sqlite3/sqlite3.php
```diff
@@ -178,7 +178,7 @@ public function prepare($query) {}
      * @param string $query <p>
      * The SQL query to execute.
      * </p>
-     * @return SQLite3Result an <b>SQLite3Result</b> object if the query returns results. Otherwise,
+     * @return SQLite3Result|bool an <b>SQLite3Result</b> object if the query returns results. Otherwise,
```
williamdes, line 181:
```suggestion
     * @return SQLite3Result|true an <b>SQLite3Result</b> object if the query returns results. Otherwise,
```

williamdes (approved):

fentie:
So what's the next step here? Is someone with write access going to merge this?

sqlite3/sqlite3.php
isfedorov, line 181: `bool` can be replaced with `false` here as according to  https://github.com/php/php-src/blob/master/ext/sqlite3/sqlite3.stub.php#L60 and https://github.com/php/php-src/blob/master/ext/sqlite3/sqlite3.c#L541 seems method can return `SQLite3Result|false` but can not return `true`

isfedorov:
Hi @fentie 
>So what's the next step here? Is someone with write access going to merge this?

Thank you for the contribution and sorry for the long response. It was vacation time :-) I've left a minor suggestion in the PR. Other than that PR is good to go and after the suggested fix will be merged to master.

PR #1116, opened 2021-04-20
Allow 'resource' as key parameter for PHP7

LeonMelis:
In PHP7 the $key parameters in openssl are expected to be 'resource' instances.
`OpenSSLAsymmetricKey` is only available in PHP8.

For example, consider the following code:

```php
$key = openssl_pkey_new();
openssl_csr_new(['example.com'], $key);
```
When setting the project to PHP7, PHPStorm will complain about the $key parameter not being an `OpenSSLAsymmetricKey` instance.


williamdes (approved):

isfedorov:
Hi @LeonMelis 
Thank you for the contribution! As far as `resource` was accepted in PHP < 8 it's better to use the attribute `#[LanguageLevelTypeAware]` for a parameter like `function openssl_pkey_export(#[LanguageLevelTypeAware(['8.0' => 'OpenSSLAsymmetricKey|OpenSSLCertificate|array|string'], default: 'resource|array|string')] $key, &$output, ?string $passphrase, ?array $options): bool {}`. Thus type inference and code inspections will be more strict and more correct

isfedorov:
Fixed in other PR

PR #1117, opened 2021-04-22
Add oci8 3.0 stubs for php 8

nightlinus:
As of https://pecl.php.net/package-changelog.php?package=oci8 
class rename + missing stubs


oci8/oci8.php
```diff
@@ -5,6 +5,7 @@
 /**
  * OCI8 LOB functionality for large binary (BLOB) and character (CLOB) objects.
  * @link https://php.net/manual/en/class.OCI-Lob.php
+ * @removed 8.0
```
williamdes, line 8:
```suggestion
 * @deprecated removed in 8.0
```
Would be better supported by tools and IDEs

oci8/oci8v3.php
```diff
@@ -0,0 +1,2710 @@
...
+     * </p>
+     * @return bool <b>TRUE</b> on success or <b>FALSE</b> on failure.
+     */
+    public function save($data, $offset = null) {}
+
+    /**
+     * (PHP 5, PECL OCI8 &gt;= 1.1.0)<br/>
+     * Alias of {@see OCILob::import}
+     * @link https://php.net/manual/en/oci-lob.savefile.php
+     * @param $filename
```
williamdes, line 261:
```suggestion
     * @param string $filename
```
Could you check for any other `@param $` block to add the types please ?

oci8/oci8v3.php
```diff
@@ -0,0 +1,2710 @@
...
+ */
+function ocicollgetelem($collection, $index) {}
+
+/**
+ * (PHP 4 &gt;= 4.0.6, PHP 5, PECL OCI8 &gt;= 1.0.0)<br/>
+ * Alias of {@see OCICollection::assignElem}
+ * @link https://php.net/manual/en/function.ocicollassignelem.php
+ * @param OCICollection $collection
+ * @param $index <p>The element index. First index is 0.</p>
+ * @param $value <p>Can be a string or a number.</p>
```
williamdes, line 2194:
```suggestion
 * @param int $index <p>The element index. First index is 0.</p>
 * @param int|string $value <p>Can be a string or a number.</p>
```

isfedorov:
Hi @nightlinus 
Thank you for the contribution! Newly added `oci8v3.php` contains functions that already exist in `oci8.php`. If new functions differ from old ones, please mark them with the appropriate version using `@since` or `PhpStormStubsElementAvailable` attribute. If those functions equal to the old ones probably it makes sense to remove duplicates and leave functions only in one of files.

isfedorov:
fixed in a separate PR

PR #1119, opened 2021-04-28
WI-60175: Fix invlalid links in XMLWriter class

RabihMonzer:


RabihMonzer:
**Note:** I also suggest that we restructure the class to follow the PSR coding standards (In a separate task). Example: Adding types to method parameters, returns types of methods (instead of annotating the return type), leaving propper blank lines in annotations, etc.

williamdes (approved):

isfedorov:
@RabihMonzer 
Thank you for the PR! Could you please rebase your branch on master as there was update of cs-fixer in master and currently tests fails on your branch due to old cs-fixer

PR #1120, opened 2021-04-29
Update @link references for intl's NumberFormatter

sbol-coolblue:
I noticed that the constant links were pointing to a page that did not actually cover these constants.

### DIFF
--- a/intl/intl.php
+++ b/intl/intl.php
@@ -435,49 +435,49 @@ class NumberFormatter
 
     /**
      * Decimal format defined by pattern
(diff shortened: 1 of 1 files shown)


williamdes (approved):

isfedorov:
@sbol-coolblue 
Thank you for the PR! Could you please rebase your branch on master as there was update of cs-fixer in master and currently tests fails on your branch due to old cs-fixer

williamdes (approved):

PR #1133, opened 2021-05-13
update stubs for couchbase 3.1.2

avsej:
Note that braces were moved to new line after I run cs-fix on the file

### DIFF
--- a/couchbase_v3/couchbase.php
+++ b/couchbase_v3/couchbase.php
@@ -1,7 +1,7 @@
 <?php
 /**
  * Couchbase extension stubs
- * Gathered from https://docs.couchbase.com/sdk-api/couchbase-php-client-3.1.0/index.html
+ * Gathered from https://docs.couchbase.com/sdk-api/couchbase-php-client-3.1.2/index.html
  * Maintainer: sergey@couchbase.com
  *
  * https://github.com/couchbase/php-couchbase/tree/master/api
@@ -65,10 +65,10 @@
 
 namespace Couchbase;
 
-use DateTimeInterface;
-use JsonSerializable;
-use Exception;
-use Throwable;
+use \JsonSerializable;
+use \Exception;
+use \Throwable;
+use \DateTimeInterface;
 
 /**
  * An object which contains meta information of the document needed to enforce query consistency.
@@ -421,6 +421,7 @@ interface AnalyticsResult
     public function rows(): ?array;
 }
 
+
 /**
  * A range (or bucket) for a term search facet result.
  * Counts the number of occurrences of a given term.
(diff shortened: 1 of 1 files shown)


isfedorov:
> Note that braces were moved to new line after I run cs-fix on the file

@avsej Probably you should rebase on master and perform `composer cs-fix` after that as there is exactly vice-versa cs-fixer rule to place braces on the same line like `foo() {}`

avsej:
this is what I did. maybe the issue in docker image or something like that.

isfedorov:
@avsej I've just tried to run `docker-compose -f docker-compose.yml run php composer cs-fix -d /opt/project/phpstorm-stubs` on your fork and file has been formatted as expected 
Screenshot 2021-05-13 at 17 21 19

avsej:
If it is not too hard for you, could you please update my PR with your result? I'm using CentOS 8, and when I run `podman-compose` -- it raises errors, so I've used just image without compose and it seems like it does not pick up correct settings.

couchbase_v3/couchbase.php
```diff
@@ -1516,7 +1532,7 @@ public function setName(string $name): CollectionSpec {}
 
     public function setScopeName(string $name): CollectionSpec {}
 
-    public function setMaxExpiry(float $ms): CollectionSpec {}
+    public function setMaxExpiry(long $ms): CollectionSpec {}
```
isfedorov, line 1535: This is incorrect change as there is no such typehint

couchbase_v3/couchbase.php
```diff
@@ -2827,7 +2843,7 @@ public function timeout(int $arg): GetAllReplicasOptions {}
      *
      *   `function decoder(string $bytes, int $flags, int $datatype): mixed`
      */
-    public function decoder(callable $arg): GetAllReplicasOptions {}
+    public function decoder(callable $arg): GetAllRepliacasOptions {}
```
isfedorov, line 2846: Seems there is a typo in class name

couchbase_v3/couchbase.php
avsej: fixed

couchbase_v3/couchbase.php
avsej: thanks, I've fixed this too

avsej:
@isfedorov, also I wonder if it is possible to swap directories like this:
* `couchbase` -> `couchbase_v2`
* `couchbase_v3` -> `couchbase`
Because V2 is not maintained anymore, and reached EOL, so people are getting confused if they install current version, but get stubs for old one. The old version V2 still makes sense if users are in process of migrating, but for new users it is really confusing.

isfedorov:
> @isfedorov, also I wonder if it is possible to swap directories like this:

@avsej I think it's possible. Please do it in a separate PR but please note that `generate-stub-map` should be updated as well as it uses hardcoded `couchbase_v3`.

avsej:
Ok, I will do it once this patch will be merged

avsej:
@isfedorov I guess this PR is ready, or there is some pending changes or open questions?

isfedorov:
@avsej PR seems to be ok to merge and will be merged soon. The only extra item for us is some of our internal tests that also use couchbase and they also should be updated.

PR #1134, opened 2021-05-14
DateTimeImmutable::modify can return false

camporter:
DateTimeImmutable::modify can return false on modification failure, update the docblock to reflect this.

### DIFF
--- a/date/date_c.php
+++ b/date/date_c.php
@@ -218,7 +218,7 @@ class DateTimeImmutable implements DateTimeInterface
      * @link https://secure.php.net/manual/en/datetimeimmutable.modify.php
      * @param string $modifier  <p>A date/time string. Valid formats are explained in
      * {@link https://secure.php.net/manual/en/datetime.formats.php Date and Time Formats}.</p>
-     * @return static
+     * @return static|false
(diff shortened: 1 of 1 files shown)


date/date_c.php
```diff
@@ -218,7 +218,7 @@ public static function getLastErrors() {}
      * @link https://secure.php.net/manual/en/datetimeimmutable.modify.php
      * @param string $modifier  <p>A date/time string. Valid formats are explained in
      * {@link https://secure.php.net/manual/en/datetime.formats.php Date and Time Formats}.</p>
-     * @return static
+     * @return static|false
```
williamdes, line 221: maybe you can add a comment about on what condition it returns false

williamdes (approved):

date/date_c.php
camporter: Should I move the PHP doc line just below onto this line? Wasn't sure whether that was done or if I should keep it separate?

williamdes: You can do like `* @return static|false False is returned ...` but avoid `* @return static|false is returned ...`
`* @return static|falseis returned ...` is obviously invalid

PR #1136, opened 2021-05-18
Update fgetcsv and fputcsv with $delimiter instead of $seperator

alexander-schranz:
See https://github.com/php/doc-en/pull/606 and https://github.com/php/php-src/blob/ac47162add6fa529836467929e7bcf20025638b7/ext/standard/file.c#L1985


williamdes (approved):

alexander-schranz:
Will set this on hold until php/doc-en#606  is solved

alexander-schranz:
stubs are correct there is correct for php 8 there is an issue in the php.net docs: https://3v4l.org/64Kch

PR #1140, opened 2021-05-25
Update type hints for session

Simbiat:
Updating type hints to what's currently reflected on php.net as per https://youtrack.jetbrains.com/issue/WI-60588

### DIFF
--- a/session/SessionHandler.php
+++ b/session/SessionHandler.php
@@ -168,7 +168,7 @@ class SessionHandler implements SessionHandlerInterface, SessionIdInterface
      * </p>
      * @since 5.4
      */
-    public function close() {}
+    public function close(): bool {}
 
     /**
      * Return a new session ID
@@ -176,7 +176,7 @@ class SessionHandler implements SessionHandlerInterface, SessionIdInterface
      * @return string <p>A session ID valid for the default session handler.</p>
      * @since 5.5.1
      */
-    public function create_sid() {}
+    public function create_sid(): string {}
 
     /**
      * Destroy a session
@@ -188,7 +188,7 @@ class SessionHandler implements SessionHandlerInterface, SessionIdInterface
      * </p>
      * @since 5.4
      */
-    public function destroy($id) {}
+    public function destroy(string $id): bool {}
 
     /**
      * Cleanup old sessions
@@ -203,7 +203,7 @@ class SessionHandler implements SessionHandlerInterface, SessionIdInterface
      * </p>
      * @since 5.4
      */
-    public function gc($max_lifetime) {}
+    public function gc(int $max_lifetime): int|bool {}
 
     /**
      * Initialize session
@@ -216,7 +216,7 @@ class SessionHandler implements SessionHandlerInterface, SessionIdInterface
      * </p>
      * @since 5.4
      */
-    public function open($path, $name) {}
+    public function open(string $path, string $name): bool {}
 
     /**
      * Read session data
@@ -229,7 +229,7 @@ class SessionHandler implements SessionHandlerInterface, SessionIdInterface
      * </p>
      * @since 5.4
      */
-    public function read($id) {}
+    public function read(string $id): string|false {}
 
     /**
      * Write session data
@@ -248,7 +248,7 @@ class SessionHandler implements SessionHandlerInterface, SessionIdInterface
      * </p>
      * @since 5.4
      */
-    public function write($id, $data) {}
+    public function write(string $id, string $data): bool {}
 
     /**
      * Validate session id
@@ -257,7 +257,7 @@ class SessionHandler implements SessionHandlerInterface, SessionIdInterface
      * Note this value is returned internally to PHP for processing.
      * </p>
      */
-    public function validateId($session_id) {}
+    public function validateId(string $id): bool {}
 
     /**
      * Update timestamp of a session
@@ -271,5 +271,5 @@ class SessionHandler implements SessionHandlerInterface, SessionIdInterface
      * </p>
      * @return bool
      */
-    public function updateTimestamp($session_id, $session_data) {}
+    public function updateTimestamp(string $id, string $data): bool {}
 }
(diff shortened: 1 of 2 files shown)


williamdes (approved):

Simbiat:
Wow... Inspection found lots of issues... Or is this because it's using the older version of the code, that's being modified?

session/session.php
```diff
@@ -246,18 +246,19 @@ function session_unset() {}
  * @param callable $update_timestamp [optional]
  * @return bool true on success or false on failure.
  */
-function session_set_save_handler(callable $open, callable $close, callable $read, callable $write, callable $destroy, callable $gc, $create_sid, $validate_sid, $update_timestamp): bool {}
+function session_set_save_handler(callable $open, callable $close, callable $read, callable $write, callable $destroy, callable $gc, callable $create_sid = ?, callable $validate_sid = ?, callable $update_timestamp = ?): bool {}
```
DmitryTronin, line 249: ` = ?` isn't a valid PHP syntax here. Per discussion in https://github.com/php/php-src/pull/5366#issuecomment-611996429 it seems to me that ` = ?` should be either replaced with `= null` or omitted at all.
As you can see [here](https://github.com/php/php-src/blob/master/ext/session/session.stub.php#L46), default values are unknown: specifying this in our stubs using PHP syntax isn't possible & hardly would be beneficial. So omitting them seems to be the best way.

session/session.php
```diff
@@ -442,7 +443,7 @@ function session_status(): int {}
  * @since 5.6
  */
 #[LanguageLevelTypeAware(["7.2" => "bool"], default: "void")]
-function session_abort() {}
+function session_abort():bool {}
```
DmitryTronin, line 446: This is not required, since we're using `LanguageLevelTypeAware` here & it's working correctly:

session/SessionHandler.php
```diff
@@ -19,7 +19,7 @@ interface SessionHandlerInterface
      * </p>
      * @since 5.4
      */
-    public function close();
+    public function close(): bool;
```
DmitryTronin, line 22: Corresponding methods from `\SessionHandler` should also be updated:
https://github.com/JetBrains/phpstorm-stubs/blob/master/session/SessionHandler.php#L171
https://github.com/JetBrains/phpstorm-stubs/blob/master/session/SessionHandler.php#L179
https://github.com/JetBrains/phpstorm-stubs/blob/master/session/SessionHandler.php#L191
https://github.com/JetBrains/phpstorm-stubs/blob/master/session/SessionHandler.php#L206
https://github.com/JetBrains/phpstorm-stubs/blob/master/session/SessionHandler.php#L219
https://github.com/JetBrains/phpstorm-stubs/blob/master/session/SessionHandler.php#L232
https://github.com/JetBrains/phpstorm-stubs/blob/master/session/SessionHandler.php#L251

https://www.php.net/manual/en/class.sessionhandler.php

DmitryTronin (changes requested):

DmitryTronin:
Thank you for taking your time & submitting the proposal. I've commented on some of the found inspection reports: we would appreciate it if you could address the issues I've highlighted & submit new changes

session/session.php
Simbiat: Ah, missed it for the abort. Corrected.

Simbiat (commented):
Submitted changed as per requests.

session/SessionHandler.php
Simbiat: Oh. Missed them. Updated.

Simbiat:
Now only annotator fails, but, I do not think I can do anything about it, since this what we have in PHP itself.

session/SessionHandler.php
```diff
@@ -271,5 +271,5 @@ public function validateId($session_id) {}
      * </p>
      * @return bool
      */
-    public function updateTimestamp($session_id, $session_data) {}
+    public function updateTimestamp(string $id, string $data): bool {}
```
DmitryTronin, line 274: We can remove this method as well as `validateId()`: they are neither in https://www.php.net/manual/en/class.sessionhandler.php nor in https://github.com/php/php-src/blob/master/ext/session/session.stub.php#L102.
These functions are already in SessionUpdateTimestampHandlerInterface in the same file:


Checked where did they come from: https://github.com/JetBrains/phpstorm-stubs/pull/294/files/95d84d28035bdb56a30e5aeebfb6fa068c4e8c32#diff-17c933311fd7e06dd8c1fe0aef6896ddb65d877b9d50eaef405f61f5707269faR262 - they were added by a user in 2017. By the looks of it, it's safe to remove them now.

DmitryTronin (changes requested):

session/SessionHandler.php
Simbiat: Done

DmitryTronin:
Thanks!
Unfortunately the changes didn't pass PhpStorm Stubs Tests. Please allow me to explain why and what can be done here.

1. First 16 failed tests failed on checking the declared stubs type VS their actual type run by the runtime. Example: `Method SessionHandler::open has invalid return type`. With proposed changes it returns `bool` type while in fact it doesn't return anything: https://3v4l.org/v3ZBA. You can also see this hinted in docs on php.net: *Note this value is returned internally to PHP for processing.*.
So, to address this problem we have to remove return type hints from `SessionHandler`, `SessionHandlerInterface` & `SessionUpdateTimestampHandlerInterface` methods

2. Failures 17-48 are caused by `Method xxx with @since 'x.x'  has parameter 'x' with typehint 'x' but typehints available only since php 7`. Let's inspect `SessionHandler::destroy` for example: it has `@since 5.4` tag meaning that you theoretically can run the code that calls this method using PHP5.4. With `public function destroy(string $id)` signature this would cause PHP fatal error because typehints available only since php 7. 
I understand that php.net documentation & sources at php-src are all indicating a `string` typehint for this method, bit with language level < 7 that's not possible. Calling this method in 5.4 is possible: https://3v4l.org/SLJi6.
That means that all method typehint changes in SessionHandler and SessionHandlerInterface should be reverted.

PR #1142, opened 2021-05-25
Update New Relic stubs to extension >= 9.8

MarkG-IEL:
Extensive (90%+) reworking of the New Relic stub:  PHP 8 return types and parameter typing; updated link URIs; added "since" annotations; copy new documentation phrasing from New Relic website; add additional function definitions for newer extensions versions; deprecate functions as needed.


MarkG-IEL:
The errors are for an extension-defined class: https://docs.newrelic.com/docs/agents/php-agent/php-agent-api/newreliccreatedistributedtracepayload-php-agent-api/#return-values

Should I define that class in the docs too to prevent this error?

MarkG-IEL:
Qodana complained about a deprecated class that is supposed to be deprecated. Looks like it still passes.

isfedorov:
Hi @MarkG-IEL 
>Qodana complained about a deprecated class that is supposed to be deprecated.

Yep, qodana warning can be ignored in this case.
>Adding the namespaces and adding the DistributedTracePayload class has caused a 100% git change.

As well as reordering of functions I guess and that's really pity :( Review changes is much harder now. May be it's possible to move DistributedTracePayload to the bottom of file and restore old order of functions to avoid 100% git change?

isfedorov:
Also tests say that StubsMap should be regenerated (`docker-compose -f docker-compose.yml run php /usr/local/bin/php /opt/project/phpstorm-stubs/generate-stub-map`)

MarkG-IEL:
> Also tests say that StubsMap should be regenerated (`docker-compose -f docker-compose.yml run php /usr/local/bin/php /opt/project/phpstorm-stubs/generate-stub-map`)

Ah, yes. I put in the new class and forgot to rerun that.

As well as reordering of functions I guess and that's really pity :( Review changes is much harder now. May be it's possible to move DistributedTracePayload to the bottom of file and restore old order of functions to avoid 100% git change?

Let me give that a try.

MarkG-IEL:
@isfedorov I put everything back in order, but it only helps some. `newrelic_notice_error()` had such an extensive rewrite that git did that wonderful "shift everything down" change that makes it pretty difficult to follow after that. All new functions and the class are at the end. Looking at the C code for the class, I can't see any explicit deprecation, so I removed that notice. The function that returns that class is still deprecated, just not the class itself.

isfedorov:
@MarkG-IEL thank you for the update! At the first glance it's actually better now.

newrelic/newrelic.php
```diff
@@ -1,452 +1,647 @@
...
+ * @param int|null                       $errline                  [optional] The line number where the error
+ *                                                                 occurred.
+ * @param string|null                    $errcontext               [optional] An array that points to the symbol
+ *                                                                 table
+ *                                                                 that was active when the error occurred.
+ *
+ * @return null
+ */
+function newrelic_notice_error(
+    string|Throwable|Exception|int $messageOrExceptionOrCode,
```
isfedorov, line 301: Hmm..I don't think that joining several signatures into one will work well here because for example with such a signature it will be possible to make a call `newrelic_notice_error(1)` and IDE will not show any warning because such call will match this stub

MarkG-IEL: I agree, but I'm not sure PHPDoc has a good solution to this. It's basically IF firstParam IS int THEN secondParam IS Required. Does JetBrains have an internal PHP 8 Attribute that covers this?

isfedorov: Unfortunately no, we don't have such an attribute yet. Why don't you like an approach with several signatures in this case?

isfedorov: From the other side, this signature was also joined before and there were no any complains so probably we can live with such signature as well.

PR #1143, opened 2021-05-28
Actualize and improve FFI extension

SerafimArts:
- Add `FFI\CType::getName()` method. It is not documented, but exists since 8.0+
- Add the simplest rule for FFI autocomplete (`.phpstorm.meta.php`)
- Add autocomplete for FFI\CData


SerafimArts:
There are problems in CI. The `CData` class is `ArrayAccessible` and `Countable`, but it is not possible to call these methods directly.
```php
$cdata = \FFI::new('int[10]');

$cdata[0]->cdata = 42;    // Ok
$cdata->offsetSet(0, 42); // Error
```

In order for the PhpStorm to allow access to this type by index and not show an error, the `ArrayAccessible` interface with private access methods is required (exclusion from the autocomplete).

isfedorov (changes requested):
Hi @SerafimArts ,
One of tests fail with next message
```
There was 1 failure:

1) StubTests\StubsMetaExpectedArgumentsTest::testStringLiteralsSingleQuoted
String literals as expectedArguments should be single-quoted
```
Could you please fix the quotes?

SerafimArts:
> Could you please fix the quotes?

Sure. Done.

PR #1151, opened 2021-06-16
fixed missing return value

brainfoolong:
JsonSerializable::jsonSerialize is defined to have "mixed" as return type, will throw errors in php 8.1 if you not correctly define that return type
Error in php 8.1 example: `Declaration of xxx::jsonSerialize() should be compatible with xxx::jsonSerialize(): mixed`

### DIFF
--- a/json/json.php
+++ b/json/json.php
@@ -19,7 +19,8 @@ interface JsonSerializable
      * which is a value of any type other than a resource.
      * @since 5.4
      */
-    public function jsonSerialize(): mixed;
+    #[LanguageLevelTypeAware(['8.1'=>'mixed'], default: '')]
+    public function jsonSerialize();
 }
 
 class JsonIncrementalParser
--- a/json/json.php
+++ b/json/json.php
@@ -19,7 +19,7 @@ interface JsonSerializable
      * which is a value of any type other than a resource.
      * @since 5.4
      */
-    public function jsonSerialize();
+    public function jsonSerialize(): mixed;
 }
 
 class JsonIncrementalParser


isfedorov:
@brainfoolong while your fix is correct for PHP 8-8.1 it's wrong for PHP < 8 as far as mixed typehint was introduced in PHP 8 and thus overriding this method in PHP < 8 will produce false positive warnings. To avoid this please use #[LanguageLevelTypeAware] attribute like 
```
#[LanguageLevelTypeAware(['8.1'=>'mixed'], default: '')]
public function jsonSerialize();
```

brainfoolong:
Thanks for the hint. Changed it.

json/json.php
```diff
@@ -19,6 +19,7 @@ interface JsonSerializable
      * which is a value of any type other than a resource.
      * @since 5.4
      */
+    #[LanguageLevelTypeAware(['8.1'=>'mixed'], default: '')]
```
isfedorov, line 22: Please also add class `LanguageLevelTypeAware` to imports and perform `docker-compose -f docker-compose.yml run php composer cs-fix -d /opt/project/phpstorm-stubs` to fix code-style related problems

isfedorov (changes requested):

brainfoolong:
Sorry, i don't have a full blown development environment for this stubs. I can't fix more than this, it was a quick change with just a texteditor. Should i close this when i can't resolve it properly?

isfedorov:
Fixed in other PR

PR #1152, opened 2021-06-16
2nd attempt and typehinting for sessions

Simbiat:
https://github.com/JetBrains/phpstorm-stubs/pull/1140 failed, but as per suggestion in https://youtrack.jetbrains.com/issue/WI-60588 retrying the same with removal of `@since 5.4` as well.

### DIFF
--- a/session/SessionHandler.php
+++ b/session/SessionHandler.php
@@ -168,7 +168,7 @@ class SessionHandler implements SessionHandlerInterface, SessionIdInterface
      * </p>
      * @since 5.4
      */
-    public function close() {}
+    public function close(): bool {}
 
     /**
      * Return a new session ID
@@ -176,7 +176,7 @@ class SessionHandler implements SessionHandlerInterface, SessionIdInterface
      * @return string <p>A session ID valid for the default session handler.</p>
      * @since 5.5.1
      */
-    public function create_sid() {}
+    public function create_sid(): string {}
 
     /**
      * Destroy a session
@@ -188,7 +188,7 @@ class SessionHandler implements SessionHandlerInterface, SessionIdInterface
      * </p>
      * @since 5.4
      */
-    public function destroy($id) {}
+    public function destroy(string $id): bool {}
 
     /**
      * Cleanup old sessions
@@ -203,7 +203,7 @@ class SessionHandler implements SessionHandlerInterface, SessionIdInterface
      * </p>
      * @since 5.4
      */
-    public function gc($max_lifetime) {}
+    public function gc(int $max_lifetime): int|bool {}
 
     /**
      * Initialize session
@@ -216,7 +216,7 @@ class SessionHandler implements SessionHandlerInterface, SessionIdInterface
      * </p>
      * @since 5.4
      */
-    public function open($path, $name) {}
+    public function open(string $path, string $name): bool {}
 
     /**
      * Read session data
@@ -229,7 +229,7 @@ class SessionHandler implements SessionHandlerInterface, SessionIdInterface
      * </p>
      * @since 5.4
      */
-    public function read($id) {}
+    public function read(string $id): string|false {}
 
     /**
      * Write session data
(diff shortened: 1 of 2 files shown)


session/session.php
```diff
@@ -174,7 +174,7 @@ function session_start(array $options = []): bool {}
  * If it is used without active session, it omits collision check.
  * @since 7.1
  */
-function session_create_id(string $prefix): string|false {}
+function session_create_id(string $prefix = ""): string|false {}
```
williamdes, line 177:
```suggestion
function session_create_id(string $prefix = ''): string|false {}
```

session/session.php
```diff
@@ -238,26 +238,27 @@ function session_unset() {}
  * The garbage collector, this is executed when the session garbage collector
  * is executed and takes the max session lifetime as its only parameter.
  * </p>
- * @param callable $create_sid [optional]
+ * @param ?callable $create_sid [optional]
```
williamdes, line 241:
```suggestion
 * @param callable|null $create_sid [optional]
```

session/session.php
```diff
@@ -238,26 +238,27 @@ function session_unset() {}
...
  * </p>
- * @param callable $create_sid [optional]
+ * @param callable|null $create_sid [optional]
  * <p>This callback is executed when a new session ID is required.
  * No parameters are provided, and the return value should be a string that is a valid
  * session ID for your handler.</p>
- * @param callable $validate_sid [optional]
- * @param callable $update_timestamp [optional]
+ * @param ?callable $validate_sid [optional]
+ * @param ?callable $update_timestamp [optional]
```
williamdes, line 246:
```suggestion
 * @param callable|null $validate_sid [optional]
 * @param callable|null $update_timestamp [optional]
```

session/SessionHandler.php
```diff
@@ -29,9 +28,8 @@ public function close();
      * The return value (usually TRUE on success, FALSE on failure).
      * Note this value is returned internally to PHP for processing.
      * </p>
-     * @since 5.4
```
williamdes, line 32: I am not to much fan of removing the `since` tags, it always helps building old compat code

williamdes (approved):

Simbiat:
@williamdes, I hear you, but this was suggested in the ticket as potential solution. Code inspector now shows the same 2 errors, which we can't avoid, so need to wait for @DmitryTronin or someone else from his team to look at this closer.

DmitryTronin:
@Simbiat 
Sorry for the lack of feedback on this.
We took a look at this & failing tests once again and came up with another suggestion: to use `LanguageLevelTypeAware` in parameters definition. Here's an example of how you can use them: https://gist.github.com/DmitryTronin/3bdb3967b5fc6d1b70fd448ffc9dde40
I've used your revision 7d6658592f83324bd979628cdacefd623b175864 named "Removed functions from SessionHandler" to create this example. I've run our tests against the "fixed" SessionHandler class above & it went without failures.

A quick rundown on required changes:
- there shouldn't be return type hints (as explained [here](https://github.com/JetBrains/phpstorm-stubs/pull/1140#issuecomment-853065698)).
- `@since 5.4 tag` should be present, my assumption on [YouTrack](https://youtrack.jetbrains.com/issue/WI-60588#focus=Comments-27-4964766.0-0) was incorrect - if `@since` tag is missing, we assume there's a `@since 5.3` tag. That's an undocumented behavior.
- parameter type hints should be written using `#[LanguageLevelTypeAware()]` construcion.

Simbiat:
Restored `@since` and added `LanguageLevelTypeAware`. Not sure if correctly, though.

DmitryTronin:
We've adjusted the changes you proposed and pushed in https://github.com/JetBrains/phpstorm-stubs/pull/1190. Everything should be looking fine now.
I've also checked https://youtrack.jetbrains.com/issue/WI-60588 - the example shared there works fine now on the latest stubs version.

PR #1153, opened 2021-06-16
fixed wrong arg name for session_set_save_handler

chatii:
I got an error when using named arguments. (PHPStan and Psalm were working correctly)

ref: https://www.php.net/manual/en/function.session-set-save-handler.php

> session_set_save_handler(object $sessionhandler, bool $register_shutdown = true): bool

### DIFF
--- a/session/session.php
+++ b/session/session.php
@@ -252,7 +252,7 @@ function session_set_save_handler(callable $open, callable $close, callable $rea
  * (PHP 5.4)<br/>
  * Sets user-level session storage functions
  * @link https://php.net/manual/en/function.session-set-save-handler.php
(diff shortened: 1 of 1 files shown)


session/session.php
```diff
@@ -257,7 +257,7 @@ function session_set_save_handler(callable $open, callable $close, callable $rea
  * @param bool $register_shutdown [optional] Register session_write_close() as a register_shutdown_function() function.
  * @return bool true on success or false on failure.
  */
-function session_set_save_handler(SessionHandlerInterface $session_handler, $register_shutdown = true): bool {}
+function session_set_save_handler(SessionHandlerInterface $sessionhandler, $register_shutdown = true): bool {}
```
williamdes, line 260: You also need to edit the doc block `@param` :)

chatii:
@williamdes 
thanks, fixed it!

PR #1154, opened 2021-06-17
Fix \Ds\Set::__construct()

enumag:
The stubs are declared according to PHP [documentation](https://www.php.net/manual/en/ds-set.construct.php). However the documentation is bugged.

See my reports:
https://bugs.php.net/bug.php?id=81157
https://github.com/php-ds/ext-ds/issues/167

### DIFF
--- a/ds/ds.php
+++ b/ds/ds.php
@@ -266,7 +266,7 @@ namespace Ds;
 
         /**
          * Returns the result of adding all given values to the sequence.
-         * @param array|Traversable $values A traversable object or an array.
+         * @param iterable $values A traversable object or an array.
          * @return Sequence The result of adding all given values to the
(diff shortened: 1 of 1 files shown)


williamdes:
The `@param` should also be updated like in https://github.com/php-ds/polyfill/blob/5652e5870b20713ebe02182050df98aee5856d5b/src/Set.php#L33

enumag:
@williamdes You meant changing `array|Traversable` to `iterable`, right? There were more occurances of that in the file so I fixed them all.

williamdes:
> @williamdes You meant changing `array|Traversable` to `iterable`, right? There were more occurances of that in the file so I fixed them all.

Yup, thank you !

PR #1160, opened 2021-06-28
Avoid using multi-line annotations for PHP 7 compatibility

morozov:
While the package requires PHP 8 for development,
https://github.com/JetBrains/phpstorm-stubs/blob/a09a014a1984d2ac875c6b1220b7f17ca7d0a873/composer.json#L16-L17
It doesn't require PHP 8 when the package is used as a library.

Although, some code in the package is currently not compatible with PHP 7:
https://github.com/JetBrains/phpstorm-stubs/blob/a09a014a1984d2ac875c6b1220b7f17ca7d0a873/PDO/PDO.php#L1166-L1170

An attempt to use this code with PHP 7 causes a parse error:
```
php -l PDO/PDO.php

Parse error: syntax error, unexpected '0' (T_LNUMBER), expecting function (T_FUNCTION) or const (T_CONST) in PDO/PDO.php on line 1167
```

### DIFF
--- a/Core/Core.php
+++ b/Core/Core.php
@@ -1064,12 +1064,7 @@ function gc_disable(): void {}
  * </ul>
  * @since 7.3
  */
-#[ArrayShape([
-    "runs" => "int",
-    "collected" => "int",
-    "threshold" => "int",
-    "roots" => "int"
-])]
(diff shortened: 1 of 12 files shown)


williamdes (approved):
💯

PR #1166, opened 2021-07-01
Synchronized method return values with the official docs.

Mistralys:
Some methods can return `false` - this was not properly documented in the stub.

See related ticket: https://youtrack.jetbrains.com/issue/WI-61262

### DIFF
--- a/date/date_c.php
+++ b/date/date_c.php
@@ -790,7 +790,7 @@ class DateInterval
      * Sets up a DateInterval from the relative parts of the string
      * @param string $datetime
      * @return DateInterval|false Returns a new {@link https://www.php.net/manual/en/class.dateinterval.php DateInterval}
-     * instance on success, or false on failure.
+     * instance on success, or <b>FALSE</b> on failure.
      * @link https://php.net/manual/en/dateinterval.createfromdatestring.php
      */
     public static function createFromDateString($datetime) {}
--- a/date/date_c.php
+++ b/date/date_c.php
@@ -80,9 +80,9 @@ interface DateTimeInterface
      * @link https://secure.php.net/manual/en/datetime.diff.php
      * @param DateTimeInterface $targetObject <p>The date to compare to.</p>
      * @param bool $absolute <p>Should the interval be forced to be positive?</p>
-     * @return DateInterval|false
+     * @return DateInterval
      * The https://secure.php.net/manual/en/class.dateinterval.php DateInterval} object representing the
-     * difference between the two dates or <b>FALSE</b> on failure.
+     * difference between the two dates.
      */
     public function diff($targetObject, $absolute = false);
 
@@ -93,9 +93,11 @@ interface DateTimeInterface
      * @param string $format <p>
      * Format accepted by  {@link https://secure.php.net/manual/en/function.date.php date()}.
      * </p>
-     * @return string|false
+     * @return string
      * Returns the formatted date string on success or <b>FALSE</b> on failure.
+     * Since PHP8, it always returns <b>STRING</b>.
      */
+    #[LanguageLevelTypeAware(["8.0" => "string"], default: "string|false")]
     public function format($format);
 
     /**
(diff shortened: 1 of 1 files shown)


date/date_c.php
```diff
@@ -80,7 +80,7 @@ interface DateTimeInterface
      * @link https://secure.php.net/manual/en/datetime.diff.php
      * @param DateTimeInterface $targetObject <p>The date to compare to.</p>
      * @param bool $absolute <p>Should the interval be forced to be positive?</p>
-     * @return DateInterval
+     * @return DateInterval|false
```
isfedorov, line 83: According to https://www.php.net/manual/en/datetime.diff.php and https://github.com/php/php-src/blob/master/ext/date/php_date.stub.php#L132 return type is just `DateInterval`

date/date_c.php
```diff
@@ -93,15 +93,15 @@ public function diff($targetObject, $absolute = false);
      * @param string $format <p>
      * Format accepted by  {@link https://secure.php.net/manual/en/function.date.php date()}.
      * </p>
-     * @return string
+     * @return string|false
```
isfedorov, line 96: According to https://www.php.net/manual/en/datetime.format.php `false` could be returned only for PHP < 8. Please use `#LanguageLevelTypeAware` attribute to set return type in this case

date/date_c.php
```diff
@@ -93,15 +93,15 @@ public function diff($targetObject, $absolute = false);
...
      * Returns the formatted date string on success or <b>FALSE</b> on failure.
      */
     public function format($format);
 
     /**
      * (PHP 5 &gt;=5.5.0)<br/>
      * Returns the timezone offset
-     * @return int
+     * @return int|false
```
isfedorov, line 104: According to https://www.php.net/manual/en/datetime.getoffset.php `false` could be returned only for PHP < 8. Please use #LanguageLevelTypeAware attribute to set return type in this case

date/date_c.php
```diff
@@ -695,7 +695,7 @@ public static function listAbbreviations() {}
      * Returns a numerically indexed array with all timezone identifiers
      * @param int $timezoneGroup
      * @param string $countryCode
-     * @return array|false
+     * @return array
      * @link https://php.net/manual/en/datetimezone.listidentifiers.php
      */
     public static function listIdentifiers($timezoneGroup = DateTimeZone::ALL, $countryCode = null) {}
```
isfedorov, line 701: According to https://php.net/manual/en/datetimezone.listidentifiers.php `false` could be returned for PHP < 8. Please use #LanguageLevelTypeAware attribute to set return type in this case

isfedorov (changes requested):

date/date_c.php
Mistralys: That's strange - the official documentation says in the return value "or false on failure", but you're right, it is not mentioned anywhere else. I'll change it.

date/date_c.php
Mistralys: I am not familiar with the attribute. The few things I found online suggest it should be used like this (for a method that returns string, and string|false for PHP8):

```
/**
 * @return string
 */
#[LanguageLevelTypeAware(["8.0" => "string|false"], default: "string")]
public function format($format);
```

Is this correct?

isfedorov: Yes, that's correct. You can also find usages of `#LanguageLevelTypeAware` attribute in stubs, there are quite a lot of such usages.

Mistralys: Thanks for the info, I made the change.

date/date_c.php
Mistralys: Done. I edited the description to explain the return type for PHP8 as well.

date/date_c.php
Mistralys: Done. I edited the description to explain the return type for PHP8 as well.

Mistralys:
All changes requested have been pushed.

date/date_c.php
isfedorov: Please also add `#LanguageLevelTypeAware` to imports with `use JetBrains\PhpStorm\Internal\LanguageLevelTypeAware;` as currently class is marked as undefined in the code

PR #1167, opened 2021-07-05
Support for multiple php versions

isfedorov:
In scope of this PR tests were updated to run on different PHP versions (7.1 - 8.1). Stubs were updated to use LanguageLevelTypeAware and PhpStormElementAvailable attributes.


MaXal:
Let's change the execution of tests from the version under test to the one version (for example, 8). This will help us to keep only Model on 7.1 compatibility-level and evolve other parts (like parser and tests).

MaXal:
Can we move `dump-reflection-to-file.php` under `tests`?

tests/Parsers/PHPReflectionParser.php
```diff
@@ -15,40 +15,44 @@ class PHPReflectionParser
...
-        $jsonData = json_decode(file_get_contents(__DIR__ . '/../TestData/mutedProblems.json'));
-        $const_groups = get_defined_constants(true);
-        unset($const_groups['user']);
-        $const_groups = Utils::flattenArray($const_groups, true);
-        foreach ($const_groups as $name => $value) {
-            $constant = (new PHPDefineConstant())->readObjectFromReflection([$name, $value]);
-            $constant->readMutedProblems($jsonData->constants);
-            $stubs->addConstant($constant);
-        }
+        if (file_exists(__DIR__ . '/../../ReflectionData.json')) {
```
MaXal, line 18: If the file exists (for example, locally), it won't be regenerated on serialization. We can either show a message or better yet avoid this possible confusion by passing the argument that we need to reserealize.

tests/Model/PHPProperty.php
```diff
@@ -38,15 +42,17 @@ public function readObjectFromReflection($reflectionObject): static
         }
         $this->access = $access;
         $this->is_static = $reflectionObject->isStatic();
-        $this->typesFromSignature = self::getReflectionTypeAsArray($reflectionObject->getType());;
+        if (doubleval(getenv('PHP_VERSION') >= 7.4)) {
```
MaXal, line 45: Maybe something like will be better: `method_exists($reflectionObject, 'getType')`?

tests/Parsers/PHPReflectionParser.php
isfedorov: This check is done intentionally to avoid generating of `ReflectionData.json` locally if one just want to run tests with some PHP version and don't want to generate file to check other PHP versions. (some kind of fallback behavior). In case of a run on CI  `ReflectionData.json` will be always generated. I think that for local development it's easier than adding some additional argument - just run it as before.

uuf6429:
Hey, I was trying to run the tests and noticed how it works and that it can be improved a bit.
What confuses me, however, is how the base/root class `BaseStubsTest` contains PHP 7.1 return types (eg `: void`) and PHP 8 union types (eg: `: float|bool|int|string|null`)... so how are the tests actually running in these cases (eg PHP 7.0)?

A may be wrong, but I suspect that where there is `docker-compose -f docker-compose.yml run -e PHP_VERSION=${{matrix.php}} ...` is actually wrong because `-e` sets the env vars inside the running container, not for building the container (I've fallen in this trap before) and what happens is that it's actually always a php:latest instance.

isfedorov:
@uuf6429 
> What confuses me, however, is how the base/root class BaseStubsTest contains PHP 7.1 return types (eg : void) and PHP 8 union types (eg: : float|bool|int|string|null)... so how are the tests actually running in these cases (eg PHP 7.0)?

The process is next:
1) We build 2 services (actually 3 but the third one is for pecl), so the first service is `php_under_test` (this php version will be used as a test version to compare stubs with), the second `test_runner` (this php version will be used to run tests). `php_under_test` is being built with different php versions according to matrix, `test_runner` is being always built using latest stable php version (php:8.1-apache at the moment). Then we serialize all reflection data (builtin classes, functions etc) using `php_under_test` and then compare stubs with that serialized data, but tests themselves are being run using php from service `test_runner` so as far as tests are being run with php 8.1 `BaseTestClass` and all child classes can have features from php 8.1. The only classes that can't have features of latest php versions are classes that are used during serialization of reflection data (generally under `namespace StubTests\Model;`)

> I may be wrong, but I suspect that where there is docker-compose -f docker-compose.yml run -e PHP_VERSION=${{matrix.php}} ... is actually wrong because -e sets the env vars inside the running container, not for building the container

Probably this part can be removed. The environment variable is set in
```
env:
          PHP_VERSION: ${{matrix.php}}
```
But this should be double-checked.

isfedorov:
@uuf6429 btw,
> it can be improved a bit

If you see some possible ways for improvement please create a PR, we will be glad to review it and merge if everything will be fine

uuf6429:
Thanks for the explanation! It's much more clear how it works now.

> @uuf6429 btw,
> 
> > it can be improved a bit
> 
> If you see some possible ways for improvement please create a PR, we will be glad to review it and merge if everything will be fine

I think my idea of an improvement might not work in this case. The idea was to have two test suite ("all php tests" and "phpdoc tests") and then move the PHP version check inside the tests. For example, this:

```xml
        <testsuite name="PHP_7.0">
            <file>tests/BaseClassesTest.php</file>
            ...
```

becomes:
```xml
        <testsuite name="PHP">
            <directory suffix=".php">tests/</directory>
        </testsuite>
        ...
```
and then the test itself:
```php
/**
 * @requires PHP 7.0
 */
class BaseClassesTest extends AbstractBaseStubsTest
{
...
```

but this wouldn't work, since as per your explanation, it will always be running in PHP 8.1.

On some further though, instead, I think I can work with PHPUnit groups. I'll post my results later. Thanks again for your time!

PR #1169, opened 2021-07-11
Update fwrite return type to include 'false'

martin-juul:
Since 7.4.0 SplFileObject::fwrite returns false instead of 0 on failure

[PHP Manual Reference](https://www.php.net/manual/en/splfileobject.fwrite.php)

### DIFF
--- a/SPL/SPL_c1.php
+++ b/SPL/SPL_c1.php
@@ -779,7 +779,7 @@ class SplFileObject extends SplFileInfo implements RecursiveIterator, SeekableIt
          * the end of <i>string</i> is reached, whichever comes
          * first.
          * </p>
-         * @return int the number of bytes written, or 0 on error.
+         * @return int|bool the number of bytes written, or false on error.
          */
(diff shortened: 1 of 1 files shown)


voku:
Maybe "int|false" would be more correct.

SPL/SPL_c1.php
```diff
@@ -779,7 +779,7 @@ public function fscanf($format, &...$vars) {}
          * the end of <i>string</i> is reached, whichever comes
          * first.
          * </p>
-         * @return int the number of bytes written, or 0 on error.
+         * @return int|bool the number of bytes written, or false on error.
          */
```
isfedorov, line 783: @martin-juul Please use #[LanguageLevelTypeAware] attribute to set return types different for different PHP versions. You can find examples of usages of attribute across stubs.

isfedorov (changes requested):

isfedorov:
Fixed in a separate PR

PR #1173, opened 2021-07-15
Add Readonly attribute

icanhazstring:
This will add a `Readonly` attribute to the stubs.

I have created a small package to support a `readonly` functionality using static analysis (currently phpstan)
https://github.com/icanhazstring/phpstan-readonly-property

I introduced my own `Readonly` attribute inside the package. But it would be nice to have it in a more global package, so that other static analysis tools can use it without the need to require my library which requires phpstan.

Edit:
I created a PHPStorm Issue you can find it here: https://youtrack.jetbrains.com/issue/WI-61559

### DIFF
--- a/meta/attributes/Readonly.php
+++ b/meta/attributes/Readonly.php
@@ -1,0 +1,17 @@
+<?php
+
+namespace JetBrains\PhpStorm;
+
+use Attribute;
+
+/**
+ * This marks a property as readonly.
+ *
+ * That means that this property can only be writting from the declaring class constructor it self.
+ *
+ * @since 8.0
+ */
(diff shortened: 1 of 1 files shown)


icanhazstring:
Hi @isfedorov. Sure thing, didn't know about that. Will do asap. Thanks 👍

icanhazstring:
@isfedorov I think also, that the attribute could be added, so open source static analysis could take use of it. Support for the IDE is an extra that could come with a later update.

isfedorov:
The feature was declined after discussion. Details in https://youtrack.jetbrains.com/issue/WI-61559

PR #1176, opened 2021-07-20
mark rand_* functions as $mayDependOnGlobalScope

isfedorov:
Functions are actually Pure in terms of PhpStorm (don't have side effects) but result of function calls is different. For such cases Pure attribute has parameter $mayDependOnGlobalScope that indicates whether the function result may be dependent on anything except passed variables

### DIFF
--- a/standard/standard_2.php
+++ b/standard/standard_2.php
@@ -927,6 +927,7 @@ function proc_nice(int $priority): bool {}
  * @return int A pseudo random value between min
  * (or 0) and max (or getrandmax, inclusive).
  */
+#[Pure(true)]
 function rand(int $min = 0, int $max): int {}
 
 /**
@@ -965,6 +966,7 @@ function getrandmax(): int {}
  * @return int A random integer value between min (or 0)
  * and max (or mt_getrandmax, inclusive)
  */
+#[Pure(true)]
 function mt_rand(int $min = 0, int $max): int {}
 
 /**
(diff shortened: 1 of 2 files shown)


jlherren:
Can I ask for clarifications on this? These functions clearly *do* have side-effects: they modify the state of the random number generator. Their return value does indeed depend on the global scope, but they also *modify* the global scope, which makes them impure. So adding `#[Pure(true)]` seems wrong to me. In particular, the PhpDoc of `Pure` states:

> This means that a function call that resolves to such a function can be safely removed if the execution result is not used in code afterwards.

This is not the case for these functions, consider this code:

```
<?php
mt_srand(42);
mt_rand(); // cannot remove this line, even if return value is unused
echo mt_rand(), PHP_EOL;
```

This code has deterministic output and always prints the same value when run.  However, removing the first `mt_rand()` cannot be safely removed, as it changes what the code outputs.

Perhaps the original intention was to implement a warning when the return value of such functions is discarded. In that case it would make sense to have a separate attribute, similar to C++17's `[[nodiscard]]` attribute.

isfedorov:
@jlherren If I'm not mistaken in your example function `mt_srand` is the one that changes the global state and `mt_rand` only returns a random value. For example, change the call to 
```
<?php
mt_srand(42);
echo mt_rand(), PHP_EOL;
echo mt_rand(), PHP_EOL;
echo mt_rand(), PHP_EOL;
```
And you will get three different numbers in the calls of `mt_rand`. (yes, they will be the same for all calls but still, those are three different numbers. At the same time if you remove `mt_srand(42);` then you will get unique numbers for every call every time.
Please correct me if I'm wrong

PR #1177, opened 2021-07-20
Use strict typing for Redis zRemRangeByScore

jspaetzel:
When `declare(strict_types=1);` is set you cant pass a float to zRemRangeByScore without triggering a type error. So I think it's better to document this as such.


jspaetzel:
Example issue this confusion has caused, https://github.com/phpredis/phpredis/issues/1322

PR #1187, opened 2021-07-29
couchbase: make v3 stubs default

avsej:
Renamed directories:

* `couchbase` -> `couchbase_v2`
* `couchbase_v3` -> `couchbase`


isfedorov (changes requested):
Tests failing now due to outdated stubs map. @avsej  Could you please call `generate-stubs-map.php` and commit updated map

isfedorov:
Fixed in #1194

PR #1189, opened 2021-07-29
Update standard_2.php

tutosfaciles48:
refactor(rand): the function require 0 or 2 parameters, according to the doc, the $min is not optional

### DIFF
--- a/standard/standard_2.php
+++ b/standard/standard_2.php
@@ -931,10 +931,10 @@ function proc_nice(int $priority): bool {}
  * @param int $min [optional]
  * @param int $max [optional]
  * @return int A pseudo random value between min
- * (or 0) and max (or getrandmax, inclusive).
+ *  and max (or getrandmax, inclusive).
  */
 #[Pure(true)]
-function rand(int $min = 0, int $max): int {}
+function rand(int $min, int $max): int {}
 
 /**
  * Seed the random number generator


standard/standard_2.php
```diff
@@ -931,10 +931,10 @@ function proc_nice(int $priority): bool {}
  * @param int $min [optional]
  * @param int $max [optional]
  * @return int A pseudo random value between min
- * (or 0) and max (or getrandmax, inclusive).
+ *  and max (or getrandmax, inclusive).
  */
 #[Pure(true)]
-function rand(int $min = 0, int $max): int {}
```
isfedorov, line 937: I'm sorry, but according to reflection function has 2 parameters and both are optional. Please see https://3v4l.org/WuoDD . Also according to docs https://www.php.net/manual/en/function.rand.php `If called without the optional min, max arguments rand() returns a pseudo-random integer between 0 and getrandmax()` Could you please elaborate on this?

isfedorov (changes requested):

standard/standard_2.php
voku: The parameters are both optional, until one parameter is used, then both parameters are needed. 🤔

isfedorov: Thank you @voku! Unfortunately, it's not possible at the moment to declare such logic in stubs. I've created a feature request https://youtrack.jetbrains.com/issue/WI-61718 to be able to process such cases.

isfedorov:
The issue was fixed on PhpStorm's side WI-61718

PR #1205, opened 2021-08-16
Flesh out enums further

Crell:
1. Enums now use the 8.1 native "readonly" flag for their exposed properties, rather than custom logic.
2. The static return seems to be missing from from/tryFrom in several cases for some reason, so I added those.


Crell:
Hm, interesting.  It seems like the inspector doesn't handle PHP 8.1 code, but it's not choking on "enum" itself for some reason.

DmitryTronin:
Thanks for the PR.
We would need to wait till php-cs-fixer also supports this. If we merge this right now, all consecutive PRs would keep failing on these changes.
As for Qodana - it is being launched on a "stable" version that doesn't support `readonly` properties yet. But that one we can safely ignore atm.

Crell:
Gotcha.  I leave it in your capable hands then. :smile:

DmitryTronin:
@Crell Could you please update & push your changes once again? 
We'll run the tests against them using the php-cs-fixer dev version that support readonly properties: https://github.com/JetBrains/phpstorm-stubs/pull/1231

Crell:
I just rebased through the GitHub UI, which triggered a new workflow run. Let's see how it does.

DmitryTronin:
It is unfortunately still failing on PHP CS Fixer inspection (dev version) when run on 8.0 environment. To workaround the issue I've created a separate [PR](https://github.com/JetBrains/phpstorm-stubs/pull/1240). It's currently failing because of the another bug in php-cs-fixer but I hope they'll address it soon & we'll continue with this PR.

isfedorov:
@Crell I've temporarily disabled cs-fixer check and added 'readonly' to enums in #1278 so closing now this PR as obsolete. Thank you for your contribution!

PR #1206, opened 2021-08-18
Support for property readonly

isfedorov:


DmitryTronin:
QD failure on line 22 in _tests/StubsPhp81Tests.php: Undefined property, Property 'isReadonly' is undefined_ happens only on 2020.3 PHP plugin version. On latest versions this doesn't happen, hence we can ignore it.

PR #1230, opened 2021-09-09
Fix return types for SessionHandlerInterface

stchr:
See https://www.php.net/manual/en/sessionhandlerinterface.gc.php and https://www.php.net/manual/en/sessionhandlerinterface.read.php

### DIFF
--- a/session/SessionHandler.php
+++ b/session/SessionHandler.php
@@ -43,8 +43,8 @@ interface SessionHandlerInterface
      * Sessions that have not updated for
      * the last maxlifetime seconds will be removed.
      * </p>
-     * @return int|bool <p>
-     * The return value (usually TRUE on success, FALSE on failure).
+     * @return int|false <p>
+     * Returns the number of deleted sessions on success, or false on failure.
      * Note this value is returned internally to PHP for processing.
      * </p>
      * @since 5.4
--- a/session/SessionHandler.php
+++ b/session/SessionHandler.php
@@ -43,8 +43,8 @@ interface SessionHandlerInterface
      * Sessions that have not updated for
      * the last maxlifetime seconds will be removed.
      * </p>
-     * @return int|false <p>
-     * Returns the number of deleted sessions on success, or false on failure.
+     * @return #[LanguageLevelTypeAware(['7.1' => 'int|false'], default: 'bool') <p>
+     * #[LanguageLevelTypeAware(['7.1' => 'Returns the number of deleted sessions on success, or false on failure.'], default: 'The return value (usually TRUE on success, FALSE on failure).')
      * Note this value is returned internally to PHP for processing.
      * </p>
      * @since 5.4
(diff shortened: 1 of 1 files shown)


session/SessionHandler.php
```diff
@@ -43,8 +43,8 @@ public function destroy(#[LanguageLevelTypeAware(['8.0' => 'string'], default: '
      * Sessions that have not updated for
      * the last maxlifetime seconds will be removed.
      * </p>
-     * @return int|bool <p>
-     * The return value (usually TRUE on success, FALSE on failure).
+     * @return #[LanguageLevelTypeAware(['7.1' => 'int|false'], default: 'bool') <p>
```
DmitryTronin, line 46: You don't need to use #[LanguageLevelTypeAware] constructions in @return phpdoc annotations. See https://youtrack.jetbrains.com/issue/WI-57040: *If attribute is encountered PS will ignore any type in type declaration/phpdoc and will get type based on map from attribute and current language level.*
This can be left in human-readable form, or like in php-src: https://github.com/php/php-src/blob/29642623fdd1a51126e0a79311ec4ce87fd263b7/ext/session/session.stub.php#L119

Instead you can add `#[LanguageLevelTypeAware(['7.1' => 'int|false'], default: 'bool')]` right above the method definition and that would work well instead.

DmitryTronin (changes requested):

session/SessionHandler.php
stchr: Sure, I got this. 
But there was a change in the return value from bool to int|false in PHP 7.1, See Changelog in https://www.php.net/manual/en/sessionhandlerinterface.gc.php. This is what I wanted to express. 

Would be glad if there is a more elegant solution.

DmitryTronin: The `#[LanguageLevelTypeAware(['7.1' => 'int|false'], default: 'bool')]` notation above the method would reflect the changes for the static analyzer. When IDE detects `LanguageLevelTypeAware` annotation for the method - it ignores the PHPDoc `@return` section: that's why there's no point to include `LanguageLevelTypeAware` also in `@return`  section.
If you want this to be reflected in a method documentation, then `LanguageLevelTypeAware` doesn't really look human-readable. I believe you can include this information in a multiline description following the `@return` tag.
The full method signature would then look like this:

```
    /**
     * Cleanup old sessions
     * @link https://php.net/manual/en/sessionhandlerinterface.gc.php
     * @param int $max_lifetime <p>
     * Sessions that have not updated for
     * the last maxlifetime seconds will be removed.
     * </p>
     * @return int|false <p>
     * The return value (usually TRUE on success, FALSE on failure).
     * Prior to PHP version 7.1, the function returned true on success.
     * Note this value is returned internally to PHP for processing.
     * </p>
     * @since 5.4
     */
    #[LanguageLevelTypeAware(['7.1' => 'int|false'], default: 'bool')]
    public function gc(#[LanguageLevelTypeAware(['8.0' => 'int'], default: '')] $max_lifetime);
```

DmitryTronin (approved):

PR #1232, opened 2021-09-16
Remove & in phpdoc

VincentLanglet:
Some tools like phpstan/phpdoc-parser doesn't support `@param int &$consumed` syntax.
By removing the `&`, it would allow phpstan to better understand stubs provided by phpstorm.

Also it's useless as the & is already present in the native signature and it's not in the standard either: https://docs.phpdoc.org/latest/guide/references/phpdoc/tags/param.html

Ref: https://github.com/phpstan/phpstan/issues/5636#issuecomment-920858119


DmitryTronin:
I dig this a bit and found why the `&`'s are there. 
They was added a year ago: https://github.com/JetBrains/phpstorm-stubs/pull/892. While working on this, a PR to support `&` in PHPDocs was created & accepted: https://github.com/phpDocumentor/ReflectionDocBlock/pull/254 so, technically, PHPDocs are supporting this.
I would appreciate if @voku could comment here describing why the `&$` syntax in PHPDocs isn't "useless". Unfortunately I can't recall any of these usages right now.

VincentLanglet:
For the record, I added the `&$` syntax to the phpstan/parser https://github.com/phpstan/phpdoc-parser/pull/83 so I don't know if this PR is still necessary (it should at least not be needed for phpstan).

DmitryTronin:
Thanks for the update. We'll leave our stubs unchanged then.

PR #1245, opened 2021-10-05
Update Core_c.php

WinterSilence:
key can be string or integer

### DIFF
--- a/Core/Core_c.php
+++ b/Core/Core_c.php
@@ -71,7 +71,7 @@ interface Iterator extends Traversable
     /**
      * Return the key of the current element
      * @link https://php.net/manual/en/iterator.key.php
-     * @return string|float|int|bool|null scalar on success, or null on failure.
+     * @return string|int|null scalar on success, or null on failure.
      */
     #[TentativeType]
     public function key(): mixed;


isfedorov:
Can't agree https://3v4l.org/PTJBY https://3v4l.org/rcpGo

WinterSilence:
@isfedorov it's must be compatible with `key()`

WinterSilence:
@isfedorov https://www.php.net/manual/ru/language.types.array.php#example-54

> The key can either be an int or a string.

WinterSilence:
@isfedorov that iterator? lol
~~~php
private $array = array(
    true => "firstelement",
    false => "secondelement",
);
~~~

WinterSilence:
@isfedorov I've done it a long time ago, for example before 5.5 key can be object... 

focus on this:
> Additionally the following key casts will occur:
    Strings containing valid decimal ints, unless the number is preceded by a + sign, will be cast to the int type. E.g. the key "8" will actually be stored under 8. On the other hand "08" will not be cast, as it isn't a valid decimal integer.
    Floats are also cast to ints, which means that the fractional part will be truncated. E.g. the key 8.7 will actually be stored under 8.
    Bools are cast to ints, too, i.e. the key true will actually be stored under 1 and the key false under 0.
    Null will be cast to the empty string, i.e. the key null will actually be stored under "".
    Arrays and objects can not be used as keys. Doing so will result in a warning: Illegal offset type.

PR #1254, opened 2021-10-15
Update standard_4.php

WinterSilence:


standard/standard_4.php
```diff
@@ -181,13 +181,16 @@ function unserialize(string $data, array $options = []): mixed {}
 /**
  * Dumps information about a variable
  * @link https://php.net/manual/en/function.var-dump.php
- * @param mixed $value <p>
- * The variable you want to export.
- * </p>
+ * @param mixed $value The variable you want to export.
  * @param mixed ...$values [optional]
+ * @param mixed ...$vars [optional]
```
DmitryTronin, line 186: Why `...$vars`? Can't see it anywhere:
- https://www.php.net/manual/en/function.var-dump.php 
- https://github.com/php/php-src/blob/master/ext/standard/basic_functions.stub.php#L1907
- https://github.com/php/php-src/blob/master/ext/standard/var.c#L217

DmitryTronin:
I suggest changing the method signature in the following way:
```
/**
 * Dumps information about a variable
 * @link https://php.net/manual/en/function.var-dump.php
 * <p>
 * @param mixed $value The variable you want to export.
 * @param mixed ...$values [optional]
 * @param mixed ...$vars Used instead $value and ...$values before 8.0
 * </p>
 * @return void
 */
function var_dump(
	#[PhpStormStubsElementAvailable(from: '5.4', to: '7.4')] mixed ...$vars,
	#[PhpStormStubsElementAvailable(from: '8.0')] mixed $value,
	#[PhpStormStubsElementAvailable(from: '8.0')] mixed ...$values): void {}
```

That way in IDE ver. 2021.3 both for PHP 8.0 & earlier parameter hint popup shows correct suggestions:
- php8.0: https://i.gyazo.com/3edf999282db57c150fa2cc58b17d7bb.mp4
- php7.4: https://gyazo.com/a7715018f34bd139938f7919e58637cf

I've run the test locally & it looked fine.

WinterSilence:
@DmitryTronin должно быть `#[PhpStormStubsElementAvailable(from: '5.3', to: '7.4')] mixed ...$values,`

Остается проблема с PhpDoc: такие инструменты как Psalm и PhpStan во многом полагаются на PhpDoc и подобные избыточные комментарии могут вызвать проблемы.

DmitryTronin:
We've implemented the proposed changes with small corrections in  [1310](https://github.com/JetBrains/phpstorm-stubs/pull/1310), thank you.

PR #1266, opened 2021-10-30
optimize phpdoc + small clean-up

voku:


FFI/FFI.php
```diff
@@ -157,7 +157,7 @@ public static function typeof(CData $ptr): CType {}
          * dimensions specified by $dimensions.
          *
          * @param CType $type
-         * @param array|int[] $dimensions
+         * @param int[] $dimensions
```
dereuromark, line 160: is it too early to use generics? `array<int>` etc?
I did run into https://youtrack.jetbrains.com/issue/WI-63759, so I guess yeah.

standard/standard_9.php
```diff
@@ -444,11 +444,11 @@ function array_uintersect_uassoc(
...
  * Computes the difference of arrays
  * @link https://php.net/manual/en/function.array-diff.php
- * @param array $array <p>
+ * @param string[]|Stringable[] $array <p>
  * The array to compare from
  * </p>
- * @param array ...$arrays
- * @return array an array containing all the entries from
+ * @param string[]|Stringable[] ...$arrays
+ * @return string[]|Stringable[] an array containing all the values and keys from
```
dereuromark, line 451: Stringable might not be defined in PHP7.x code. Is that a problem?

standard/standard_9.php
```diff
@@ -444,11 +444,11 @@ function array_uintersect_uassoc(
 /**
  * Computes the difference of arrays
  * @link https://php.net/manual/en/function.array-diff.php
- * @param array $array <p>
+ * @param string[]|Stringable[] $array <p>
  * The array to compare from
  * </p>
- * @param array ...$arrays
- * @return array an array containing all the entries from
+ * @param string[]|Stringable[] ...$arrays
```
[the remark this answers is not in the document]
voku: https://twitter.com/suckup_de/status/1455140344585244673?s=20

That was the question. But I think if 50% of the php developer (from the developers that try to answer php questions in there free time) do not expect that behaviour we should at least give some hints, or? 🤔

standard/standard_1.php
```diff
@@ -996,7 +996,7 @@ function explode(string $separator, string $string, int $limit) {}
  * implode as glue would be
  * the second parameter and thus, the bad prototype would be used.
  * </p>
- * @param array|null $array <p>
+ * @param float[]|int[]|string[]|Stringable[]|null $array <p>
```
dereuromark, line 999: Shouldnt it be FQCN in docblocks?
`...|\Stringable[]|...` etc?

FFI/FFI.php
staabm: `array<int>` has no advantages over `int[]` - and as you already menioned... its less supported in tooling

voku:
Is there anything in the pull request that can be used, otherwise I would close it?

DmitryTronin:
We can merge all the changes that are not commented or questioned. You can revert them, update the project & push for a fresh workflow run.

voku:
I think most changes are already fixed... so I will close the PR.

PR #1275, opened 2021-11-13
fix unserialize() phpdoc-return-type

staabm:
refs https://github.com/phpstan/phpstan/issues/5997

### DIFF
--- a/SPL/SPL.php
+++ b/SPL/SPL.php
@@ -1752,7 +1752,7 @@ class ArrayObject implements IteratorAggregate, ArrayAccess, Serializable, Count
      * @param string $data <p>
      * The serialized <b>ArrayObject</b>.
      * </p>
-     * @return void The unserialized <b>ArrayObject</b>.
+     * @return void
      */
     #[TentativeType]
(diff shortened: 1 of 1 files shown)


SPL/SPL.php
```diff
@@ -1752,7 +1752,7 @@ public function natcasesort(): bool {}
      * @param string $data <p>
      * The serialized <b>ArrayObject</b>.
      * </p>
-     * @return void The unserialized <b>ArrayObject</b>.
```
DmitryTronin, line 1755: Could you please clarify why you want to remove this comment? It exists at https://www.php.net/manual/en/arrayobject.unserialize.php

staabm: a comment behind `return typeX` describes what the method returns. since this method doesn't return anything - aka `void` - the description which reads like a `unserialized ArrayObject` would be returned is wrong.

DmitryTronin: Thank you. I've also reported this as https://github.com/php/doc-en/pull/1101

staabm: > Thank you. I've also reported this as [php/doc-en#1101](https://github.com/php/doc-en/pull/1101)

good catch, thank you!

PR #1282, opened 2021-11-17
Improve mysqli stubs

kamil-tekiela:
Improve mysqli_fetch_* stubs.
Adjusted the stubs to the changes I made in the PHP manual over the past year.

I double-checked my changes, but please check again to make sure I didn't make any mistakes.

### DIFF
--- a/mysqli/mysqli.php
+++ b/mysqli/mysqli.php
@@ -235,9 +235,9 @@ class mysqli
     public function change_user(string $username, string $password, ?string $database): bool {}
 
     /**
-     * Returns the default character set for the database connection
+     * Returns the current character set of the database connection
      * @link https://php.net/manual/en/mysqli.character-set-name.php
-     * @return string The default character set for the current connection
+     * @return string The current character set of the connection
      */
     #[TentativeType]
     public function character_set_name(): string {}
@@ -263,7 +263,7 @@ class mysqli
      * @return bool true on success or false on failure.
      */
     #[TentativeType]
-    public function commit(int $flags = -1, ?string $name = null): bool {}
+    public function commit(int $flags = 0, ?string $name = null): bool {}
 
     /**
      * @link https://php.net/manual/en/function.mysqli-connect.php
@@ -336,13 +336,13 @@ class mysqli
     /**
      * Returns statistics about the client connection
      * @link https://php.net/manual/en/mysqli.get-connection-stats.php
-     * @return array|false an array with connection stats if success, false otherwise.
+     * @return array an array with connection stats.
      */
     #[TentativeType]
     public function get_connection_stats(): array {}
 
     /**
-     * An undocumented function equivalent to the $server_info property
+     * Returns the version of the MySQL server
      * @link https://php.net/manual/en/mysqli.get-server-info.php
      * @return string A character string representing the server version.
      */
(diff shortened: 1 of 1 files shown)


mysqli/mysqli.php
```diff
@@ -358,9 +358,9 @@ public function get_server_info(): string {}
     public function get_warnings(): mysqli_warning|false {}
 
     /**
-     * Initializes MySQLi and returns a resource for use with mysqli_real_connect()
+     * Initializes MySQLi object
      * @link https://php.net/manual/en/mysqli.init.php
-     * @return mysqli an object.
+     * @return void
```
DmitryTronin, line 363: Could you please share your arguments behind `void`? php-src [has](https://github.com/php/php-src/blob/master/ext/mysqli/mysqli.stub.php#L165) this hinted as `bool|null`

mysqli/mysqli.php
```diff
@@ -918,7 +930,7 @@ class mysqli_result implements IteratorAggregate
     #[LanguageLevelTypeAware(['8.1' => 'int'], default: '')]
     public $field_count;
     /**
-     * @var array
+     * @var int[]|null
```
DmitryTronin, line 933: Shouldn't be just `?array`? That's a more safe way of hinting this. See also https://github.com/php/php-src/blob/master/ext/mysqli/mysqli.stub.php#L359

mysqli/mysqli.php
```diff
@@ -2109,6 +2125,7 @@ function mysqli_kill(mysqli $mysql, int $process_id): bool {}
  * @link https://php.net/manual/en/mysqli.set-local-infile-default.php
  * @param mysqli $mysql A link identifier returned by mysqli_connect() or mysqli_init()
  * @return void
+ * @removed 5.5
```
DmitryTronin, line 2128: for the reference: https://github.com/php/php-src/commit/522595086b8d654a2fd954977f7f443f2578de22

williamdes (approved):

mysqli/mysqli.php
kamil-tekiela: I can simply be `?array` but I just copied it over from the procedural variant https://github.com/kamil-tekiela/phpstorm-stubs/blob/a8c609c8ad9e56e34967252d024b36aec13dc401/mysqli/mysqli.php#L1672
Which one is better? I think it should be one or the other in both circumstances.

mysqli/mysqli.php
kamil-tekiela: My bad. The correct signature is [`?false`](https://www.php.net/manual/en/mysqli.init.php), but I guess it should be `bool|null` in here as well, correct?

DmitryTronin: Yes, thanks.

mysqli/mysqli.php
DmitryTronin: I believe this one should be left as `?array` or `array|null`, because procedural `mysqli_fetch_lengths()` is described separately & `$lengths` is more of an OOP context.

mysqli/mysqli.php
```diff
@@ -918,7 +930,7 @@ class mysqli_result implements IteratorAggregate
     #[LanguageLevelTypeAware(['8.1' => 'int'], default: '')]
     public $field_count;
     /**
-     * @var array
+     * @var ?array
```
williamdes, line 933:
```suggestion
     * @var array|null
```

PR #1288, opened 2021-11-23
Fix minor typos in sone methods

niladam:
This file had a small typo where instead of `key`, it was `kye`. This PR fixes this without any other modifications.


niladam:
hey @isfedorov - since this is my first contribution apparently there's a need for an approval.

This is just a minor typo fix so it could be a quick-win :)

DmitryTronin:
These aren't typos: https://github.com/php-gnupg/php-gnupg/blob/master/gnupg.c#L252 :)
You might want to submit a PR into the extension repository instead.

PR #1289, opened 2021-11-25
Add tests on php 7.0

isfedorov:
Now stubs tests can be run against PHP 7.0.


Phar/Phar.php
```diff
@@ -73,7 +73,10 @@ public function __destruct() {}
      * @return void no return value, exception is thrown on failure.
      */
     #[TentativeType]
-    public function addEmptyDir(#[LanguageLevelTypeAware(['8.0' => 'string'], default: '')] $directory): void {}
+    public function addEmptyDir(
+        #[PhpStormStubsElementAvailable(from: '5.3', to: '7.0')] $directory = '',
```
DmitryTronin, line 77: Sorry, is there a default value for `$directory`? I've checked that with https://3v4l.org/9fb0m and looks like there isn't.
Based on this, it looks like parameter is only available in version 7.3-7.4 and > 8.0.

The signature looks like this for me:
```
    public function addEmptyDir(
        #[PhpStormStubsElementAvailable(from: '7.3', to: '7.4')] $dirname,
        #[PhpStormStubsElementAvailable(from: '8.0')] #[LanguageLevelTypeAware(['8.0' => 'string'], default: '')] $directory
    )
```
Let me know if I'm reading that wrong

DmitryTronin: Edit: please disregard, I've run it only on supported versions. Will double-check

Phar/Phar.php
```diff
@@ -109,7 +112,8 @@ public function addFile(
     #[TentativeType]
     public function addFromString(
         #[LanguageLevelTypeAware(['8.0' => 'string'], default: '')] $localName,
-        #[LanguageLevelTypeAware(['8.0' => 'string'], default: '')] $contents
+        #[PhpStormStubsElementAvailable(from: '5.3', to: '7.0')] $contents = '',
```
DmitryTronin, line 115: Looks like `$contents = '',` doesn't actually have a default value: https://3v4l.org/fb4Fq

Phar/Phar.php
DmitryTronin: So, taking previous 3v4l result in mind, the signature would look like
```
    public function addEmptyDir(
        #[PhpStormStubsElementAvailable(from: '5.3', to: '7.4')] #[LanguageLevelTypeAware(['7.4' => 'string'], default: '')] $dirname,
        #[PhpStormStubsElementAvailable(from: '8.0')] string $directory
    )
```
instead?

date/date.php
```diff
@@ -469,7 +469,7 @@ function mktime(
     ?int $day = null,
     ?int $year = null,
     #[Deprecated('Use the new timezone handling functions instead', since: '5.3')]
-    #[PhpStormStubsElementAvailable(from: '5.3', to: '7.0')] $is_dst = -1
+    #[PhpStormStubsElementAvailable(from: '5.3', to: '5.6')] $is_dst = -1
```
DmitryTronin, line 472: Optional `$is_dst` was available only for a short period: https://3v4l.org/u4D30#v5523 - these are latest PHP 5.5 & 5.6 releases.
I'd change this to `(from: '5.5', to: '5.6')`.

date/date.php
```diff
@@ -508,7 +508,7 @@ function gmmktime(
     ?int $month = null,
     ?int $day = null,
     ?int $year = null,
-    #[PhpStormStubsElementAvailable(from: '5.3', to: '7.0')] $is_dst = null
+    #[PhpStormStubsElementAvailable(from: '5.3', to: '5.6')] $is_dst = null
```
DmitryTronin, line 511: The same as above: https://3v4l.org/nqTZf#v5523, I'd say it's `(from: '5.5', to: '5.6')`

dom/dom_c.php
```diff
@@ -1312,7 +1312,10 @@ public function removeNamedItem($name) {}
      * in this map).
      */
     #[TentativeType]
-    public function item(#[LanguageLevelTypeAware(['8.0' => 'int'], default: '')] $index): ?DOMNode {}
+    public function item(
+        #[PhpStormStubsElementAvailable(from: '5.3', to: '7.0')] $index = 0,
```
DmitryTronin, line 1316: No default value for $index? See https://3v4l.org/kYiPo

dom/dom_c.php
```diff
@@ -1328,8 +1331,10 @@ public function item(#[LanguageLevelTypeAware(['8.0' => 'int'], default: '')] $i
      */
     #[TentativeType]
     public function getNamedItemNS(
-        #[LanguageLevelTypeAware(['8.0' => 'string|null'], default: '')] $namespace,
-        #[LanguageLevelTypeAware(['8.0' => 'string'], default: '')] $localName
+        #[PhpStormStubsElementAvailable(from: '5.3', to: '7.0')] $namespaceURI = '',
```
DmitryTronin, line 1334: `$namespaceURI` is available till 7.4: https://3v4l.org/mRrva#v700, without default value.
`$namespace` is available only since 8.0.
`$localName` doesn't have a default value?: https://3v4l.org/em9Kr#v540

sqlite3/sqlite3.php
```diff
@@ -256,7 +256,7 @@ public function createFunction(
         #[LanguageLevelTypeAware(['8.0' => 'string'], default: '')] $name,
         #[LanguageLevelTypeAware(['8.0' => 'callable'], default: '')] $callback,
         #[LanguageLevelTypeAware(['8.0' => 'int'], default: '')] $argCount = -1,
-        int $flags = 0
+        #[PhpStormStubsElementAvailable(from: '7.1')] int $flags = 0
```
DmitryTronin, line 259: Default value for `$flags` is available only starting PHP 8.0.

standard/standard_0.php
```diff
@@ -1386,9 +1387,13 @@ function strcspn(string $string, string $characters, int $offset, ?int $length):
...
+ * @param string|null $token <p>
  * The delimiter used when splitting up str.
  * </p>
  * @return string|false A string token.
  */
-function strtok(string $string, ?string $token): string|false {}
+function strtok(
+    string $string,
+    #[PhpStormStubsElementAvailable(from: '5.3', to: '7.0')] $token,
+    #[PhpStormStubsElementAvailable(from: '7.1')] ?string $token = null
```
DmitryTronin, line 1398: 1) `$token` is nullable only starting 8.0?: https://3v4l.org/EBQM8#v800
2) From 5.4 till 7.4 `$string` was called `$str`. Not a big deal I guess?..

standard/standard_1.php
```diff
@@ -418,7 +418,11 @@ function str_split(string $string, int $length = 1): array|false {}
  * not found.
  */
 #[Pure]
-function strpbrk(string $string, string $characters): string|false {}
+function strpbrk(
```
DmitryTronin, line 421: https://3v4l.org/RW51p#v540:
1) till 7.4 `$string` called `$haystack` 
2) `$char_list` had no default value

standard/standard_2.php
```diff
@@ -949,7 +949,10 @@ function rand(int $min = 0, int $max): int {}
  * </p>
  * @return void
  */
-function srand(int $seed, int $mode = MT_RAND_MT19937): void {}
+function srand(
+    int $seed,
+    #[PhpStormStubsElementAvailable(from: '7.1')] int $mode = MT_RAND_MT19937
```
DmitryTronin, line 954: Even though it [says](https://www.php.net/manual/en/function.srand.php) `$mode = MT_RAND_MT19937`, it returns `0` as a default value: https://3v4l.org/t9nVj#v800
Looks like a documentation issue?
Also: `int` parameter type appears since 8.0 only.

standard/standard_2.php
```diff
@@ -984,7 +987,10 @@ function mt_rand(int $min = 0, int $max): int {}
  * </p>
  * @return void
  */
-function mt_srand(int $seed, int $mode = MT_RAND_MT19937): void {}
+function mt_srand(
```
DmitryTronin, line 990: same as for `srand()`

Phar/Phar.php
isfedorov: Yeah, according to reflection this parameter doesn't have a default value, but still, it's optional and unfortunately, in terms of stubs, we don't have any other mechanism to say that parameter is optional other than adding default value or `[optional]` to PhpDoc. The latter wouldn't work since it would be applied for all PHP versions, so the default value is the only way here.

dom/dom_c.php
isfedorov: The same case as above. Default value require to mark the parameter as optional

standard/standard_0.php
isfedorov: 1. Actually no. https://3v4l.org/ZTWUb `allowsNull` actually shows that since PHP 8 parameter has nullable typehint `?string`
2. Yep. We care about names only for PHP 8 and later

standard/standard_2.php
isfedorov: Reflection returns numeric value of `define("MT_RAND_MT19937", 0);`
`int` as a typehint appears since 8.0 because generally typehints were introduced in reflection in php 8.0. As far as this is `function` but not `method` it can't be overridden and thus there will be no problems with hierarchy checks so typehint here is ok.

standard/standard_2.php
isfedorov: See above

standard/standard_1.php
isfedorov: 1. we don't care about param names prior 8.0 so it's fine
2. yep, but still it was optional so we have to put default value to make it optional

Phar/Phar.php
isfedorov: Updated to
```
#[TentativeType]
    public function addEmptyDir(
        #[PhpStormStubsElementAvailable(from: '5.3', to: '7.4')] $directory = '',
        #[PhpStormStubsElementAvailable(from: '8.0')] string $directory
    ): void {}
```

dom/dom_c.php
isfedorov: Updated to
```
#[TentativeType]
    public function getNamedItemNS(
        #[PhpStormStubsElementAvailable(from: '5.3', to: '7.4')] $namespaceURI = '',
        #[PhpStormStubsElementAvailable(from: '8.0')] ?string $namespace,
        #[PhpStormStubsElementAvailable(from: '5.3', to: '7.4')] $localName = '',
        #[PhpStormStubsElementAvailable(from: '8.0')] string $localName
    ): ?DOMNode {}
```

date/date.php
isfedorov: Right! Fixed it

date/date.php
isfedorov: Fixed

sqlite3/sqlite3.php
isfedorov: Yep, but still it was optional in earlier versions and default value here is a way to mark it as optional

standard/standard_3.php
```diff
@@ -759,7 +760,11 @@ function long2ip(int $ip): string|false {}
  * is provided, or false on an error.
  */
 #[Pure]
-function getenv(?string $name, bool $local_only = false): array|string|false {}
+function getenv(
+    #[PhpStormStubsElementAvailable(from: '5.3', to: '7.0')] $varname,
+    #[PhpStormStubsElementAvailable(from: '7.1')] ?string $name = null,
+    bool $local_only = false
```
DmitryTronin, line 766: Hm, `$local_only` was introduced in 5.6, so from 5.3 till 5.5 we can consider it as unavailable.

isfedorov: Good catch! More over according to https://www.php.net/manual/en/function.getenv.php `7.0.9 | The local_only parameter has been added.` so there is a bug in docs as well. I'll update stubs to `#[PhpStormStubsElementAvailable(from: '5.6')] bool $local_only = false`

standard/standard_3.php
```diff
@@ -790,7 +795,11 @@ function putenv(string $assignment): bool {}
  * @return string[]|false[]|false This function will return an array of option / argument pairs or false on
  * failure.
  */
-function getopt(string $short_options, array $long_options, &$rest_index): array|false {}
+function getopt(
+    string $short_options,
+    array $long_options,
+    #[PhpStormStubsElementAvailable(from: '7.1')] &$rest_index
```
DmitryTronin, line 801: Small note: since 8.0 `$rest_index`  has a default value of `null`: https://3v4l.org/gFr1v#v800 & https://www.php.net/manual/en/function.getopt.php

standard/standard_7.php
```diff
@@ -47,20 +47,34 @@
...
  * feof). If the call fails, it will return false
  */
-function fsockopen(string $hostname, int $port = -1, &$error_code, &$error_message, ?float $timeout) {}
+function fsockopen(
+    string $hostname,
+    #[PhpStormStubsElementAvailable(from: '5.3', to: '7.0')] int $port,
+    #[PhpStormStubsElementAvailable(from: '7.1')] int $port = -1,
+    &$error_code,
+    &$error_message,
+    ?float $timeout
```
DmitryTronin, line 56: `$error_code`, `$error_message`, `$timeout` have default value `null` since 8.0.

standard/standard_7.php
```diff
@@ -47,20 +47,34 @@
...
  * @return resource|false
  */
-function pfsockopen(string $hostname, int $port = -1, &$error_code, &$error_message, ?float $timeout) {}
+function pfsockopen(
+    string $hostname,
+    #[PhpStormStubsElementAvailable(from: '5.3', to: '7.0')] int $port,
+    #[PhpStormStubsElementAvailable(from: '7.1')] int $port = -1,
+    &$error_code,
+    &$error_message,
+    ?float $timeout
```
DmitryTronin, line 76: same as for `fsockopen()`: $error_code, $error_message, $timeout have default value null

standard/standard_8.php
```diff
@@ -1090,5 +1090,6 @@ function array_slice(array $array, int $offset, ?int $length, bool $preserve_key
 #[Pure]
 function array_merge(
     #[PhpStormStubsElementAvailable(from: '5.3', to: '7.3')] $array,
-    array ...$arrays
+    #[PhpStormStubsElementAvailable(from: '5.3', to: '7.0')] $arrays,
+    #[PhpStormStubsElementAvailable(from: '7.1')] array ...$arrays
```
DmitryTronin, line 1094: 1) There's weird result for 5.4-5.5: https://3v4l.org/Y0Ggj#v540:
Parameter name: string(4) "arr1"
Parameter name: string(4) "arr2"
Parameter name: string(3) "...". It is optional. 
We should probably add it?

standard/standard_9.php
```diff
@@ -36,27 +40,35 @@ function array_merge_recursive(#[PhpStormStubsElementAvailable(from: '5.3', to:
...
  * The array from which elements will be extracted.
  * </p>
  * @return array or null if an error occurs.
  */
 #[Pure]
-function array_replace(array $array, array ...$replacements): array {}
+function array_replace(
+    array $array,
+    #[PhpStormStubsElementAvailable(from: '5.3', to: '7.0')] $arrays,
+    #[PhpStormStubsElementAvailable(from: '7.1')] array ...$replacements
```
DmitryTronin, line 52: same stuff here: 3rd parameter named `...` for 5.4-5.5 PHP versions: https://3v4l.org/31rEt#v540

standard/standard_9.php
```diff
@@ -36,27 +40,35 @@ function array_merge_recursive(#[PhpStormStubsElementAvailable(from: '5.3', to:
...
  * The array from which elements will be extracted.
  * </p>
  * @return array an array, or null if an error occurs.
  */
 #[Pure]
-function array_replace_recursive(array $array, array ...$replacements): array {}
+function array_replace_recursive(
+    array $array,
+    #[PhpStormStubsElementAvailable(from: '5.3', to: '7.0')] $arrays,
+    #[PhpStormStubsElementAvailable(from: '7.1')] array ...$replacements
```
DmitryTronin, line 70: and here: https://3v4l.org/FrptN#v540 - 3rd parameter named `...` for 5.4-5.5 PHP versions

bz2/bz2.php
```diff
@@ -137,7 +138,12 @@ function bzerror($bz): array {}
  * @return string|int The compressed string, or an error number if an error occurred.
  */
 #[Pure]
-function bzcompress(string $data, int $block_size = 4, int $work_factor = 0): string|int {}
+function bzcompress(
+    string $data,
+    #[PhpStormStubsElementAvailable(from: '5.3', to: '7.0')] int $blocksize,
+    #[PhpStormStubsElementAvailable(from: '7.1')] int $block_size = 4,
```
DmitryTronin, line 144: $block_size doesn't have a default value till 8.0.
7.4 result:
```
Parameter name: string(6) "source"
Default value unavailable
Parameter name: string(9) "blocksize"
It is optional.
Default value unavailable
Parameter name: string(10) "workfactor"
It is optional.
Default value unavailable
```

vs 8.0

```
Parameter name: string(4) "data"
Default value unavailable
Has paramter type: array(1) {
  [0]=>
  string(6) "string"
}
Parameter name: string(10) "block_size"
It is optional.
Default Value available. It is: int(4)
Has paramter type: array(1) {
  [0]=>
  string(3) "int"
}
Parameter name: string(11) "work_factor"
It is optional.
Default Value available. It is: int(0)
Has paramter type: array(1) {
  [0]=>
  string(3) "int"
}
```

mysqli/mysqli.php
```diff
@@ -947,11 +950,11 @@ class mysqli_result implements IteratorAggregate
 
     /**
      * Constructor (no docs available)
-     * @param object $mysql [optional]
+     * @param object $mysql
      * @param int $result_mode [optional]
      */
     public function __construct(
-        #[PhpStormStubsElementAvailable(from: '8.0')] mysqli $mysql = null,
+        #[PhpStormStubsElementAvailable(from: '8.0')] mysqli $mysql,
```
DmitryTronin, line 957: Hm, `var_dump((new ReflectionMethod('mysqli_result::__construct'))->getNumberOfParameters());` returns 0.
All my `->getParameters()` tests are also returning nothing. Looks like this particular constructor doesn't have parameters or I'm missing something

openssl/openssl.php
```diff
@@ -418,7 +428,8 @@ function openssl_pkcs12_export(
     &$output,
     #[LanguageLevelTypeAware(['8.0' => 'OpenSSLAsymmetricKey|OpenSSLCertificate|array|string'], default: 'resource|array|string')] $private_key,
     string $passphrase,
-    array $options
+    #[PhpStormStubsElementAvailable(from: '5.3', to: '7.0')] $args,
+    #[PhpStormStubsElementAvailable(from: '7.1')] array $options = []
```
DmitryTronin, line 432: optional `$options` is called `$args`. Probably we can rename it to `$arguments` to avoid duplication

DmitryTronin (approved):
All checked. What a tremendous amount of work!
All good to me, after addressing the rest of my comments.

mysqli/mysqli.php
DmitryTronin: My bad, run in on < 8.0.

standard/standard_3.php
isfedorov: Hm..That's true, but let's leave it as is as currently it's marked as optional using phpdoc for all PHP versions and replacing it with default value we can introduce false positive inspection warnings "Redundant parameter" for versions < 8.0.

standard/standard_7.php
isfedorov: The answer is the same as above. let's leave it as is as they are already marked as optional

standard/standard_7.php
isfedorov: The answer is the same as above. let's leave it as is as they are already marked as optional

standard/standard_8.php
isfedorov: I don't think it makes sense as 5.4-5.5 are too old versions and complication of stubs for the sake of 2 almost not used versions doesn't worth it.

standard/standard_9.php
isfedorov: Answer it the same as above

bz2/bz2.php
isfedorov: Yep, but still, it was optional that's why I had to add a default value to it.

openssl/openssl.php
isfedorov: We can't rename `options` because `options` is the name that comes from reflection and if we will change it then it will break named arguments. If you mean to rename `args` then it actually no matter what this name is as it's used in PHP < 8 and it will not bring problems with named arguments

PR #1291, opened 2021-11-27
`SplFixedArray` does not implement `Iterator`

kukulich:
See https://3v4l.org/UaXVt


kukulich:
Hmm, it' changed in PHP 8.0, so it cannot be fixed this way.

PR #1292, opened 2021-11-27
`Exception::__clone()` and `Error::__clone()` are not final in PHP 8.1

kukulich:
See https://3v4l.org/M1KSI


isfedorov:
@kukulich Thank you for the PR! Could you please remove these methods from `tests/TestData/mutedProblems.json` as well?

kukulich:
@isfedorov Done.

PR #1293, opened 2021-11-27
`Directory` methods do not accept `$dir_handle` parameter since PHP 8.0

kukulich:
See https://3v4l.org/688Ut or https://www.php.net/manual/en/directory.close.php


isfedorov:
@kukulich Thank you for the PR! Could you please remove these methods from tests/TestData/mutedProblems.json as well?

kukulich:
@isfedorov Done.

PR #1295, opened 2021-11-27
Fixed `array_multisort()` in PHP < 8.0

kukulich:


standard/standard_8.php
```diff
@@ -933,12 +965,8 @@ function range($start, $end, int|float $step = 1): array {}
  * </p>
  * @return bool true on success or false on failure.
  */
-function array_multisort(
-    &$array,
-    #[PhpStormStubsElementAvailable(from: '5.3', to: '7.4')] $sort_order = SORT_ASC,
```
DmitryTronin, line 938: Could you please describe why the previous signature didn't work for you? It looks valid for us right now.

kukulich: There are differences in parameter names.

isfedorov: @kukulich parameter names are important only since PHP 8.0 as named arguments were introduced in PHP 8.0

kukulich: If you don't care about parameters names, the PR can be closed - I supposed. I expected that you want to solve `mutedProblems` when possible.

isfedorov: > I expected that you want to solve mutedProblems when possible.
That's true. We really want to solve as many muted problems as possible and I think that muted problem for array_multisort is already obsolete even without changing the signature (will check it). Anyway thank you very much for your effort, you have done a great work!

DmitryTronin:
Closing as per conversation in the comments.

PR #1299, opened 2021-12-02
Fix RedisCluster::setOption|getOption types

ostrolucky:
See https://github.com/phpredis/phpredis/issues/2038 and https://github.com/vimeo/psalm/pull/7030

### DIFF
--- a/redis/RedisCluster.php
+++ b/redis/RedisCluster.php
@@ -2817,20 +2817,20 @@ class RedisCluster
     /**
      * Get client option
      *
-     * @param   string $name parameter name
+     * @param   int $option parameter
      *
-     * @return  int     Parameter value.
+     * @return  int|string     Parameter value.
      * @example
      * // return RedisCluster::SERIALIZER_NONE, RedisCluster::SERIALIZER_PHP, or RedisCluster::SERIALIZER_IGBINARY.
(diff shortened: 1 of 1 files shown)


redis/RedisCluster.php
```diff
@@ -2817,20 +2817,20 @@ public function clearLastError() {}
...
+     * @param   string $option parameter name
      *
-     * @return  int     Parameter value.
+     * @return  int|string     Parameter value.
      * @example
      * // return RedisCluster::SERIALIZER_NONE, RedisCluster::SERIALIZER_PHP, or RedisCluster::SERIALIZER_IGBINARY.
      * $redisCluster->getOption(RedisCluster::OPT_SERIALIZER);
      */
-    public function getOption($name) {}
+    public function getOption($option) {}
```
DmitryTronin, line 2827: Based on https://github.com/phpredis/phpredis/blob/develop/redis_cluster.stub.php#L122  it's `int $option`, right? 
It means that `@param   string $option parameter name` should be `int`.

DmitryTronin (changes requested):

redis/RedisCluster.php
ostrolucky: yep. fixed

PR #1312, opened 2021-12-14
Update dom_c.php

WinterSilence:
https://3v4l.org/CM4gJ#v8.1.0

### DIFF
--- a/dom/dom_c.php
+++ b/dom/dom_c.php
@@ -1245,17 +1245,10 @@ class DOMDocument extends DOMNode implements DOMParentNode
 /**
  * The DOMNodeList class
  * @link https://php.net/manual/en/class.domnodelist.php
+ * @property-read int $length The number of nodes in the list. The range of valid child node indices is 0 to length - 1 inclusive.
  */
-class DOMNodeList implements IteratorAggregate, Countable
+class DOMNodeList implements IteratorAggregate, Traversable, Countable
 {
-    /**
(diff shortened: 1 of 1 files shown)


dom/dom_c.php
```diff
@@ -1245,17 +1245,10 @@ public function registerNodeClass(
...
-class DOMNodeList implements IteratorAggregate, Countable
+class DOMNodeList implements IteratorAggregate, Traversable, Countable
 {
-    /**
-     * @var int
-     * The number of nodes in the list. The range of valid child node indices is 0 to length - 1 inclusive.
-     * @link https://php.net/manual/en/class.domnodelist.php#domnodelist.props.length
-     */
-    #[LanguageLevelTypeAware(['8.1' => 'int'], default: '')]
-    public $length;
```
DmitryTronin, line 1257: `$length` exists in previous PHP versions even though reflection shows otherwise:
- https://3v4l.org/NfD2G
- https://github.com/php/php-src/blob/PHP-7.4.27/ext/dom/nodelist.c#L59 - you can check other branches to check that the property exists there also

DmitryTronin (changes requested):

dom/dom_c.php
WinterSilence: @DmitryTronin that's parameter "read only", you can declare that's only by PHPDoc, see DOMNamedNodeMap

PR #1322, opened 2022-01-14
Fix `SimpleXMLElement::children()` return type declaration

mpdude:
Hey there 👋🏼,

this is my first-time contribution in this repo, so apologies if I am missing something or am not familiar with the typical workflows for reporting issues.

I noticed the return type for `SimpleXMLElement::children()` is written as nullable in the stub, but in fact the PHP.net documentation (https://www.php.net/manual/en/simplexmlelement.children.php) says it...

> Returns a SimpleXMLElement element, whether the node has children or not.

Obviously, php.net itself shows the return type as being nullable as well. So, I wonder if that's something that should be fixed upstream first and then will make its way here magically?

### DIFF
--- a/SimpleXML/SimpleXML.php
+++ b/SimpleXML/SimpleXML.php
@@ -140,7 +140,7 @@ class SimpleXMLElement implements Traversable, ArrayAccess, Countable, Iterator,
      * <i>ns</i> will be regarded as a namespace
      * URL.
      * </p>
-     * @return static|null a <b>SimpleXMLElement</b> element, whether the node
+     * @return static a <b>SimpleXMLElement</b> element, whether the node
      * has children or not.
      * @since 5.0.1
      */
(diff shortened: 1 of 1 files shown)


isfedorov:
Hi @mpdude , thank you for the PR, but according to stubs from php-src the return type is nullable https://github.com/php/php-src/blob/master/ext/simplexml/simplexml.stub.php#L36 so probably this is a bug in php.net docs and a separate PR should be submitted to the https://github.com/php/doc-en and other doc related repositories

mpdude:
X-Ref https://github.com/php/php-src/pull/7969

isfedorov:
Please note that the next code
```
<?php

$xml = '
<root attr="Hello"/>
';

$sxe = new SimpleXMLElement($xml);

$sxe_xpath = $sxe->xpath('/root/@attr')[0];

$children = $sxe_xpath->children();
var_dump($sxe_xpath);
var_dump($children);
```
Will return 
```
object(SimpleXMLElement)#2 (1) {
  ["@attributes"]=>
  array(1) {
    ["attr"]=>
    string(5) "Hello"
  }
}
NULL
```
https://3v4l.org/6LXbY

mpdude:
Yes, this was also pointed out over at https://github.com/php/php-src/pull/7969#issuecomment-1017190402. 

`SimpleXMLElement::children()` will always return another `SimpleXMLELement` for elements that represent tags, and `NULL` for elements that represent attributes.

Probably the documentation needs to be updated to mention this case.

PR #1324, opened 2022-01-17
Fixed return type for methods when MULTI/EXEC is used.

michnovka:
As specified in [official docs](https://github.com/phpredis/phpredis#return-value-142) , if MULTI is used:

> ... Once in multi-mode, all subsequent method calls return the same object until exec() is called.

### DIFF
--- a/redis/Redis.php
+++ b/redis/Redis.php
@@ -3975,7 +3975,7 @@ class Redis
      * @param   string  $script
      * @param   array   $args
      * @param   int     $numKeys
-     * @return  mixed|Redis   @see eval() or Redis if in multi mode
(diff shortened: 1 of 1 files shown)


isfedorov:
@michnovka Thank you for the PR! Could you please also update phpdocs for the return values to reflect that Redis can be returned in multi-mode. Something like `@return bool|Redis TRUE if the command is successful or Redis in multi-mode`

michnovka:
@isfedorov I ran it through regexp replace, but should be fine

michnovka:
I manually reviewed the changes, I believe the file is now ready for merge.

PR #1339, opened 2022-02-02
Update values of `Memcached::HAVE_**` constants

paulbalandan:
Related to https://github.com/phpstan/phpstan/issues/6263

According to [PHP documentation](https://www.php.net/manual/en/memcached.constants.php), the `Memcached::HAVE_**` constants are booleans. The stub here is correct as to the PHPDoc only, correctly typing them as booleans. However, the values themselves are integers of `1` and `0`. I think this was set as such because in the C source code, the values used are `1` and `0` but the registering function is `REGISTER_MEMC_CLASS_CONST_BOOL`.
https://github.com/php-memcached-dev/php-memcached/blob/d7daff649406efbd43af3253212b1cb946ff3f85/php_memcached.c#L4362-L4366

However, checking with `var_dump`, these constants are actually booleans.
https://github.com/paulbalandan/phpstan-bug6263/runs/5032025575?check_suite_focus=true

### DIFF
--- a/memcached/memcached.php
+++ b/memcached/memcached.php
@@ -50,14 +50,14 @@ class Memcached
      * <p>Type: boolean.</p>
      * @link https://php.net/manual/en/memcached.constants.php
      */
-    public const HAVE_IGBINARY = 0;
+    public const HAVE_IGBINARY = false;
 
     /**
      * <p>Indicates whether JSON serializer support is available.</p>
      * <p>Type: boolean.</p>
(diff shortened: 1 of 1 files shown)


paulbalandan:
What should I do to fix the Docker failure? Or it seems unrelated to my PR?

isfedorov:
@paulbalandan this is a known problem with php-src, that sockets extension fails to build. Please ignore this build problem for now. We will check your changes anyway.

PR #1343, opened 2022-02-22
Language injection for common function args

uuf6429:
Fixes [WI-65511](https://youtrack.jetbrains.com/issue/WI-65511)

### DIFF
--- a/pcre/pcre.php
+++ b/pcre/pcre.php
@@ -3,6 +3,7 @@
 // Start of pcre v.
 use JetBrains\PhpStorm\Internal\LanguageLevelTypeAware;
 use JetBrains\PhpStorm\Internal\PhpStormStubsElementAvailable;
+use JetBrains\PhpStorm\Language;
 use JetBrains\PhpStorm\Pure;
 
 /**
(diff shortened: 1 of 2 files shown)


uuf6429:
Any thoughts on adding `Language` attribute to arguments with union types? For example:
```php
function preg_replace_callback(
    #[Language('PhpRegExp')]     // is this correct? 🤨
    array|string $pattern,
````

**Comment by uuf6429** [comment]
Would it make sense to add the attribute to as many places as possible in this PR? eg:
```php
function mysqli_query(
    mysqli $mysql,
    #[Language('MySQL')] string $query,
```
etc..

DmitryTronin (approved):

pcre/pcre.php
```diff
@@ -157,7 +158,7 @@
  * matches given <i>subject</i>, 0 if it does not, or <b>FALSE</b>
  * if an error occurred.
  */
-function preg_match(string $pattern, string $subject, &$matches, int $flags = 0, int $offset = 0): int|false {}
+function preg_match(#[Language('PhpRegExp')] string $pattern, string $subject, &$matches, int $flags = 0, int $offset = 0): int|false {}
```
DmitryTronin, line 161: This is not required since IDE insert regexp injection automatically.

DmitryTronin (changes requested):

pcre/pcre.php
uuf6429: When that (Language Injection) functionality was originally implemented, PHP attributes didn't exist.

Going forward, it might make sense to implement the injection as an attribute now and remove the old/current functionality later on.

The SQL injection happens by the IDE guesses by checking if strings start with a keyword. Additionally, the SQL flavour is based on scope (which users have to set up). I personally consider this approach to be fundamentally wrong now that we have better ways to hint the language.

I don't have a strong opinion on any of this, my thought is that it may benefit the IDE development.

DmitryTronin (approved):

PR #1346, opened 2022-02-23
Generate dump correctly, gitignore it, display docker build output

uuf6429:
There are 3 main changes:
- the run tests script was calling the dumper incorrectly, causing a php error and missing json dump
- the dumped file should have been gitignored
- since it takes some time to build the docker images, I opted for showing the output

I don't have a strong opinion on these changes, but I thought they would be useful.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,9 +11,11 @@ rest-client.private.env.json
 .idea/workspace.xml
 .idea/shelf
 
-#Composer
+# Composer
 vendor
 composer.lock
 
 .php_cs.cache
 .phpunit.result.cache
+
+ReflectionData.json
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -11,9 +11,11 @@ rest-client.private.env.json
...
 
-#Composer
+# Composer
 vendor
 composer.lock
 
 .php_cs.cache
 .phpunit.result.cache
+
+ReflectionData.json
```
uuf6429, line 21: I just saw that the shell script also tries to remove this file in the end, but maybe gitignore'ing it too would be safer?

isfedorov: Agree

PR #1347, opened 2022-02-23
Move stub files into src 💥

uuf6429:
Crazy PR 😄  luckily, git history is not lost. 👍 

The purpose of this is mostly organisational. Changes:
- move all stub files to `src/` (I chose that because it seems to be standard in PHP, but we could also use something else)
- replaced `check-stub-map` shell script with a PHPUnit test (less OS-specific dependencies + added advantage of showing mismatch diff)
- add the test as part of allTests.sh


tests/Tools/generate-stub-map
```diff
@@ -243,11 +244,11 @@ use const PHP_EOL;
...
     }
 
-    $mapWithRelativeFilePaths = array_map(static function (array $files) use ($phpStormStubsDirectory) : array {
+    $mapWithRelativeFilePaths = array_map(static function (array $files) use ($mapFile) : array {
         ksort($files);
 
-        return array_map(static function (string $filePath) use ($phpStormStubsDirectory) : string {
-            return str_replace('\\', '/', substr($filePath, strlen($phpStormStubsDirectory)));
+        return array_map(static function (string $filePath) use ($mapFile) : string {
+            return str_replace('\\', '/', substr($filePath, strlen(dirname($mapFile)) + 1));
```
uuf6429, line 251: Presumably, the path that we want should be relative to the map file, not the stub directory.
This works (better) in the current case, but it would be a bad solution in case of diverging paths.

tests/Tools/generate-stub-map
```diff
@@ -205,7 +206,7 @@ use const PHP_EOL;
             continue;
         }
 
-        if (in_array($directoryInfo->getBasename(), ['tests', 'meta', 'vendor', 'couchbase_v2'], true)) {
+        if (in_array($directoryInfo->getBasename(), $excludedStubDirNames, true)) {
```
uuf6429, line 209: Thanks to the move, tests/vendor will no longer conflict with the generator, so these can be removed from here.

tests/Tools/generate-stub-map
```diff
@@ -94,7 +94,8 @@ use const PHP_EOL;
         }
     }
 
-    $phpStormStubsDirectory = __DIR__ . '/../../';
+    $phpStormStubsDirectory = __DIR__ . '/../../src/';
+    $excludedStubDirNames = ['meta', 'couchbase_v2'];
```
uuf6429, line 98: Not sure why these two need to be excluded - maybe it would be a good idea to add comments @isfedorov?
```php
// excluded because ...
$excludedStubDirNames = ['meta', 'couchbase_v2'];
```
or
```php
$excludedStubDirNames = [
    'meta',           // because this is meta
    'couchbase_v2',   // because some other reason
];
```
(second one could be nicer because of individual git history)

isfedorov:
Hi @uuf6429 , 
Unfortunately, this PR can't be merged and there are several reasons for that.
1) Such change will break PhpStorm's internal build process. (Process of copying of stubs to the build will have to be fixed)
2) Such change potentially can break 3rd party tools that use these stubs. Such a structure of repo has a very long history and probably a lot of tools rely on exactly this structure.
3) Such change will require updates of "Custom stubs path" setting. (Either docs or processing of setting should be updated)
Taking into account these disadvantages now we need to answer the question, what problem does this change solve? As you properly mentioned, this change is mostly organizational and it seems that it doesn't worth all the fixes that will have to be done after merge. But anyway thank you for the contribution! We highly appreciate it!

uuf6429:
> Hi @uuf6429 , Unfortunately, this PR can't be merged and there are several reasons for that.
> 
> 1. Such change will break PhpStorm's internal build process. (Process of copying of stubs to the build will have to be fixed)
> 2. Such change potentially can break 3rd party tools that use these stubs. Such a structure of repo has a very long history and probably a lot of tools rely on exactly this structure.
> 3. Such change will require updates of "Custom stubs path" setting. (Either docs or processing of setting should be updated)
>    Taking into account these disadvantages now we need to answer the question, what problem does this change solve? As you properly mentioned, this change is mostly organizational and it seems that it doesn't worth all the fixes that will have to be done after merge. But anyway thank you for the contribution! We highly appreciate it!

I figured there would be some historical reason to keep it as is, but no problem.
It was an interesting exercise anyway.

If you don't mind, I'll extract anything useful I had in this PR into another one.

isfedorov:
@uuf6429 
> If you don't mind, I'll extract anything useful I had in this PR into another one.

Yes, sure

PR #571, opened 2019-04-18
Add .editorconfig

SerafimArts:
Settings for different IDE in order to adhere to a common style when editing.

For example, the `composer.json` contains 2 spaces, however, by default, the IDE settings correspond to 4 and `Ctrl + Alt + Shift + L` break formatting if `.editorconfig` file is missing.

PR #1349, opened 2022-02-28
Add missing methods and constants of DIO

alphp:
Add missing methods dio_raw() and dio_serial().
Add missing constants.


PhpStormStubsMap.php
```diff
@@ -6986,7 +6988,44 @@ final class PhpStormStubsMap
   'DATE_RFC850' => 'date/date_d.php',
   'DATE_RSS' => 'date/date_d.php',
   'DATE_W3C' => 'date/date_d.php',
-  'DAY_1' => 'standard/standard_defines.php',
+  'DATE_W3C' => 'dio/dio_d.php',
```
DmitryTronin, line 6991: 1) That's a duplicate array key (`DATE_W3C`)
2) Please adjust formatting. It is using 2 spaces by default, while you have this fragment indented with a tab.

dio/dio.php
```diff
@@ -12,117 +12,160 @@
...
  * Some commands require additional arguments args to be supplied.
  * @link https://www.php.net/manual/en/function.dio-fcntl.php
- * @param resource $fd The file descriptor returned by dio_open().
- * @param int $cmd The file descriptor returned by dio_open().
- * @param mixed ...$args args is an associative array, when cmd is F_SETLK or F_SETLLW, with the following keys:
- * start - offset where lock begins
- * length - size of locked area. zero means to end of file
- * whence - Where l_start is relative to: can be SEEK_SET, SEEK_END and SEEK_CUR
- * type - type of lock: can be F_RDLCK (read lock), F_WRLCK (write lock) or F_UNLCK (unlock)
+ * @param resource $fd
```
DmitryTronin, line 38: Is there a reason you've split tags description in multiple lines?

dio/dio.php
```diff
@@ -12,117 +12,160 @@
...
- * O_EXCL - if both O_CREAT and O_EXCL are set and the file already exists, dio_open() will fail.
- * O_TRUNC - if the file exists and is opened for write access, the file will be truncated to zero length.
- * O_APPEND - write operations write data at the end of the file.
- * O_NONBLOCK - sets non blocking mode.
- * O_NOCTTY - prevent the OS from assigning the opened file as the process's controlling terminal when opening a TTY device file.
- * @param int $mode If flags contains O_CREAT, mode will set the permissions of the file (creation permissions). mode is required for correct operation when O_CREAT is specified in flags and is ignored otherwise.
+ * The flags parameter is a bitwise-ORed value comprising flags from the following list.
+ * This value must include one of **O_RDONLY**, **O_WRONLY**, or **O_RDWR**.
+ * Additionally, it may include any combination of the other flags from this list.
+ * - **O_RDONLY** - opens the file for read access.
```
DmitryTronin, line 63: Resulting formatting of these changes doesn't look correct, unfortunately: 
image

Please fix or just leave as it is, since I don't see any benefits in these changes.

dio/dio.php
```diff
@@ -12,117 +12,160 @@
...
+ *
  * dio_read ( resource $fd [, int $len = 1024 ] ) : string
- * The function dio_read() reads and returns len bytes from file with descriptor fd.
- * @param resource $fd The file descriptor returned by dio_open().
- * @param int $len The number of bytes to read. If not specified, dio_read() reads 1K sized block.
- * @return int The bytes read from fd.
+ *
+ * @param resource $fd
+ * The file descriptor returned by dio_open().
+ * @param int $len
```
DmitryTronin, line 87: Unnecessary line break

dio/dio.php
```diff
@@ -12,117 +12,160 @@
...
 
 /**
+ * Closes the file descriptor given by fd.
+ *
  * dio_close ( resource $fd ) : void
- * The function dio_close() closes the file descriptor fd.
+ *
  * @link https://www.php.net/manual/en/function.dio-close.php
- * @param resource $fd The file descriptor returned by dio_open()
+ * @param resource $fd
```
DmitryTronin, line 25: Unnecessary line break

DmitryTronin (changes requested):

PhpStormStubsMap.php
```diff
@@ -2050,7 +2050,9 @@ final class PhpStormStubsMap
   'dio_tcsetattr' => 'dio/dio.php',
   'dio_truncate' => 'dio/dio.php',
   'dio_write' => 'dio/dio.php',
-  'dir' => 'standard/standard_7.php',
+  'dio_raw' => 'dio/dio.php',
+  'dio_serial' => 'dio/dio.php',
+	'dir' => 'standard/standard_7.php',
```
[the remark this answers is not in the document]
alphp: My fault, I didn't check if the indentation was done correctly.
I've added an .editconfig that helps keep the indentation and line endings formatted correctly.

alphp:
The format of the DocBlock corresponds to the MarkDown style that I use with vscode. I have corrected the differences with the specified format.

alphp:
With the formatting I had applied, this is what it looks like in vscode:
image

And if you check the source code it reads perfectly, much clearer than with ul and li tags thanks to the MarkDown syntax

```php
/**
 * Opens a file (creating it if necessary) at a lower level than theC library input/ouput stream functions allow
 *
 * dio_open ( string $filename , int $flags [, int $mode = 0 ] ) : resource
 *
 * @link https://www.php.net/manual/en/function.dio-open.php
 * @param string $filename The pathname of the file to open.
 * @param int $flags
 * The flags parameter is a bitwise-ORed value comprising flags from the following list.
 * This value must include one of **O_RDONLY**, **O_WRONLY**, or **O_RDWR**.
 * Additionally, it may include any combination of the other flags from this list.
 * - **O_RDONLY** - opens the file for read access.
 * - **O_WRONLY** - opens the file for write access.
 * - **O_RDWR** - opens the file for both reading and writing.
 * - **O_CREAT** - creates the file, if it doesn't already exist.
 * - **O_EXCL** - if both O_CREAT and O_EXCL are set and the file already exists, dio_open() will fail.
 * - **O_TRUNC** - if the file exists and is opened for write access, the file will be truncated to zero length.
 * - **O_APPEND** - write operations write data at the end of the file.
 * - **O_NONBLOCK** - sets non blocking mode.
 * - **O_NOCTTY** - prevent the OS from assigning the opened file as the process's controllingterminal when opening a TTY device file.
 * @param int $mode
 * If flags contains **O_CREAT**, mode will set the permissions of the file (creation permissions).
 * mode is required for correct operation when **O_CREAT** is specified in flags and is ignored otherwise.
 * The actual permissions assigned to the created file will be affected by the process's umask setting as per usual.
 * @return resource|false A file descriptor or FALSE on error.
 */
```
I also find it very useful to have as much information about the functions as possible without having to refer to the documentation outside of the IDE.
That's my opinion, but this pull-request is not trying to change the style of the project, but to provide missing definitions.
This is the explanation of why I have formatted the DocBlocks as I had.

PhpStormStubsMap.php
alphp: Removed

DmitryTronin:
Thanks for the your update.
Unfortunately markdown lists aren't supported in PhpDocs by us yet: https://youtrack.jetbrains.com/issue/WI-25185. That's why we're using `<li>` currently.
Regarding the addition of editorconfig file: we do not want to integrate it into our workflow currently. We already have PHP CS Fixer for code quality inspections. See the reasoning behind this at https://github.com/JetBrains/phpstorm-stubs/pull/571 & https://youtrack.jetbrains.com/issue/WI-46428

DmitryTronin (changes requested):
Please remove the .editorconfig file

alphp:
I don't have a linux machine that supports docker, I'll have to create a VM.

PhpStormStubsMap.php
```diff
@@ -11302,47 +11328,47 @@ final class PhpStormStubsMap
   'SE_NOPREFETCH' => 'imap/imap.php',
   'SE_UID' => 'imap/imap.php',
   'SID' => 'standard/standard_defines.php',
-  'SIGABRT' => 'pcntl/pcntl.php',
-  'SIGALRM' => 'pcntl/pcntl.php',
+  'SIGABRT' => 'swoole/constants.php',
```
DmitryTronin, line 11331: Please update the rebase your changes on a master branch & re-generate stubs map.
These changes are outdated because of https://github.com/JetBrains/phpstorm-stubs/pull/1353

DmitryTronin (changes requested):

DmitryTronin:
Hi, Something went wrong with your merge (you can see duplicate commits were included) so I've created a clean PR based on the changes you've introduced: #1357. It is now merged. Feel free to open a new PR if I forget something.

PR #1354, opened 2022-03-13
Internal classes should not be part of `PhpStormStubsMap`

kukulich:


DmitryTronin:
Thanks for the PR. Theses classes are required for our engine, they were added synthetically as a result of https://youtrack.jetbrains.com/issue/WI-60439 & https://youtrack.jetbrains.com/issue/WI-60460.

PR #1358, opened 2022-03-28
added new GD functions

DmitryTronin:


williamdes (approved):
> @return bool Returns true on success or false on failure. However, **if libgd fails to output the image, this function returns true.**

😄

PR #1359, opened 2022-03-30
Fix `DOMNode` & `DOMElement` property types

VasekPurchart:
Due to false positive errors from PHPStan I found out that the stubs from this repo have incorrect types for `DOMNode` & `DOMElement` properties. All of these are actually correctly noted in the docs: [DOMNode](https://www.php.net/manual/en/class.domnode.php), [DOMElement](https://www.php.net/manual/en/class.domelement.php) and I have also verified all of these in runtime:

* `DomNode::$nodeValue` - example: https://3v4l.org/qa86V
* `DomNode::$localName` - example: https://3v4l.org/5613l
* `DomElement::$parentNode` - example: https://3v4l.org/Dv8CN
* `DomElement::$firstChild` - example: https://3v4l.org/FcLun
* `DomElement::$lastChild` - example: https://3v4l.org/TifIr
* `DomElement::$previousSibling` - example: https://3v4l.org/t3rAL
* `DomElement::$nextSibling` - example: https://3v4l.org/H77QM

I am not sure why the properties on `DomElement` which are inherited from `DomNode` are "redeclared", maybe they should not? Also generally does it make any sense since properties should be invariant?

### DIFF
--- a/dom/dom_c.php
+++ b/dom/dom_c.php
@@ -22,7 +22,7 @@ class DOMNode
     public $nodeName;
 
     /**
-     * @var string
+     * @var string|null
      * The value of this node, depending on its type
      * @link https://php.net/manual/en/class.domnode.php#domnode.props.nodevalue
      */
@@ -119,7 +119,7 @@ class DOMNode
     public $prefix;
 
     /**
-     * @var string
+     * @var string|null
      * Returns the local part of the qualified name of this node.
      * @link https://php.net/manual/en/class.domnode.php#domnode.props.localname
      */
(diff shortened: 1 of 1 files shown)


williamdes (approved):

VasekPurchart:
@DmitryTronin Qodana finally came through, could you please allow to run again the other workflows?

VasekPurchart:
Thanks for merging :)

DmitryTronin:
As for re-declared properties: they are there because of https://youtrack.jetbrains.com/issue/WI-6760. You can also find them documented at https://www.php.net/manual/en/class.domelement.php.
Currently they are required for a correct code completion in cases described in WI-6760. Removing those declaration would break code completion & I don't see a quick & efficient way of removing them currently.

VasekPurchart:
> You can also find them documented at https://www.php.net/manual/en/class.domelement.php.

Yea, but they are documented only as inherited, I think PHP documentation shows these for all classes. See for example https://www.php.net/manual/en/class.typeerror.php and https://www.php.net/manual/en/class.exception.php .

> they are there because of https://youtrack.jetbrains.com/issue/WI-6760
> Removing those declaration would break code completion

Unfortunately I think that this PR has the same effect - but only because the original usecase in the issue is "wrong".

```php
/* @var \DOMElement $node */
$node->parentNode->removeAttributeNS
```

The fact that `$node->parentNode` does not have `removeAttributeNS` is correct, precisely because it can *not have* it - as seen in the example above and if I add accessing this attribute, it obviously results in an error: https://3v4l.org/2JMN3 .

So the only correct way in such a case is either handle all the cases, or if the parent can only indeed be a DOMElement, than code should look something like this (which would be obviously fine in PHPStorm and with code completion):

```php
/* @var \DOMElement $node */
$parentElement = $node->parentNode;
assert($parentElement instanceof DOMElement);
$parentElement->removeAttributeNS;
```

DmitryTronin:
Thanks for the comments. I'll review the info & update you in the nearest future.

DmitryTronin:
@VasekPurchart Yeah, that looks valid. You can give it a go & create a new PR addressing this, or let me know - I'll report this to our tracker to address in the future.

PR #1365, opened 2022-04-07
Adding stub for https://github.com/kjdev/php-ext-snappy.

schodemeiss:
Added stubs for the Snappy Compression extension found here: https://github.com/kjdev/php-ext-snappy.


DmitryTronin:
Do you have any installations or usage stats for this extension?

schodemeiss:
> Do you have any installations or usage stats for this extension?

I do not; only anecdotal evidence of a handful of companies using it in production. The extension I reference in the docs does seem to be the most current and the one I've seen most used in the field.

schodemeiss:
Ah, I didn't notice we had to update the map. I shall do that, make another commit and go again :-).

DmitryTronin:
Just one more thing, as the test error advises: _Stubs directories provider doesn't contain 'snappy'. Please add 'snappy' to 'PhpCoreStubsProvider::StubDirectoryMap'_

PR #1371, opened 2022-04-28
Add return type description for passthru()

cookieguru:


standard/standard_2.php
```diff
@@ -719,7 +719,7 @@ function escapeshellarg(string $arg): string {}
  * If the return_var argument is present, the
  * return status of the Unix command will be placed here.
  * </p>
- * @return bool|null
+ * @return bool|null null on success or false on failure.
```
williamdes, line 722:
```suggestion
 * @return false|null null on success or false on failure.
```

cookieguru:
Why does [ReflectionFunctionAbstract::getReturnType()](https://www.php.net/manual/en/reflectionfunctionabstract.getreturntype) return [`null|bool`](https://github.com/JetBrains/phpstorm-stubs/runs/6221142968?check_suite_focus=true#step:6:1230)?  I can't find the definition in the PHP src, but there is [this which declares `false|null`](https://github.com/php/php-src/blob/master/ext/standard/basic_functions.stub.php#L1090)

DmitryTronin:
Yeah, sadly I'm not proficient in C. The function is defined [here](https://github.com/php/php-src/blob/master/ext/standard/exec.c#L267), being basically [php_exec_ex function](https://github.com/php/php-src/blob/master/ext/standard/exec.c#L206) with mode `3`. Unfortunately I can't clearly grasp the return types here & instead I would suggest to stick to the return values returned by the runtime tests.
As for the `null|false` in php-src stubs - that notation [can't be used](https://www.php.net/manual/en/language.types.declarations.php#language.types.declarations.composite) in the runtime, see "Caution" section at `false pseudo-type`.
Let me know if you think otherwise or have any other info to add.

PR #1381, opened 2022-05-05
MongoDB Serializable interface is not deprecated

lku:
I don't believe this interface is deprecated. 

I think it was marked by mistake. Maybe confused with one in the core (different namespace), which is deprecated in PHP 8.1


DmitryTronin (approved):
Indeed, also nothing at https://github.com/mongodb/mongo-php-driver/blob/master/src/BSON/Serializable.c

PR #1384, opened 2022-05-16
fix: Redis::mset should be mSet

taka-oyama:
I believe this is a typo.

see https://github.com/phpredis/phpredis#mset-msetnx


DmitryTronin (approved):

taka-oyama:
I've rebased it with master and pushed again.

PR #1388, opened 2022-05-18
Fix: `Ds\Collection` should extend `IteratorAggregate`.

ZebulanStanphill:
Discovered in https://github.com/phpstan/phpstan/issues/7258.

`Collection` extends `\IteratorAggregate` in both [the `php-ds/php-ds` polyfill](https://github.com/php-ds/polyfill/blob/master/src/Collection.php) and [the `ext-ds` extension](https://github.com/php-ds/ext-ds/blob/master/src/php/classes/php_collection_ce.c), so the stubs here should match.

Note that `Traversable` is removed from the `extends` because `IteratorAggregate` already extends `Traversable`.

### DIFF
--- a/ds/ds.php
+++ b/ds/ds.php
@@ -11,6 +11,7 @@
 namespace Ds;
 
     use Countable;
+    use IteratorAggregate;
     use JsonSerializable;
     use OutOfBoundsException;
     use OutOfRangeException;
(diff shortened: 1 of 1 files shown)


DmitryTronin (approved):
Confirmed on runtime.

DmitryTronin:
You might want to also submit a pr at https://github.com/php/doc-en to reflect this in the docs. If that's a bit of an overkill for you - I can do that a bit later.

ZebulanStanphill:
@DmitryTronin I already did: https://github.com/php/doc-en/pull/1587

DmitryTronin:
Ah thanks, that's interesting. Appreciated!

PR #1389, opened 2022-05-18
Add missing `Ds\PriorityQueue#peek` method.

ZebulanStanphill:
Discovered in https://github.com/phpstan/phpstan/issues/7232.

https://www.php.net/manual/en/ds-priorityqueue.peek.php


staabm:
Thanks!

DmitryTronin (approved):
Confirmed on runtime

PR #1390, opened 2022-06-05
Fix what seems to be a minor typo

alexandreelise:
It seems that there is a small typo from "sift up" to "shift up". Just my 2 cents. Trying to help as much as I can.
Take care Jetbrains team.

### DIFF
--- a/SPL/SPL_c1.php
+++ b/SPL/SPL_c1.php
@@ -1699,7 +1699,7 @@ class SplPriorityQueue implements Iterator, Countable
     public function top(): mixed {}
 
     /**
-     * Extracts a node from top of the heap and sift up.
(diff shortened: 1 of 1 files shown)


DmitryTronin:
@alexandreelise 
Apparently, this is not a typo: https://github.com/php/doc-en/commit/d06666a2f71a703358c95df2b26cd162499c0add

alexandreelise:
@DmitryTronin you are right. According to what seems the authority https://xlinux.nist.gov/dads/HTML/siftup.html . I should have checked that before. Take care. Have a great day you and all Jetbrains team. And thanks again for your awesome products I use them all day. Especially PhpStorm. See ya.

DmitryTronin:
Thanks for sharing this & have a great day as well 😊

PR #1394, opened 2022-06-15
Update date_c.php

WinterSilence:
> DateInterval объект представляет разницу между двумя датами или FALSE в случае возникновения ошибки.

### DIFF
--- a/date/date_c.php
+++ b/date/date_c.php
@@ -344,7 +344,7 @@ class DateTimeImmutable implements DateTimeInterface
      * @link https://secure.php.net/manual/en/datetime.diff.php
      * @param DateTimeInterface $targetObject <p>The date to compare to.</p>
      * @param bool $absolute [optional] <p>Should the interval be forced to be positive?</p>
-     * @return DateInterval
+     * @return DateInterval|false
      * The {@link https://secure.php.net/manual/en/class.dateinterval.php DateInterval} object representing the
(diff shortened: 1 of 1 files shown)


date/date_c.php
```diff
@@ -344,7 +344,7 @@ public function sub(DateInterval $interval): DateTimeImmutable {}
      * @link https://secure.php.net/manual/en/datetime.diff.php
      * @param DateTimeInterface $targetObject <p>The date to compare to.</p>
      * @param bool $absolute [optional] <p>Should the interval be forced to be positive?</p>
-     * @return DateInterval
+     * @return DateInterval|false
```
williamdes, line 347: Please document why it can be false

WinterSilence: @williamdes https://www.php.net/manual/en/datetime.diff.php#refsect1-datetime.diff-returnvalues

WinterSilence:
@DmitryTronin new hell to parsing (:
https://github.com/php/doc-en/blob/b08895614bbc3bfb0273a41c29eda921161538c7/reference/datetime/datetimeinterface/diff.xml#L66

WinterSilence:
@DmitryTronin sorry, last news: in 8+ ALL functions and methods returning false, return null. https://github.com/php/doc-en/issues/1627

DmitryTronin:
Thanks for sharing, we'll look into this.

PR #1397, opened 2022-06-24
Fix incorrect documentation notation a to an

skh6075:
I found and corrected the incorrect notation while checking the documentation.


DmitryTronin (changes requested):
.idea/.gitignore isn't required since there's already a .gitignore file in a root of a project.

DmitryTronin:
Sorry, the changes are too minor to be included. Plus, original documentation comes with `a` instead of `an`.

PR #1407, opened 2022-07-18
Bulk 8.2alpha fixes

DmitryTronin:
Fixed suite naming and version in tests/Model/PhpVersions.php


DmitryTronin:
On pause till https://youtrack.jetbrains.com/issue/WI-67698

Reflection/ReflectionEnum.php
```diff
@@ -33,5 +35,6 @@ public function isBacked(): bool {}
     /**
      * @return ReflectionType|null
      */
+    #[LanguageLevelTypeAware(['8.2' => 'null|ReflectionNamedType'], default: 'null|ReflectionType')]
     public function getBackingType(): ?ReflectionType {}
```
isfedorov, line 39: @DmitryTronin Please remove ?ReflectionType from signature as far as you declare type in attribute

SPL/SPL_c1.php
```diff
@@ -998,6 +999,7 @@ public function hasChildren(): bool {}
      * @return null|RecursiveIterator An SplFileObject does not have children so this method returns NULL.
      */
     #[TentativeType]
+    #[LanguageLevelTypeAware(['8.2' => 'null|null'], default: 'null|RecursiveIterator')]
```
isfedorov, line 1002: '8.2' => 'null|null' :) kek... As I understand this is due to unfinished implementation of PHP 8.2 and I hope it will be fixed but still..kek

date/date_c.php
```diff
@@ -947,6 +949,7 @@ class DatePeriod implements IteratorAggregate
     /**
      * @since 8.2
      */
+    #[Immutable]
     public const INCLUDE_END_DATE = 2;
```
isfedorov, line 953: Why Immutable? It's already constant.

tests/TestData/mutedProblems.json
```diff
@@ -987,6 +987,17 @@
           ]
         }
       ]
+    },
+    {
+      "name": "sodium_crypto_stream_xchacha20_xor_ic",
```
isfedorov, line 992: Can we just mark this function as removed since 8.2 instead of mute the problem?

phpunit.xml.dist
```diff
@@ -76,7 +77,7 @@
             <file>tests/StubsForbiddenTypeHintsTest.php</file>
             <file>tests/StubsPhp81Tests.php</file>
         </testsuite>
-        <testsuite name="PHP_8.2alpha">
+        <testsuite name="PHP_8.2">
```
isfedorov, line 80: I think that it also makes sense to remove StubsConstantsAndParametersValuesTest from 8.1 suite and leave it only on 8.2 suite since constant values change on every release but we support values check only against the latest release

SPL/SPL_c1.php
DmitryTronin: Yea, that's 🔥

date/date_c.php
DmitryTronin: Thanks, must be a mistake

tests/TestData/mutedProblems.json
DmitryTronin: Not sure how I do this. The problem is that this is a[ new function](https://github.com/php/php-src/blob/1a4401d40aabd9c33fcd122e929f3fbedfbf216f/ext/sodium/libsodium.c#L1585) in 8.2, but reflection returns the function as missing yet.
My plan is to revise this with the next 8.2 release

PR #1413, opened 2022-07-25
Added (!$functions) handling

DmitryTronin:
Added this while tracking down "Attempt to read property "parameters" on null" error.
Happens when \StubTests\Model\BasePHPElement::entitySuitsCurrentPhpVersion returns false when running tests on PHP beta. It is not expected that getenv('PHP_VERSION') would return e.g. PHP8.2_beta:
image

### DIFF
--- a/tests/Model/StubsContainer.php
+++ b/tests/Model/StubsContainer.php
@@ -126,6 +126,11 @@ class StubsContainer
         if (!empty($functions)) {
             return array_pop($functions);
         }
+        // Added this while tracking down "Attempt to read property "parameters" on null" error.
+        // Happens when BasePHPElement::entitySuitsCurrentPhpVersion returns false when running tests on PHP beta: getenv('PHP_VERSION')
+        if (!$functions) {
+            throw new RuntimeException("Could not get function " . $name . " from reflection, possibly unsupported PHP_VERSION in getenv");
+        }
         return null;
     }
 
--- a/tests/Model/StubsContainer.php
+++ b/tests/Model/StubsContainer.php
@@ -97,7 +97,7 @@ class StubsContainer
      * @param string $name
      * @param string|null $sourceFilePath
      * @param bool $shouldSuitCurrentPhpVersion
-     * @return PHPFunction|null
+     * @return PHPFunction
      * @throws RuntimeException
      */
     public function getFunction($name, $sourceFilePath = null, $shouldSuitCurrentPhpVersion = true)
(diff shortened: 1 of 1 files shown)


tests/Model/StubsContainer.php
```diff
@@ -126,6 +126,11 @@ public function getFunction($name, $sourceFilePath = null, $shouldSuitCurrentPhp
         if (!empty($functions)) {
             return array_pop($functions);
         }
+        // Added this while tracking down "Attempt to read property "parameters" on null" error.
```
isfedorov, line 129: Let's remove this comment and write the reason in the commit message.

tests/Model/StubsContainer.php
```diff
@@ -126,6 +126,11 @@ public function getFunction($name, $sourceFilePath = null, $shouldSuitCurrentPhp
...
+        // Happens when BasePHPElement::entitySuitsCurrentPhpVersion returns false when running tests on PHP beta: getenv('PHP_VERSION')
+        if (!$functions) {
+            throw new RuntimeException("Could not get function " . $name . " from reflection, possibly unsupported PHP_VERSION in getenv");
```
isfedorov, line 132: I think concatenation can be replaced with interpolation here since it's just one simple variable

tests/Model/StubsContainer.php
```diff
@@ -126,6 +126,11 @@ public function getFunction($name, $sourceFilePath = null, $shouldSuitCurrentPhp
         if (!empty($functions)) {
             return array_pop($functions);
         }
+        // Added this while tracking down "Attempt to read property "parameters" on null" error.
+        // Happens when BasePHPElement::entitySuitsCurrentPhpVersion returns false when running tests on PHP beta: getenv('PHP_VERSION')
+        if (!$functions) {
```
isfedorov, line 131: I think we can move this check to the `else` branch of the `if (!empty($functions))` above

tests/Model/StubsContainer.php
```diff
@@ -126,6 +126,11 @@ public function getFunction($name, $sourceFilePath = null, $shouldSuitCurrentPhp
...
+            throw new RuntimeException("Could not get function " . $name . " from reflection, possibly unsupported PHP_VERSION in getenv");
+        }
         return null;
```
isfedorov, line 134: If we will move check to the `else` branch according to the comment above `return null` could be removed as redundant and return type in declaration also can be changed to non-nullable

isfedorov (changes requested):

tests/Model/StubsContainer.php
```diff
@@ -125,8 +125,9 @@ public function getFunction($name, $sourceFilePath = null, $shouldSuitCurrentPhp
         }
         if (!empty($functions)) {
             return array_pop($functions);
+        } else {
+            throw new RuntimeException("Could not get function {$name} from reflection, possibly unsupported PHP_VERSION in getenv");
```
isfedorov, line 129: Minor: Curly braces are unnecessary here. Other than that PR is good to go.

isfedorov (approved):

tests/Model/StubsContainer.php
DmitryTronin: Thanks, noted. Will try to address within one of the next PRs.

PR #1416, opened 2022-07-26
[phpstorm-stubs] added missed sodium functions

isfedorov:


sodium/sodium.php
```diff
@@ -1129,6 +1132,9 @@ function sodium_crypto_secretstream_xchacha20poly1305_keygen(): string {}
  */
 function sodium_crypto_secretstream_xchacha20poly1305_init_push(string $key): array {}
 
+#[PhpStormStubsElementAvailable('8.2')]
```
DmitryTronin, line 1135: That's an [old function](https://www.php.net/manual/en/function.sodium-crypto-secretstream-xchacha20poly1305-push.php), it's `@since 7.2`

sodium/sodium.php
```diff
@@ -1139,6 +1145,10 @@ function sodium_crypto_secretstream_xchacha20poly1305_init_push(string $key): ar
  */
 function sodium_crypto_secretstream_xchacha20poly1305_init_pull(string $header, string $key): string {}
 
+/** @return array<int, int|string>|false */
+#[PhpStormStubsElementAvailable('8.2')]
```
DmitryTronin, line 1149: This one is also simply `@since 7.2`

DmitryTronin (changes requested):

DmitryTronin (approved):

PR #1423, opened 2022-07-29
Added testClassesReadonly test & related adjustments

DmitryTronin:
They are: 

- testClassesReadonly  provider, 
- isReadonly PHPClass property, 
- removed nonexistent isReadonly ReflectionProperty property, 
- added CLASS_READONLY StubProblemType

### DIFF
--- a/Reflection/ReflectionProperty.php
+++ b/Reflection/ReflectionProperty.php
@@ -30,13 +30,6 @@ class ReflectionProperty implements Reflector
     public $class;
 
     /**
-     * @var bool
-     * @since 8.1
-     */
-    #[Immutable]
-    public bool $isReadonly;
-
-    /**
      * Indicates that the property is static.
      *
      * @link https://www.php.net/manual/en/class.reflectionproperty.php#reflectionproperty.constants.is-static
--- a/tests/Model/PHPClass.php
+++ b/tests/Model/PHPClass.php
@@ -40,7 +40,7 @@ class PHPClass extends BasePHPClass
         }
         $this->interfaces = $reflectionObject->getInterfaceNames();
         $this->isFinal = $reflectionObject->isFinal();
-        if (method_exists($reflectionObject,'isReadOnly')) {
+        if (method_exists($reflectionObject, 'isReadOnly')) {
             $this->isReadonly = $reflectionObject->isReadOnly();
         }
         foreach ($reflectionObject->getMethods() as $method) {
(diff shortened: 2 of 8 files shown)


Reflection/ReflectionProperty.php
```diff
@@ -29,13 +29,6 @@ class ReflectionProperty implements Reflector
     #[LanguageLevelTypeAware(['8.1' => 'string'], default: '')]
     public $class;
 
-    /**
-     * @var bool
```
isfedorov, line 33: Why have you removed `readonly` of property?

tests/TestData/Providers/Reflection/ReflectionClassesTestDataProviders.php
```diff
@@ -64,4 +64,17 @@ public static function finalClassesProvider(): ?Generator
             yield "class $class->name" => [$class];
         }
     }
+
+        public static function readonlyClassesProvider(): ?Generator
+    {
+        $classes = ReflectionStubsSingleton::getReflectionStubs()->getClasses();
+        $filtered = EntitiesFilter::getFiltered(
+            $classes,
+            null,
```
isfedorov, line 73: If I'm not mistaken,  some closure that will filter out non-readonly classes should be here instead of `null`. like `fn(PhpClass $class) => $class->readonly === false`

tests/Model/StubProblemType.php
```diff
@@ -31,4 +31,5 @@ interface StubProblemType
     const WRONG_RETURN_TYPEHINT = 24;
     const WRONG_OPTIONALLITY = 27;
     const PROPERTY_READONLY = 28;
+    const CLASS_READONLY = 29;
```
isfedorov, line 34: I'd actually rename CLASS_READONLY and PROPERTY_READONLY to just one constant `WRONG_READONLY` and use it for properties and for classes

tests/Model/PHPClass.php
```diff
@@ -24,6 +24,9 @@ class PHPClass extends BasePHPClass
     /** @var PHPProperty[] */
     public $properties = [];
 
+    /** @since 8.2 */
```
isfedorov, line 27: `since` can be removed here because this class is a part of model and `$isReadonly` here is just a property of an object and isn't related with php version

isfedorov (changes requested):

Reflection/ReflectionProperty.php
isfedorov: Oops, sorry for the noise. You are completely right. Probably we need a test to check such cases at the end.

DmitryTronin: Added this to the list

PR #1428, opened 2022-08-11
Fix 'key' typo in gnupg.php stub

muli:


DmitryTronin:
Sorry, these are not typos, see https://github.com/JetBrains/phpstorm-stubs/pull/1288

PR #1429, opened 2022-08-15
string|true fixes for print_r based on team discussion

DmitryTronin:
Also musing "wrong return typehint" for that

### DIFF
--- a/tests/TestData/Providers/Reflection/ReflectionFunctionsProvider.php
+++ b/tests/TestData/Providers/Reflection/ReflectionFunctionsProvider.php
@@ -17,6 +17,17 @@ class ReflectionFunctionsProvider
         }
     }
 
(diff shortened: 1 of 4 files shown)


tests/TestData/Providers/Reflection/ReflectionFunctionsProvider.php
```diff
@@ -17,6 +17,17 @@ public static function allFunctionsProvider(): ?Generator
         }
     }
 
+    public static function functionsForTypeHintsTestProvider(): ?Generator
```
isfedorov, line 20: Maybe `functionsForReturnTypeHintTestsProvider` to be more precise?

isfedorov (approved):
Please rename the provider method to be more precise. Other than that everything is good to go

PR #1450, opened 2022-10-10
https://youtrack.jetbrains.com/issue/WI-69083

DmitryTronin:
Changed `key_type` type & default value & expected values according to sources


isfedorov:
Fixed in #1486 1486

PR #1480, opened 2022-11-06
fix: `$position` parameter to `lInsert()` is string

tlamy:
Even the documentation suggests to use \RedisCluster::BEFORE (or ::AFTER), which are strings (`before`, `after`).


isfedorov:
@tlamy Thank you for the contribution!

PR #1491, opened 2022-11-11
[phpstorm-stubs] update stubs to migration version for php 8.2

isfedorov:


lptn:
Hey @isfedorov 
thanks for insights from this PR! I'm going to apply some type changes (mostly from `bool` to `true` or `false`) on Psalm, but have a question for you.

I see you used `true` or `false` return types even in cases where the doc block and official PHP docs explicitly says like "true on success or false on failure".  Example:
```php
/*
 * @return true <b>TRUE</b> on success or <b>FALSE</b> on failure.
 */
function xml_set_external_entity_ref_handler(XMLParser $parser, $handler): true {}
```


What is the correct return type in such cases?

Thanks!

isfedorov:
@lptn please ignore this PR :) This is experimental work. Answering your question regarding return type: xml_set_external_entity_ref_handler returns only true since PHP 8.2 https://3v4l.org/fnFO5#v8.2rc7

PR #1494, opened 2022-11-19
Update 404 link for stream constants

JoshuaBehrens:
I know the constant `STREAM_MUST_SEEK` below the link comment is only written there in the comments of the linked docs page but at least it is a page mentioning it and that is not 404


isfedorov:
@JoshuaBehrens Thank you for the contribution! Btw, probably it makes sense to make a CR to https://github.com/php/doc-en to add this constant to the help

JoshuaBehrens:
Thank you for merging.
Certainly! But for that someone with more php internal knowledge would be more suitable :D I am fine to debug the php-src but I am not sure whether I am the right person to do that

PR #1497, opened 2022-11-30
Fixed wording for nl2br

AndrewDawes:
Wording didn't make sense, updated wording to match official php docs


isfedorov:
@AndrewDawes Thank you for the contribution!

PR #1498, opened 2022-12-01
Updated wording to match php docs

AndrewDawes:
Updated some param and return descriptions for some functions, to better match PHP docs :)


standard/standard_0.php
```diff
@@ -177,7 +177,7 @@ function bin2hex(string $string): string {}
  * Delay execution
  * @link https://php.net/manual/en/function.sleep.php
  * @param int $seconds <p>
- * Halt time in seconds.
+ * Halt time in seconds (must be greater than or equal to 0).
```
isfedorov, line 180: This clarification can be replaced with psalm's `int<0,max>` parameter type declaration

PR #1499, opened 2022-12-01
ZipArchive::open(): bool|int

voku:
The documentation is not correct here: (I will try to update it, also) "Returns true on success, or one of the following error codes" | https://www.php.net/manual/en/ziparchive.open.php

- https://github.com/vimeo/psalm/issues/6052


isfedorov:
Thank you @voku !

PR #1500, opened 2022-12-01
Generator template covariant yield\return

Fuck4ik:
Fix \Generator type template from:
psalm - https://github.com/vimeo/psalm/blob/c6551a61e7a62426c0e9c8f92b33cf09e8346db9/stubs/CoreGenericClasses.phpstub#L21
phpstan - https://github.com/phpstan/phpstan/blob/53a2caa820fb5d603c5d623212e7ce421a09019e/website/src/_posts/generics-in-php-using-phpdocs.md#type-safe-iterators-and-generators

### DIFF
--- a/standard/_types.php
+++ b/standard/_types.php
@@ -149,14 +149,14 @@ namespace {
     function PS_UNRESERVE_PREFIX_eval($code) {}
 
     /**
-     * @template TKey of array-key
-     * @template TSend
-     * @template TReturn
-     * @template TYield
-     *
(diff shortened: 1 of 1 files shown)


Fuck4ik:
Example my problem:
```php
<?php

namespace App;

class Foo {
    public function bar(): int {
        return 1;
    }
}

/**
 * @return \Generator<array-key, Foo>
 */
$generator = static function(): \Generator {
    yield new Foo();
};

foreach ($generator() as $item) {
    $item-> // IDE fail suggestion
}
```

PR #1509, opened 2023-01-04
Change `implode` signature according to php.net

RSickenberg:
https://www.php.net/manual/en/function.implode.php

- Removed array for the `$separator`. This is no more supported since PHP 8.0.
- `$array` should not be nullable.

### DIFF
--- a/standard/standard_1.php
+++ b/standard/standard_1.php
@@ -1002,37 +1002,37 @@ function explode(string $separator, string $string, int $limit) {}
 /**
  * Join array elements with a string
  * @link https://php.net/manual/en/function.implode.php
- * @param array|string  $separator [optional]<p>
+ * @param string  $separator [optional]<p>
  * Defaults to an empty string. This is not the preferred usage of
  * implode as glue would be
  * the second parameter and thus, the bad prototype would be used.
  * </p>
- * @param array|null $array <p>
+ * @param array $array <p>
  * The array of strings to implode.
  * </p>
  * @return string a string containing a string representation of all the array
  * elements in the same order, with the glue string between each element.
  */
 #[Pure]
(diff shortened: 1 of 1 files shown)


standard/standard_1.php
```diff
@@ -1002,37 +1002,37 @@ function explode(string $separator, string $string, int $limit) {}
 /**
  * Join array elements with a string
  * @link https://php.net/manual/en/function.implode.php
- * @param array|string  $separator [optional]<p>
+ * @param string  $separator [optional]<p>
```
RSickenberg, line 1005:
```suggestion
 * @param string $separator [optional]<p>
```

standard/standard_1.php
```diff
@@ -1002,37 +1002,37 @@ function explode(string $separator, string $string, int $limit) {}
...
 #[Pure]
-function implode(array|string $separator = "", ?array $array): string {}
+function implode(string $separator = "", array $array): string {}
 
 /**
  * Alias:
  * {@see implode}
  * @link https://php.net/manual/en/function.join.php
- * @param array|string  $separator [optional] <p>
+ * @param string  $separator [optional] <p>
```
RSickenberg, line 1023:
```suggestion
 * @param string $separator [optional] <p>
```

RSickenberg (commented):
This may also be fixed

isfedorov:
@RSickenberg unfortunately such change will not work. It will produce false positive errors when declare(strict_types=1)
Screenshot 2023-01-05 at 21 47 21

isfedorov:
> Removed array for the $separator. This is no more supported since PHP 8.0.

Unfortunately stubs at the current implementation should suit all the versions of PHP but not only the latest one

PR #1510, opened 2023-01-05
fix Phar::offsetGet return type

bmewburn:
https://www.php.net/manual/en/phar.offsetget.php

Signature in docs suggests that it's SplFileInfo but all the other documentation indicates it should be PharFileInfo


bmewburn:
nevermind
https://github.com/php/php-src/commit/3fc3cfbad01a1046fc13ac864fa82bc83b34efcd#diff-c076bfd33f8848d4371bddb15aafc0ad1232a3db55dfe6cbc198f2d59dd0e0b5

PR #1511, opened 2023-01-06
Templatize ArrayObject’s stub

MatTheCat:
Attempt at [WI-70459](https://youtrack.jetbrains.com/issue/WI-70459).

### DIFF
--- a/SPL/SPL.php
+++ b/SPL/SPL.php
@@ -1547,7 +1547,7 @@ class ArrayObject implements IteratorAggregate, ArrayAccess, Serializable, Count
      * @link https://php.net/manual/en/arrayobject.construct.php
      * @param array<TKey, TValue>|object $array The input parameter accepts an array or an Object.
      * @param int $flags Flags to control the behaviour of the ArrayObject object.
-     * @param class-string<ArrayIterator<TKey,TValue>> $iteratorClass Specify the class that will be used for iteration of the ArrayObject object. ArrayIterator is the default class used.
+     * @param class-string<ArrayIterator> $iteratorClass Specify the class that will be used for iteration of the ArrayObject object. ArrayIterator is the default class used.
      */
     public function __construct(
         #[LanguageLevelTypeAware(['8.0' => 'object|array'], default: '')] $array = [],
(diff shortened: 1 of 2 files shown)


isfedorov:
Tests fail due to tag `@param callable(TValue, TValue):int`. We use `"phpdocumentor/reflection-docblock"` to parse PHPDoc tags and phpdocumentor can't parse such custom tag. That's why parsing of such tag should be processed separately. Please see `\StubTests\Model\PHPDocElement::collectTags` and update method so it change `@param callable(TValue, TValue):int` to `@param callable` (like it's done for `int<a,b>` with just replacing it to `preg_replace("/int\<\w+,\s*\w+\>/", "int", $text);`

MatTheCat:
@isfedorov thanks; is https://github.com/JetBrains/phpstorm-stubs/pull/1511/commits/c482b1ee0c2bd717f7a50bbbfb0a8dde15069c60 okay?

isfedorov:
@MatTheCat yes, but unfortunately, it's still not enough. The same trick should be done for other custom type declarations like `ArrayIterator<TKey,TValue>` and `array<Tkey, TValue>`. Could you please update `\StubTests\Model\PHPDocElement::collectTags` accordingly?

MatTheCat:
Hmm I see such types are already present in the codebase, like https://github.com/JetBrains/phpstorm-stubs/blob/ccff41f1a6b40ff4ecfaa4964fa5007fa78f6225/Core/Core_c.php#L798

Are you sure `PHPDocElement` has to be updated here?

isfedorov:
@MatTheCat Oh, you are right. Sorry, I gave you the wrong spot. The problem is with `class-string`. phpdocumentor fails to parse this tag and it should be changed to `string`

MatTheCat:
I see `class-string` is already used as well: https://github.com/JetBrains/phpstorm-stubs/blob/b543df50f5d71e4350b9e09f6758ff08f476d50b/Reflection/ReflectionParameter.php#L284

Now I don’t understand :sweat_smile:

isfedorov:
@MatTheCat well..this is how the `"phpdocumentor/reflection-docblock"` works =\ It can parse simple cases like `class-string<T>`
Screenshot 2023-01-09 at 11 48 54
But can't parse complex cases like `class-string<ArrayIterator<TKey,TValue>>`
Screenshot 2023-01-09 at 11 50 10

isfedorov:
@MatTheCat in the latest changes you missed the change for `callable` that you've done before. It's still required

MatTheCat:
Daaaamn, updated :crossed_fingers:

isfedorov:
Fixed in #1517 1517

PR #1512, opened 2023-01-07
Allow DNF types in ReflectionUnionType::getTypes()

BenMorel:
Since [DNF types](https://wiki.php.net/rfc/dnf_types) have been introduced in PHP 8.2, types returned by `ReflectionUnionType::getTypes()` include `ReflectionIntersectionType` in addition to `ReflectionNamedType`:

https://3v4l.org/ZFKZ6#v8.2.1

I'm not entirely sure of the syntax in this patch, suggestions welcome.

### DIFF
--- a/Reflection/ReflectionUnionType.php
+++ b/Reflection/ReflectionUnionType.php
@@ -1,5 +1,6 @@
 <?php
 
+use JetBrains\PhpStorm\Internal\LanguageLevelTypeAware;
 use JetBrains\PhpStorm\Pure;
 
 /**
@@ -8,10 +9,16 @@ use JetBrains\PhpStorm\Pure;
 class ReflectionUnionType extends ReflectionType
 {
     /**
-     * Get list of named types of union type
+     * Get list of types of union type
      *
-     * @return ReflectionNamedType[]
+     * @return ReflectionNamedType[]|ReflectionIntersectionType[]
      */
     #[Pure]
+    #[LanguageLevelTypeAware(
+        [
+            '8.2' => 'ReflectionNamedType[]|ReflectionIntersectionType[]'
+        ],
+        default: 'ReflectionNamedType[]'
+    )]
     public function getTypes(): array {}
 }


isfedorov:
@BenMorel Thank you for the PR! But I'd change the custom (psalm) type declaration `(ReflectionNamedType|ReflectionIntersectionType)[]` to the one that is supported by https://docs.phpdoc.org/guide/guides/types.html (kind of default standard) `ReflectionNamedType[]|ReflectionIntersectionType[]` since stubs can be used by some other tools the latter style will be better for tools compatibility. Test fail isn't related to your change but is related to the new function `imap_is_open` so please ignore it.

BenMorel:
@isfedorov I'll change it if you think it has better compatibility, but they actually have slightly different meaning:

- `ReflectionNamedType[]|ReflectionIntersectionType[]` would mean that all objects in the array are of the same type (i.e. an array of `ReflectionNamedType` **or** an array of `ReflectionIntersectionType`)
- `(ReflectionNamedType|ReflectionIntersectionType)[]` accepts both types simultaneously in the same array

Also, is the `#[LanguageLevelTypeAware]` attribute OK? I wasn't sure about the meaning of the `default` value.

isfedorov:
@BenMorel 
>  but they actually have slightly different meaning:

That's true, but I'm still not sure if it's a good way to use custom psalm tags in stubs (despite the fact that we use them in some places for template annotations) so let's align with phpdocumentor at the moment and possibly will change it later.

> Also, is the #[LanguageLevelTypeAware] attribute OK? I wasn't sure about the meaning of the default value.

Since `ReflectionUnionType` has `since 8.0`, the attribute can be changed to `#[LanguageLevelTypeAware(['8.2' => 'ReflectionNamedType[]|ReflectionIntersectionType[]'], default: 'ReflectionNamedType[]')]`. In this way it will say that since 8.2 return type is `'ReflectionNamedType[]|ReflectionIntersectionType[]` and for other versions (8.0, 8.1) it is `ReflectionNamedType[]`

BenMorel:
@isfedorov Understood. Done!

PR #1513, opened 2023-01-09
hrtime(): $as_number is optional

mabar:
$as_number was always optional, see https://3v4l.org/uK5ro and https://www.php.net/manual/en/function.hrtime.php

### DIFF
--- a/standard/standard_4.php
+++ b/standard/standard_4.php
@@ -384,7 +384,7 @@ function highlight_string(string $string, bool $return = false): string|bool {}
  */
 #[Pure(true)]
 function hrtime(
-    #[PhpStormStubsElementAvailable(from: '7.3', to: '7.4')] bool $as_number,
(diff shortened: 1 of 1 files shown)


isfedorov:
Hm, that's interesting. ReflectionFunction::isOptional gives another result https://3v4l.org/ekJNn

mabar:
https://3v4l.org/576CC

> Cannot determine default value for internal functions

There always was a default value, but since PHP 8 it is possible to check it. 

Check changelog of [getDefaultValue](https://www.php.net/manual/en/reflectionparameter.getdefaultvalue.php)

isfedorov:
@mabar Ok! Let's remove `PhpStormStubsElementAvailable` attributes in the signature and leave only `bool $as_number = false` then

isfedorov:
@mabar Thank you for the contribution!

PR #1522, opened 2023-01-25
`$iterator` in Redis scan() can be also null

RobiNN1:
See https://github.com/phpredis/phpredis/blob/develop/redis.stub.php#L2915

It might be better to sync the whole stub. Phpredis devs finally updated it and added the missing docblocks.

### DIFF
--- a/redis/Redis.php
+++ b/redis/Redis.php
@@ -4723,9 +4723,9 @@ class Redis
     /**
      * Scan the keyspace for keys
      *
-     * @param int    &$iterator Iterator, initialized to NULL.
-     * @param string $pattern  Pattern to match.
-     * @param int    $count    Count of keys per iteration (only a suggestion to Redis).
+     * @param int|null &$iterator Iterator, initialized to NULL.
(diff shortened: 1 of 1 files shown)


isfedorov:
@RobiNN1 Thank you for the PR! 
> It might be better to sync the whole stub.

That's a great idea!

RobiNN1:
I'll update the whole stub later, just wondering, can I use native types and union types or does everything have to be only in docblocks?

isfedorov:
@RobiNN1 
> just wondering can I use native types and union types or does everything have to be only in docblocks?

Yes, you can but this should be done carefully to avoid hierarchy problems in the future. Let's suppose there is a class `A` with method `b`, and you add scalar type hints for this method. Then some user with PHP 5.6 overrides class `A` and gets hierarchy checks problems since 5.6 doesn't support scalar type hints. So I'd say that you can safely add native type hints for the final methods or methods in final classes since they can't be overridden or if the `since` version of the class supports that native typehints

PR #1524, opened 2023-02-02
Split MongoDB stubs into separate files

alcaeus:
Since the MongoDB extension uses stub files to generate arginfo, having classes in individual files will make it easier for us to keep them up-to-date in the future.

No changes to classes or signatures has been made in this PR.


isfedorov:
@alcaeus Please call `docker-compose -f docker-compose.yml run test_runner composer cs-fix` to fix cs-fixer problems

PR #1525, opened 2023-02-03
Sync redis stub with official file.

RobiNN1:
As I mentioned in my previous PR, here is the synced full stub file. I tried to reduce the changes to make diff smaller, but there are still a lot of changes.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
     "php": "^8.0",
     "nikic/php-parser": "@stable",
     "phpdocumentor/reflection-docblock": "@stable",
-    "phpunit/phpunit": "@stable",
+    "phpunit/phpunit": "^9.6",
     "friendsofphp/php-cs-fixer": "@stable"
   },
   "autoload": {
--- a/redis/Redis.php
+++ b/redis/Redis.php
@@ -5685,6 +5685,7 @@ class Redis
      *                                depending on whether any elements could be popped within the specified timeout.
      *
      * NOTE: If Redis::OPT_NULL_MULTIBULK_AS_NULL is set to true via Redis::setOption(), this method will instead return NULL when Redis doesn't pop any elements.
+     * @since phpredis 6.0
      */
     public function bzmpop($timeout, $keys, $from, $count = 1) {}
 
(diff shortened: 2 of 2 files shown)


RobiNN1:
Tests are not ready for new PHPUnit 10?

isfedorov:
@RobiNN1 hard to say. Failure says `
Fatal error: Allowed memory size of 1073741824 bytes exhausted (tried to allocate 483328 bytes)` so debug required

RobiNN1:
https://github.com/RobiNN1/phpstorm-stubs/actions/runs/4088195738/jobs/7049600548

Looks like it fails because of new methods that were added in phpredis 6.0 (not released yet)

isfedorov:
No, the problem is that method `Redis::blmove` contains a kind of invalid PHPDoc. We use `"phpdocumentor/reflection-docblock"` to parse PHPDoc and check it for valid tags, and I think that the library tries to read `@redis` in 
```
     * @example
     * @redis->lPush('numbers', 'one');
     * @redis->blmove('numbers', 'odds', Redis::LEFT, Redis::LEFT 1.0);
     * // This call will block, if no additional elements are in 'numbers'
     * @redis->blmove('numbers', 'odds', Redis::LEFT, Redis::LEFT, 1.0);
```
as a tag and fails here. So I'd advise just updating that part and, for example, quoting those expressions

RobiNN1:
Oh didn't noticed this, and actually it's not my typo https://github.com/phpredis/phpredis/blob/develop/redis.stub.php#L1996

RobiNN1:
Latest changes but with phpunit 9 https://github.com/RobiNN1/phpstorm-stubs/commit/3712a46fe6d2fedfa8dd6bbc6b567c66ad322e91

isfedorov:
@RobiNN1 Does PhpUnit 9 solves the OOM problem? But I don't see composer changes in changed files.

RobiNN1:
I've tested in a separate branch https://github.com/RobiNN1/phpstorm-stubs/tree/test2 (to see if it's really a phpunit issue) so it's not included in this PR.

I haven't tried phpunit 10 yet, so I don't know what has changed.

isfedorov:
@RobiNN1 Could you please update composer.json in your PR as well for now? I'll take a look at the problem with PHPUnit 10 later, but for now I don't want to freeze you PR due to this problem.

RobiNN1:
@isfedorov done

PR #1527, opened 2023-02-08
DOMElement::$attributes is not nullable

Jean85:
I got an inconsistency here between PHPStan and Psalm, and at first I though Psalm was wrong: https://github.com/vimeo/psalm/issues/9242

But, after reading the doc description, they are right and this PR fixes it on this side.

### DIFF
--- a/dom/dom_c.php
+++ b/dom/dom_c.php
@@ -1641,6 +1641,14 @@ class DOMElement extends DOMNode implements DOMParentNode, DOMChildNode
     public $nextSibling;
 
     /**
+     * @var DOMNamedNodeMap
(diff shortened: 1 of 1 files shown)


isfedorov:
@Jean85 this property inherited from DOMNode, probably it should be updated there instead.

Jean85:
No that would be incorrect, because in the parent it IS nullable. It's not nullable only in the child class.

isfedorov:
> because in the parent it IS nullable. 
Sorry, but are you sure? see the link I've provided https://github.com/php/php-src/blob/master/ext/dom/php_dom.stub.php#L324 as well as https://www.php.net/manual/en/class.domnode.php#domnode.props.attributes `or null otherwise.` Do I miss something?

isfedorov:
Ah, sorry, my bad. Yeah, that's interesting, despite the property is inherited it can't contain nulls..ok, let's merge it.

PR #1528, opened 2023-02-14
Remove iterable pseudotype

ondrejmirtes:
This isn't a real interface. Because of this stub PhpStorm currently doesn't complain about such code:

Screenshot 2023-02-14 at 22 46 09

But PHP ends with fatal error: https://3v4l.org/UYkWD

The RFC https://wiki.php.net/rfc/iterable states:

> This RFC proposes a new iterable pseudo-type. This type is analogous to callable, accepting multiple types instead of one single type. 

jetbrains/phpstorm-stubs do not contain "interface callable" either...

### DIFF
--- a/Core/Core_c.php
+++ b/Core/Core_c.php
@@ -14,11 +14,6 @@ use JetBrains\PhpStorm\Pure;
 class stdClass {}
 
 /**
- * @link https://wiki.php.net/rfc/iterable
- */
-interface iterable {}
-
-/**
  * Interface to detect if a class is traversable using foreach.
  * Abstract base interface that cannot be implemented alone.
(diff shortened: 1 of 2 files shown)


isfedorov:
@ondrejmirtes Unfortunately, PR can't be merged right away because a lot of functionality (such as type inference, completion, inspections) depends on this type declaration in stubs. The type was added in scope of [WI-32378](https://youtrack.jetbrains.com/issue/WI-32378). I've filled [WI-71397](https://youtrack.jetbrains.com/issue/WI-71397) to fix the problem with not highlighted `implements iterable` (possible in scope of the fix declaration will be removed, but can't say for sure at the moment)

PR #1549, opened 2023-06-08
Update RedisCluster constructor

mstrelan:
As per https://github.com/phpredis/phpredis/blob/develop/redis_cluster.stub.php

### DIFF
--- a/redis/RedisCluster.php
+++ b/redis/RedisCluster.php
@@ -78,10 +78,11 @@ class RedisCluster
      *
      * @param string|null   $name
      * @param array         $seeds
-     * @param float         $timeout
-     * @param float         $readTimeout
+     * @param int|float     $timeout
+     * @param int|float     $readTimeout
      * @param bool          $persistent
-     * @param string|null   $auth
+     * @param mixed         $auth
+     * @param array         $context
      * @throws RedisClusterException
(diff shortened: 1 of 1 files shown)


redis/RedisCluster.php
```diff
@@ -78,10 +78,11 @@ class RedisCluster
...
      * @param string|null   $name
      * @param array         $seeds
-     * @param float         $timeout
-     * @param float         $readTimeout
+     * @param int|float     $timeout
+     * @param int|float     $readTimeout
      * @param bool          $persistent
-     * @param string|null   $auth
+     * @param mixed         $auth
+     * @param array         $context
```
LolGleb, line 85: It seems that the type is supposed to be `array|null`, isn't it?

LolGleb (changes requested):

redis/RedisCluster.php
mstrelan: Looks like this was [only recently updated](https://github.com/phpredis/phpredis/commit/5e4bdf97c887e6b150785629ec59ece7fa0b58e6#diff-38ca794748966b7375aa7255d3776994a5a8a3a1222b39e57fb684a2753104aeL50-R50).

PR #1554, opened 2023-06-22
Update random_int and random_bytes exception thrown

VincentLanglet:
Cf https://www.php.net/manual/en/function.random-int.php#refsect1-function.random-int-changelog
and https://www.php.net/manual/en/function.random-bytes.php#refsect1-function.random-bytes-changelog


isfedorov:
@VincentLanglet Yes, the failures looks unrelated

VincentLanglet:
> @VincentLanglet Yes, the failures looks unrelated

So is this ok to be merged like this @isfedorov or do I have something else to do ?

PR #1555, opened 2023-07-03
Add throws PHPDoc to `constant()`

herndlm:
Since PHP 8 it will throw if the constant is undefined. See https://www.php.net/manual/en/function.constant.php
But this is most likely not the best way to add that info. There is no PHP-version specific throws, right?

UPDATE: found https://youtrack.jetbrains.com/issue/WI-72311 afterwards and adapted the return type description too

### DIFF
--- a/standard/standard_0.php
+++ b/standard/standard_0.php
@@ -158,6 +158,7 @@ class Directory
  * </p>
  * @return mixed the value of the constant, or null if the constant is not
  * defined.
+ * @throws Error If the constant is not defined
  */
 #[Pure(true)]
 function constant(string $name): mixed {}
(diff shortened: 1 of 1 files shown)


herndlm:
@isfedorov sorry for pinging you, but is this the proper way of doing this or is there a better one?

isfedorov:
@herndlm 
> There is no PHP-version specific throws

Theoretically, it's possible to add one more signature of this function (with throws) with internal attribute `PhpStormStubsElementAvailable` for both of functions so that one function to be applied for PHP <8.0 and another for PHP >=8.0 but since 8.0 is the minimal supported PHP version according to https://www.php.net/supported-versions.php and duplicating of signatures isn't really a good way I think this change is OK

PR #1557, opened 2023-07-13
Fix RedisCluster::hSet parameter $value type

szutoutou:


redis/RedisCluster.php
```diff
@@ -246,7 +246,7 @@ public function del($key1, ...$otherKeys) {}
      *
      * @param   string $key
      * @param   int    $ttl
-     * @param   string $value
+     * @param   mix $value
```
LolGleb, line 249: Change to `mixed` please

LolGleb (changes requested):

redis/RedisCluster.php
szutoutou: changed

PR #1560, opened 2023-07-29
mark connection_status() and connection_aborted() impure

staabm:
in chunked requests/response mode, connection status can change in flight

see e.g. https://github.com/sabre-io/http/pull/207

### DIFF
--- a/standard/standard_4.php
+++ b/standard/standard_4.php
@@ -760,7 +760,6 @@ function getallheaders(): false|array {}
  * @link https://php.net/manual/en/function.connection-aborted.php
  * @return int 1 if client disconnected, 0 otherwise.
  */
-#[Pure(true)]
 function connection_aborted(): int {}
 
 /**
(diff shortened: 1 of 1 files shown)


isfedorov:
> in chunked requests/response mode, connection status can change in flight

@staabm That's why the attribute has the parameter `true`. This parameter means that the result of function execution can change depending on the state of the "system". `Pure` attribute by its own means that a function doesn't have side effects. Could you please elaborate on your change? Did you want to say that functions don't have side effects or do you want to say that functions' result can change depends on global state of "system"?

staabm:
thanks for the feedback.

I am doing the PR because I wan't PHPStan to not cache the return-ed value of previous invocations as the result of the function depends on the system

> you want to say that functions' result can change depending on global state of "system"

if I understand correctly, PHPStan should take this parameter into account...?
I need to check whether thats already the case

isfedorov:
@staabm 

> if I understand correctly, PHPStan should take this parameter into account...?

Yes, it should. 

>  I wan't PHPStan to not cache the return-ed value of previous invocations as the result of the function depends on the state of the system

This is precisely what value `true` of parameter is supposed to reflect. So PhpStan should take into account parameter of `Pure` attribute

PR #1561, opened 2023-08-07
Remove compareDocumentPosition

nielsdos:
This was never implemented in PHP, and got removed as a consequence in PHP 8.0. This might be introduced later, but for now the stubs shouldn't contain a method that isn't there.

### DIFF
--- a/dom/dom_c.php
+++ b/dom/dom_c.php
@@ -254,6 +254,7 @@ class DOMNode
 
     /**
      * @param DOMNode $other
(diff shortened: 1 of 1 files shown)


LolGleb:
`compareDocumentPosition` still used in older versions of PHP - https://3v4l.org/QqPSB
It would be better to use `@removed 8.0` in the PHPDoc instead of removing.

nielsdos:
Ok done, I've added the removed annotation

PR #1563, opened 2023-08-12
WI-73896: Fix T_BAD_CHARACTER annotation

lkrms:


LolGleb:
Could you please completely remove the PHPDoc? `T_BAD_CHARACTER` token previously already existed in PHP, but was removed in PHP 7.0. After that it was reintroduced in PHP 7.4. So saying that it is only available since 7.4 not entirely correct.

tokenizer/tokenizer.php
```diff
@@ -191,7 +191,7 @@ function token_name(int $id): string {}
  */
 define('T_FN', 311);
 /**
- * @removed 7.0
+ * @since 7.4
```
LolGleb, line 194: Please, remove the PHPDoc completely

LolGleb (changes requested):

lkrms:
Apologies for the delay, I've made the requested change and rebased.

PR #1566, opened 2023-08-18
[phpstorm-stubs] Tried to fix tests that used to fail for PHP 8.1 and PHP 8.2

LolGleb:
Updated constants, changed return type for WeakMap::getIterator in PHPDoc and added one more parameter to function hash_pbkdf2().

### DIFF
--- a/Core/Core_c.php
+++ b/Core/Core_c.php
@@ -795,7 +795,7 @@ final class WeakMap implements ArrayAccess, Countable, IteratorAggregate
     /**
      * Returns an iterator in the "[object => mixed]" format.
      *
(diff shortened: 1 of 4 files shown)


Core/Core_c.php
```diff
@@ -795,7 +795,7 @@ public function offsetUnset($object): void {}
     /**
      * Returns an iterator in the "[object => mixed]" format.
      *
-     * @return Traversable<TKey, TValue>
+     * @return Iterator
```
isfedorov, line 798: Traversable<TKey, TValue> was using templates intentionally. So the new Type also should have them

PR #1575, opened 2023-08-31
ReflectionProperty::setAccessible is not Pure

TiGR:


isfedorov:
@TiGR Could you please alaborate on your change? According to https://www.php.net/manual/en/reflectionproperty.setaccessible.php `As of PHP 8.1.0, calling this method has no effect`. Why it shouldn't be marked as Pure?

TiGR:
@isfedorov Sorry, my mistake.

PR #1576, opened 2023-09-01
Update Relay.php Redis::setBit

szutoutou:


isfedorov:
@szutoutou could you please elaborate on your change, based on what data you have changed the type of parameter? I'm looking at stub https://builds.r2.relay.so/v0.6.6/relay.stub.php and it still contains `@param  int  $val`

szutoutou:
> @szutoutou您能否根据您更改参数类型的数据详细说明您的更改？我正在查看存根https://builds.r2.relay.so/v0.6.6/relay.stub.php它仍然包含`@param int $val`. 存根的链接由存根作者在[#1514提供](https://github.com/JetBrains/phpstorm-stubs/pull/1514)

Redis/RedisCluster setBit function that param 3 $value can be bool

isfedorov:
> Redis/RedisCluster setBit function that param 3 $value can be bool

but a dedicated stub under `redis/RedisCluster.php` `\RedisCluster::setBit` has such typehint `@param   bool|int $value bool or int (1 or 0)`.

szutoutou:
> 

phpstan will report Parameter #3 $value of method Redis::setBit() expects int, true given.

isfedorov:
@szutoutou could you please provide full code sample to reproduce the problem with PhpStan? Possibly it's a problem with phpstan itself

szutoutou:
> @szutoutou could you please provide full code sample to reproduce the problem with PhpStan? Possibly it's a problem with phpstan itself
test.php
(new Redis())->setBit('a', 1, true);

phpstan analyse test.php -l 5

isfedorov:
@ondrejmirtes could you please check? PHPStan gives an error indeed
Screenshot 2023-09-04 at 10 07 50
Despite stub has declared type as `int|bool`

Screenshot 2023-09-04 at 10 08 59

Screenshot 2023-09-04 at 10 10 45

ondrejmirtes:
@isfedorov I don't get it. What's the expected vs. actual behaviour? Also there's several classes in question - Redis, RedisCluster, Relay\Relay. I'm not sure which one you want me to debug.

Besides jetbrains/phpstorm-stubs PHPStan also uses this file https://github.com/phpstan/phpstan-src/blob/1.11.x/resources/functionMap.php and https://github.com/phpstan/php-8-stubs.

szutoutou:
@ondrejmirtes @isfedorov

szutoutou:
@isfedorov @ondrejmirtes 
maybe the problem is here  
'Redis::setBit' => ['int', 'key'=>'string', 'offset'=>'int', 'value'=>'int'],
[https://github.com/phpstan/phpstan-src/blob/37dbb8cf93547240f34ec0e7b66a7af88773f858/resources/functionMap.php#L9610C78-L9610C78](url)

isfedorov:
@ondrejmirtes PHPStan highlights parameter `true` in call `(new Redis())->setBit('a', 1, true);` with error `Parameter #3 $value of method Redis::setBit() expects int, true given` but it shouldn't highlight it with such error since type of parameter according to [redist PHP extension](https://github.com/phpredis/phpredis/blob/develop/redis.c#L955) is `bool|int` and the type in stub is declared with typehint `bool|int` so looks like a problem somewhere in PhpStan

PR #1586, opened 2023-11-01
Made `ReflectionProperty::setAccessible()` non-`Pure`

7-zete-7:
The `ReflectionProperty::setAccessible()` method is no value returns. This method cannot be `Pure`.

### DIFF
--- a/Reflection/ReflectionProperty.php
+++ b/Reflection/ReflectionProperty.php
@@ -239,7 +239,6 @@ class ReflectionProperty implements Reflector
      * @param bool $accessible A boolean {@see true} to allow accessibility, or {@see false}
      * @return void No value is returned.
      */
-    #[Pure]
     #[PhpStormStubsElementAvailable(from: "8.1")]
     #[TentativeType]
     public function setAccessible(bool $accessible): void {}


isfedorov:
@7-zete-7 in terms of stubs `#[Pure]` means that a method doesn't have side-effects. (it doesn't match traditional determination of `Pure` but until https://youtrack.jetbrains.com/issue/WI-56211/Rename-Pure-to-something-else-like-NoSideEffect  is fixed it is as it is. Taking this into account do you still believe that the method `ReflectionProperty::setAccessible()` has side effects? If so, could you please provide some links where this is said?

7-zete-7:
@isfedorov, a few lines above there is another definition of the same method. In the definition above, the method is non-`Pure`. It is noted that starting from version 8.1, calling this method does not make sense (all properties are accessible by default) ([source](https://php.net/reflectionproperty.setaccessible)). It's not clear how the pointlessness of calling this method makes it Pure.
```php
    #[PhpStormStubsElementAvailable(to: "8.0")]
    #[TentativeType]
    public function setAccessible(#[LanguageLevelTypeAware(['8.0' => 'bool'], default: '')] $accessible): void {}

    #[Pure]
    #[PhpStormStubsElementAvailable(from: "8.1")]
    #[TentativeType]
    public function setAccessible(bool $accessible): void {}
```

Due to the presence of the `Pure` attribute, PhpStorm reports a _"Expression result is not used anywhere "_ error, but this is not possible - this method does not return anything.

Before version 8.1, this method is configuration. In case of supporting multiple PHP versions I have to use this method. I would like to avoid receiving false error messages when doing this.

If the `NoSideEffect` attribute solves the problem of this PhpStorm message, that would be wonderful. It's very confusing now.

isfedorov:
@7-zete-7 I see. The point here is that the signature marked with `Pure` was marked this way to show, that starting from PHP 8.1, as you properly noted, this method is no-op and its call doesn't make any sense (please see [WI-62098](https://youtrack.jetbrains.com/issue/WI-62098))
> Due to the presence of the Pure attribute, PhpStorm reports a "Expression result is not used anywhere " error, but this is not possible - this method does not return anything.

Yes, this is a usability problem indeed. I've filled [WI-74934](https://youtrack.jetbrains.com/issue/WI-74934) to fix the message in this case. Thank you for bringing up the attention!

PR #1588, opened 2023-11-19
There is no such thing as ReflectionEnumPureCase

ondrejmirtes:


isfedorov:
@ondrejmirtes Totally agree! Thank you!

PR #1589, opened 2023-11-23
Add Pure attribute to Closure and WeakReference/WeakMap classes

mvorisek:


mvorisek:
Is `master` branch the right one to target?

isfedorov:
@mvorisek the master branch is the correct one. Although it will be targeted for the PhpStorm 2024.1 release only. But we will try to back port the change to 2023.3 so it would be available in minor update 2023.3.1

PR #1590, opened 2023-11-23
Update MongoDB stubs for 1.17

alcaeus:
Updates stubs for the latest minor version of ext-mongodb.

Note that I wasn't able to run any checks locally:
```
docker-compose -f docker-compose.yml run -e PHP_VERSION=8.0 test_runner composer cs
WARN[0000] The "PHP_VERSION" variable is not set. Defaulting to a blank string.
[...]
PHP needs to be a minimum version of PHP 7.4.0 and maximum version of PHP 8.1.*.
Current PHP version: 8.3.0RC6.
To ignore this requirement please set `PHP_CS_FIXER_IGNORE_ENV`.
If you use PHP version higher than supported, you may experience code modified in a wrong way.
Please report such cases at https://github.com/PHP-CS-Fixer/PHP-CS-Fixer .
Script php-cs-fixer fix -v --diff --dry-run handling the cs event returned with error code 1
```

Notice how despite me passing `PHP_VERSION` as is done in CI and other places, the variable was still ignored. Is that a known issue?

### DIFF
--- a/PhpStormStubsMap.php
+++ b/PhpStormStubsMap.php
@@ -693,6 +693,7 @@ const CLASSES = array (
   'MongoDB\\Driver\\Monitoring\\CommandStartedEvent' => 'mongodb/Monitoring/CommandStartedEvent.php',
   'MongoDB\\Driver\\Monitoring\\CommandSubscriber' => 'mongodb/Monitoring/CommandSubscriber.php',
   'MongoDB\\Driver\\Monitoring\\CommandSucceededEvent' => 'mongodb/Monitoring/CommandSucceededEvent.php',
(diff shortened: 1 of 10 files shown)


isfedorov:
@alcaeus yes, this is a known issue. We have to run tests using php 8.3 but php-cs-fixer doesn't support it yet. please try to run cs using command `docker-compose -f docker-compose.yml run -e PHP_CS_FIXER_IGNORE_ENV=true test_runner composer cs` for now

alcaeus:
That worked. I confirmed that everything works and also updated the stub map which I previously forgot.

PR #1594, opened 2023-12-13
Add type declaration for argument 3 in `php_user_filter::filter()`

phansys:
BTW, I don't know if this change should be provided through `LanguageLevelTypeAware`, as I am not aware when this type was defined.

See: https://github.com/php/php-src/pull/12924#pullrequestreview-1777059768.


LolGleb:
The type should be specified in the signature only if it is the same in the php-src. Unfortunately, I can't accept these changes.

PR #1608, opened 2024-01-16
Updated sleep function documentation

jeijei4:
Updated documentation for the sleep function based on current php information ( https://www.php.net/manual/en/function.sleep.php )

### DIFF
--- a/standard/standard_0.php
+++ b/standard/standard_0.php
@@ -195,7 +195,7 @@ function bin2hex(string $string): string {}
  */
 #[Pure(true)]
 #[LanguageLevelTypeAware(["8.0" => "int"], default: "int|false")]
-function sleep(int $seconds): int {}
+function sleep(int $seconds) {}
 
 /**
  * Delay execution in microseconds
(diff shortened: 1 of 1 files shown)


standard/standard_0.php
```diff
@@ -174,17 +174,28 @@ function constant(string $name): mixed {}
...
+ * Before PHP 8.0, an E_WARNING was raised instead, and the function returned false.
  * </p>
- * @return int|false zero on success, or false on errors. If the call was interrupted
- * by a signal, sleep returns the number of seconds left
- * to sleep.
  */
+#[Pure(true)]
 #[LanguageLevelTypeAware(["8.0" => "int"], default: "int|false")]
-function sleep(int $seconds) {}
+function sleep(int $seconds): int {}
```
isfedorov, line 198: Please remove return typehint since return type is declared above by `LanguageLevelTypeAware` attribute (since 8.0 function returns `int` and prior it `int|false`

isfedorov (changes requested):

PR #1609, opened 2024-01-17
Fix xTrim threshold parameter

alexander-schranz:
threshold accept both int and string.

I'M not sure if a whole update from https://github.com/phpredis/phpredis/blob/5d293245cdd55cc16b8735a788f06d4cb98bf65c/redis.stub.php#L3964 make sense. But it look like there is a big diff between them.

### DIFF
--- a/redis/Redis.php
+++ b/redis/Redis.php
@@ -5527,7 +5527,7 @@ class Redis
      * (this is more efficient)
      *
      * @param string $stream    The STREAM key to trim.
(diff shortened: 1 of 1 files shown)


isfedorov:
@alexander-schranz generally if https://github.com/phpredis/phpredis/blob/release/6.0.2/redis.stub.php is completely valid according to 6.0.2 version of Redis extension (I'm not sure if those stubs are properly updated or not) it makes sense to update all signatures.

PR #1613, opened 2024-02-06
WI-75686 Add Pest stub

smirok:
I'm planning to move Pest stub for PhpStorm from [Pest plugin](https://github.com/JetBrains/pest-intellij/blob/97b03eca7cf313f26d4d5628fb7c10b624aa18cf/src/main/resources/library/Pest/.phpstorm.meta.php) to phpstorm-stubs


isfedorov (changes requested):
@smirok Please move meta file to directory `meta`. Current implementation will lead to adding `Pest` extension in `Settings | PHP | PHP Runtime` which is wrong. `Pest` from `PhpStormStubsProvider` also should be removed since `Pest` isn't extension

PR #1616, opened 2024-02-13
Update return descriptions of array diff functions

cookieguru:
Several array diff functions had their params changed to use `array` instead of `array1`.  This commit updates the return value _descriptions_ to match both the updated variable names and the manual's textual descriptions


isfedorov:
@cookieguru Thank you for the PR! Please fix cs-fixer errors before merge (you can run cs-fixer with `docker-compose -f docker-compose.yml run test_runner composer cs-fix` )

cookieguru:
Fixed

PR #1618, opened 2024-02-14
Tighter return types for Redis::(keys|hKeys|hVals|hGetAll)()

BenMorel:


LolGleb:
I checked Redis sources here https://github.com/phpredis/phpredis/blob/develop/redis.stub.php and I wasn't able to find any confirmation that this change is necessary. Could you please provide any links to the Redis sources where you could see `list<string>` usage instead of `array`?

BenMorel:
Actually you're right, I didn't know about `Redis::OPT_SERIALIZER`, so actually values may be `mixed` and not just `string`. I adjusted this and opened a PR on phpredis, let's see what they think: https://github.com/phpredis/phpredis/pull/2445

LolGleb:
I see that [the PR](https://github.com/phpredis/phpredis/pull/2445) has been approved but the one thing that still concerns me is that the types still don't match. For example, I see `array<string, mixed>` in phpredis and I see `array<string, string>` in PhpStorm stubs. Is that how it's meant to be?

PR #1620, opened 2024-02-17
Let EvLoop::defaultLoop() return an EvLoop instance

sandacode:
Currently in PHPStorm: 'void' method 'defaultLoop' result used.

According to the documentation it either returns the default loop, or creates + returns it.

### DIFF
--- a/Ev/Ev.php
+++ b/Ev/Ev.php
@@ -1576,11 +1576,12 @@ final class EvLoop
      * @param mixed $data
      * @param float $io_interval
      * @param float $timeout_interval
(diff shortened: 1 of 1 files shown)


LolGleb:
Ref: https://www.php.net/manual/en/evloop.defaultloop.php

LolGleb:
Thank you for the PR! The cs-fixer tests noticed some problem with the code style. Please fix it before merge (you may run cs-fixer with `docker-compose -f docker-compose.yml run test_runner composer cs-fix` )

PR #1624, opened 2024-02-27
Sync headers_sent

xepozz:
https://www.php.net/manual/en/function.headers-sent.php

### DIFF
--- a/standard/standard_4.php
+++ b/standard/standard_4.php
@@ -728,7 +728,7 @@ function header_remove(?string $name): void {}
  * @return bool headers_sent will return false if no HTTP headers
  * have already been sent or true otherwise.
  */
-function headers_sent(&$filename, &$line): bool {}
+function headers_sent(string &$filename = null, int &$line = null): bool {}
 
 /**
  * Returns a list of response headers sent (or ready to send)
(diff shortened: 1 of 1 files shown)


LolGleb:
The tests failed because the `headers_sent` function in the php source code lacks a parameter type declaration  - https://github.com/php/php-src/blob/master/ext/standard/basic_functions.stub.php#L2264
It should be done the same for stubs. Please remove the unnecessary type declarations.

standard/standard_4.php
```diff
@@ -728,7 +728,7 @@ function header_remove(?string $name): void {}
  * @return bool headers_sent will return false if no HTTP headers
  * have already been sent or true otherwise.
  */
-function headers_sent(&$filename, &$line): bool {}
+function headers_sent(string &$filename = null, int &$line = null): bool {}
```
xepozz, line 731:
```suggestion
function headers_sent(&$filename = null, &$line = null): bool {}
```

xepozz: So can we add only the default values then?

LolGleb: That is correct! Tests are green now

PR #1632, opened 2024-04-10
Brotli extension support

Simbiat:
Stub for Brotli extension https://pecl.php.net/package/brotli, https://github.com/kjdev/php-ext-brotli
First time doing this, so hope everything is correct. Extension also has steam prefix `compress.brotli` and OB handler `ob_brotli_handler`, but I am unsure how to add them, if they can/should be added even. I am also not entirely sure about values for constants besides compression, since I do not see them in the extension's code itself, so I took them from https://brotli.org/encode.html and https://rcos.io/static/internal_docs/brotli_sys/index.html under assumption, that they would be the same.


isfedorov:
@Simbiat Could you please update stubs map https://github.com/JetBrains/phpstorm-stubs?tab=readme-ov-file#how-to-update-stub-map and add appropriate folder name to `\StubTests\TestData\Providers\Stubs\PhpCoreStubsProvider::$StubDirectoryMap` under 'PECL' array? thank you!

Simbiat:
What if I do not have docker? Or PHPUnit, for that matter, which is required for running tests. For the map, can I just run `tests/Tools/generate-stub-map` script using my local build of PHP?

isfedorov:
> For the map, can I just run tests/Tools/generate-stub-map script using my local build of PHP?

Yes, generally you can run scripts using local interpreter as well

Simbiat:
Added updated map along with minor fixes for types

PR #1640, opened 2024-05-13
fix(amqp): many parameters of the name and some types, as well as the…

TimurTurdyev:
… return of the value have changed

A simple example:
in v1.11.0, the bool values were increased -
https://github.com/php-amqp/php-amqp/blob/v1.11.0/stubs/AMQPExchange.php#L173 https://github.com/php-amqp/php-amqp/blob/v1.11.0/amqp_exchange.c#L636 since version v2 it has become void - https://github.com/php-amqp/php-amqp/blob/v2.0.0/stubs/AMQPExchange.php#L195 https://github.com/php-amqp/php-amqp/blob/v2.0.0/amqp_exchange.c#L672

Link to stubs:
https://github.com/php-amqp/php-amqp/tree/latest/stubs


amqp/amqp.php
```diff
@@ -42,153 +42,178 @@
...
-     * @param string $expiration
-     * @param string $message_id
-     * @param int    $timestamp
-     * @param string $type
-     * @param string $user_id
-     * @param string $app_id
-     * @param string $cluster_id
+     * @param ?string $contentType
+     * @param ?string $contentEncoding
+     * @param ?string $correlationId
```
isfedorov, line 179: PhpDoc misses info for $headers, $deliveryMode and $priority. But actually this PhpDoc doesn't bring any useful information since all types declaration and names of parameters are already available in signature so probably it makes sense to remove it completely. Could you please either update PHPDoc or remove it completely?

amqp/amqp.php
```diff
@@ -952,17 +1018,32 @@ public function setSaslMethod($method) {}
...
  */
 class AMQPConnectionException extends AMQPException {}
 
+/**
+ * Interface representing AMQP values
+ */
+interface AMQPValue
+{
+    /**
+     * @return bool|int|float|string|null|array|AMQPValue|AMQPDecimal|AMQPTimestamp
```
isfedorov, line 1038: This PHPDoc can be safely removed since return type is declared in signature.

isfedorov (changes requested):

amqp/amqp.php
TimurTurdyev: Yes, you're right.

PR #1648, opened 2024-06-19
Update Redis stub from an very outdated version to the latest stable version (6.0.2)

karstennilsen:
The redis stub contains:
```php
@link https://github.com/phpredis/phpredis/blob/develop/redis.stub.php
```

But the stub itself was not even close to similar / up to date with the linked version.
This is causing unexpected situations where e.g. the `ttl()` method in reality returns `-2` when a key is not found but the old stub from this project specified `false`.


karstennilsen:
FYI: https://redis.io/docs/latest/commands/ttl/ states `In Redis 2.6 or older the command returns -1 if the key does not exist or if the key exist but has no associated expire.`

So the current stub is referencing an implementation of Redis 2.6 or older. Redis 2.6 last release was in 2013. You can't even find it anymore in the Redis.com download archive.

isfedorov:
Manually pushed at 1ff620818192f73554b913e4751041875dbc2a3d

PR #1651, opened 2024-07-05
SplFile{Info,Object} stub fixes

stephenreay:
Closes https://youtrack.jetbrains.com/issue/WI-78159/SplFileInfoObject-stubs-are-missing-Exception-and-class-string-details

### DIFF
--- a/SPL/SPL_c1.php
+++ b/SPL/SPL_c1.php
@@ -13,308 +13,327 @@ use JetBrains\PhpStorm\Internal\TentativeType;
  */
 class SplFileInfo implements Stringable
 {
-    /**
-     * Construct a new SplFileInfo object
-     * @link https://php.net/manual/en/splfileinfo.construct.php
-     * @param string $filename
-     * @since 5.1.2
-     */
-    public function __construct(#[LanguageLevelTypeAware(['8.0' => 'string'], default: '')] $filename) {}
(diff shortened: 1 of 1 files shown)


SPL/SPL_c1.php
```diff
@@ -227,23 +235,29 @@ public function getRealPath(): string|false {}
 
     /**
      * Gets an SplFileInfo object for the file
+     *
+     * @template T of SplFileInfo
```
VincentLanglet, line 239: Most of template in the stub have names, like 
- TValue
- TDefault
- TKey
- TPriority
- ...

Maybe there is something better to do here than `T`.

SPL/SPL_c1.php
```diff
@@ -285,10 +302,12 @@ public function setFileClass(#[LanguageLevelTypeAware(['8.0' => 'string'], defau
 
     /**
      * Sets the class used with getFileInfo and getPathInfo
+     * @template T of SplFileInfo*
```
VincentLanglet, line 305:
```suggestion
     * @template T of SplFileInfo
```
no ?

I never saw a syntax with `*`

isfedorov:
Pushed manually in 640e800cc70e84fc39d5a3e552f0cf9998b17a2d

PR #1655, opened 2024-08-05
Mark `Imagick::destroy` as deprecated

Eelinki:
According to https://www.php.net/manual/en/imagick.destroy.php


Eelinki:
YouTrack issue: https://youtrack.jetbrains.com/issue/WI-78483/Imagickdestroy-not-marked-as-deprecated

isfedorov:
Pushed manually in dedd7ae48f921eaa800d2589be4b34b41dfecabb

PR #1657, opened 2024-08-10
Fix error default value for pam functions

VincentLanglet:
Related to https://github.com/phpstan/phpstan/issues/10499

According to
https://github.com/amishmm/php-pam/blob/master/README
https://github.com/amishmm/php-pam/blob/master/pam.stub.php

The error param is optional.

### DIFF
--- a/pam/pam.php
+++ b/pam/pam.php
@@ -11,7 +11,7 @@ use JetBrains\PhpStorm\Pure;
  * @param string $password <p>
  * The user-supplied password to check.
  * </p>
- * @param string $error <p>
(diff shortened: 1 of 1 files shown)


VincentLanglet:
Hi @isfedorov, all the tests are failing but I don't think it's related to my changes.

isfedorov:
> Hi @isfedorov, all the tests are failing but I don't think it's related to my changes.

Definitely! I'm in progress of adapting stubs for PHP 8.4 and will fix this issue along with other tasks and will merge PR after that

PR #1658, opened 2024-08-12
Replace @template-implements with @template-extends in interfaces' phpDocs

vudaltsov:


isfedorov:
Pushed manually at 70e50338f1015b8e9d8c9bfcd1ca163542a2bdf9

PR #1659, opened 2024-08-12
Fix ReflectionProperty::IS_READONLY value

vudaltsov:
https://3v4l.org/1BdOW


vudaltsov:
Any problems with merging this? Bit mask by no chance can equal `5`. The actual value is `2^7 = 128`, see the attached 3v4l link.

isfedorov:
Pushed manually in da9c264637b037303f57d858d6b34753935dccc0

PR #1664, opened 2024-08-16
Fixes types on $unique on Gearman task methods

donatj:
Fixes the type for `$unique` on a handful of task methods.

This corrects them, as well as an incorrect link in the docs.

---

As you can see below, a small handful already had $unique as `string|null` but the majority did not.

https://github.com/JetBrains/phpstorm-stubs/blob/5686f9ceebde3d9338bea53b78d70ebde5fb5710/gearman/gearman.php#L1510-L1510

https://github.com/JetBrains/phpstorm-stubs/blob/5686f9ceebde3d9338bea53b78d70ebde5fb5710/gearman/gearman.php#L1542-L1542

https://github.com/JetBrains/phpstorm-stubs/blob/5686f9ceebde3d9338bea53b78d70ebde5fb5710/gearman/gearman.php#L1555-L1555

https://github.com/donatj/phpstorm-stubs/blob/5686f9ceebde3d9338bea53b78d70ebde5fb5710/gearman/gearman.php#L1568-L1568

### DIFF
--- a/gearman/gearman.php
+++ b/gearman/gearman.php
@@ -1479,7 +1479,7 @@ class GearmanClient
      * @link https://php.net/manual/en/gearmanclient.dohigh.php
      * @param string $function_name
      * @param string $workload
-     * @param string $unique
+     * @param string|null $unique
      * @return string A string representing the results of running a task
      */
     public function doHigh($function_name, $workload, $unique = null) {}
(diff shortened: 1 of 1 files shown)


donatj:
I don't think the failing build process here is anything to do with my changes…

PR #1665, opened 2024-08-25
Add stubs for FrankenPHP

dunglas:
Documents PHP functions provided by [FrankenPHP](https://frankenphp.dev).


isfedorov:
@dunglas Could you please sync your branch with upstream/master and force push your branch again?

dunglas:
@isfedorov done!

frankenphp/frankenphp.php
```diff
@@ -0,0 +1,58 @@
...
+/**
+ * Fetches all HTTP response headers.
+ *
+ * Alias of <b>apache_response_headers</b>.
+ *
+ * @link https://php.net/manual/en/function.apache-response-headers.php
+ *
+ * @return array|false An array of all FrankenPHP response headers on success or <b>FALSE</b> on failure.
+ */
+function frankenphp_response_headers(): array|bool {}
```
isfedorov, line 58: Maybe change `bool` in signature to `false` as well? Despite `false` as a standalone typehint appeared only in 8.2 from PhpStorm (or other static analyses tools) point of view it can be used here since function can't be overridden (and thus such typehint will not cause inheritance issues) and type inference will be more strict in this case and inspections will be more correct.

isfedorov (changes requested):

frankenphp/frankenphp.php
dunglas: done!

PR #1667, opened 2024-08-25
massive update stubs for PHP 8.4 and fix for old versions

isfedorov:


isfedorov:
@VincentLanglet releases of stubs are aligned with releases of PhpStorm, so release of stubs v2024.3 will be after release of PhpStorm 2024.3 in november

PR #1685, opened 2024-10-20
DomNode unimplemented methods have been removed in PHP 8.0

Bellangelo:
Reference: https://www.php.net/manual/en/class.domnode.php

> 8.0.0 | The unimplemented methods DOMNode::compareDocumentPosition(), DOMNode::isEqualNode(), DOMNode::getFeature(), DOMNode::setUserData() and DOMNode::getUserData() have been removed.

### DIFF
--- a/dom/dom_c.php
+++ b/dom/dom_c.php
@@ -270,6 +270,7 @@ class DOMNode
 
     /**
      * @return int
+     * @removed 8.0
      */
     #[LanguageLevelTypeAware(['8.4' => 'int'], default: '')]
     public function compareDocumentPosition(DOMNode $other) {}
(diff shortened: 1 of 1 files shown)


LolGleb:
Thank you for your PR! I have checked all the mentioned methods and it seems that `DOMNode::isEqualNode()` was reintroduced in php 8.3 https://github.com/php/php-src/pull/11690
I believe we shouldn't mark it as removed in this case.

Bellangelo:
> Thank you for your PR! I have checked all the mentioned methods and it seems that `DOMNode::isEqualNode()` was reintroduced in php 8.3 [php/php-src#11690](https://github.com/php/php-src/pull/11690)
> I believe we shouldn't mark it as removed in this case.

Hmm.. interesting. I guess I will have to open a PR in the docs as well :D.

Nevertheless, will something like this work? How do handle such cases?
```php
/**
 * @since 5.6
 * @removed 8.0
 * @since 8.3
*/
public function method(): void;
```

LolGleb:
I wasn't able to find similar cases in the project... I don't think that it's allowed to have more than one `@since` mark in a PHPDoc. I guess the best thing we can do is simply leave this method without `@removed` or `@since` marks.

nielsdos:
compareDocumentPosition is also reintroduced with an actual implementation in PHP 8.4.

LolGleb:
You are right. That means it should be also left without `@removed` mark.

Bellangelo:
@LolGleb @nielsdos I have removed the `@removed` from the `compareDocumentPosition` and `isEqualNode`.

PR #1700, opened 2024-12-04
Adding \Swoole\Coroutine\{run, go} functions

CViniciusSDias:
The functions `go` and `run` are not available from the `\Swoole\Coroutine\` namespace in the stubs. This PR adds them.

### DIFF
--- a/swoole/functions.php
+++ b/swoole/functions.php
@@ -2,6 +2,11 @@
 
 declare(strict_types=1);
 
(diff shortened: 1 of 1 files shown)


isfedorov:
@CViniciusSDias Could you please update stubs map with script `tests/Tools/generate-stub-map` as commit result file as well?

isfedorov:
Updated in main branch

PR #1709, opened 2025-01-06
adds stubs for Pdo\Pgsql

aldilop:
these are taken from the documentation in https://www.php.net/manual/en/class.pdo-pgsql.php, which can be sourced back to https://github.com/php/php-src/blob/master/ext/pdo_pgsql/pgsql_driver.stub.php

### DIFF
--- a/PDO/PDO.php
+++ b/PDO/PDO.php
@@ -2058,4 +2058,57 @@ namespace Pdo {
 
         public function getWarningCount(): int {}
     }
+
+    /**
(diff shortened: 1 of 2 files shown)


isfedorov:
@aldilop Thank you for the PR! Could you please update StubsMap with command `docker compose -f docker-compose.yml run test_runner /usr/local/bin/php tests/Tools/generate-stub-map` (or use your local interpreter) and commit result map file along with other changers?

PR #1712, opened 2025-01-25
Update ob_start signature

VincentLanglet:
cf https://www.php.net/manual/en/function.ob-start.php

ob_start is 
```
ob_start(?callable $callback = null, int $chunk_size = 0, int $flags = PHP_OUTPUT_HANDLER_STDFLAGS): bool
```

### DIFF
--- a/standard/standard_8.php
+++ b/standard/standard_8.php
@@ -143,7 +143,7 @@ function metaphone(string $string, int $max_phonemes = 0) {}
 /**
  * Turn on output buffering
  * @link https://php.net/manual/en/function.ob-start.php
- * @param callable|null $callback [optional] <p>
+ * @param callable|null $callback <p>
  * An optional output_callback function may be
  * specified. This function takes a string as a parameter and should
  * return a string. The function will be called when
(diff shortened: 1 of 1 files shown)


VincentLanglet:
Should I do something for the failing check @isfedorov ? I'm not sure to understand the error.

standard/standard_8.php
```diff
@@ -143,7 +143,7 @@ function metaphone(string $string, int $max_phonemes = 0) {}
 /**
  * Turn on output buffering
  * @link https://php.net/manual/en/function.ob-start.php
- * @param callable $callback [optional] <p>
+ * @param callable|null $callback [optional] <p>
```
isfedorov, line 146: @VincentLanglet since you have added `null` as a default value in signature, `[optional]` can be removed here. It's clear that parameter is optional based on signature

isfedorov (changes requested):

isfedorov:
@VincentLanglet the failure is related to not supported yet php-cs-fixer + PHP 8.4. You may ignore it.

PR #1719, opened 2025-02-22
feat: add .gitattributes for export-ignore

savinmikhail:
Add .gitattributes for export-ignore directive for 100% dev-only files


savinmikhail:
I created issue

https://youtrack.jetbrains.com/issue/WI-80628/Add-export-ignore-directives-for-dev-only-files

PR #1726, opened 2025-03-16
Fix function purity

zonuexe:


date/date.php
```diff
@@ -528,7 +528,7 @@ function gmmktime(
  * </p>
  * @return bool true if the date given is valid; otherwise returns false.
  */
-#[Pure(true)]
+#[Pure]
```
zonuexe, line 531: The [`checkdate()`](https://www.php.net/checkdate) function is independent of the current time and locale.

PR #1728, opened 2025-03-26
Remove second parameter incorrect nullable

AntoineDly:
Linked to this issue : [WI-80919](https://youtrack.jetbrains.com/issue/WI-80919)

As we can see in this pull request, PHP doesn't tolerate null as an array in implode/join functions : https://github.com/php/php-src/pull/12683


LolGleb:
Sorry, I made a mistake when I merged this PR. I checked what reflection returns and what is in the [PHP source code](https://github.com/search?q=repo%3Aphp%2Fphp-src%20implode&type=code). It turns out that $array is indeed an optional parameter and its default value is null.

Reverted it here #1736

PR #1729, opened 2025-04-10
Update stubs for ext-mongodb 2.0

alcaeus:
I opted for making the 2.0 version the only version we support. I'm not sure if we could use conditional class definitions to cover both 1.x and 2.x, I'd rather have people use this as motivation to move away from 1.x


alcaeus:
@isfedorov CI seems to be having a bad day - it fails before even finishing setting up docker containers.

isfedorov:
@alcaeus Please try to rebase. I've removed problematic dependency

alcaeus:
@isfedorov thank you, that looks "better". Now it's my fault for the red CI, will check that out and fix it in the coming days.

alcaeus:
I'm not completely familiar with the test setup, but I can only assume that the failures are because we removed some constants in the 2.0 version of the extension, while reflection output still sees them there. This would indicate that the 1.x extension is somehow installed, but I can't find where, the only place that explicitly mentions the MongoDB extension is the tests/DockerImages/peclExtensions image, but installation is disabled there and can't be enabled as the image uses PHP 8.0 which is no longer supported by us.

@isfedorov can you point me in the right direction for fixing tests? Is there a way to mark constants, methods, or classes as removed in an extension release instead of completely removing them from the stubs?

isfedorov:
@alcaeus as far as I can see the issue is related to the fact that removed constant is used in .meta file `meta/.phpstorm.meta.php` . It uses this constant as an expected argument's value `expectedArguments(\MongoDB\Driver\ReadPreference::__construct(), 0, \MongoDB\Driver\ReadPreference::RP_PRIMARY,\MongoDB\Driver\ReadPreference::RP_PRIMARY_PREFERRED,\MongoDB\Driver\ReadPreference::RP_SECONDARY,\MongoDB\Driver\ReadPreference::RP_SECONDARY_PREFERRED,\MongoDB\Driver\ReadPreference::RP_NEAREST);` so you need to update this usage

alcaeus:
Thanks @isfedorov, I wasn't aware that file was used for stubs as well. I changed to the correct constants there and it fixed the tests. The cs-fixer task is failing due to errors in unrelated files ([see logs](https://github.com/JetBrains/phpstorm-stubs/actions/runs/14856023557/job/41709492312?pr=1729#step:6:92)), so I think this PR is ready to go.

isfedorov:
@alcaeus Yes, please ignore cs-fixer failure. PR is good to go. But I suppose this version of stubs will be bundled only into PhpStorm 2025.2, (in July) unfortunately.

alcaeus:
> But I suppose this version of stubs will be bundled only into PhpStorm 2025.2, (in July) unfortunately.

No worries. Out of curiosity, I was under the impression that the stubs are updated automatically from the repo - is that not the case?

isfedorov:
They are updated automatically from the repo but are being merged only to main branch. And due to internal policies and workflow with main/stable/release branches they will not be backported to the 2025.1 minor updates.

PR #1730, opened 2025-04-10
Fix bad `Pure` attributes

cs278:
This PR is based on *my* interpretation of the documentation on the attribute that `Pure(true)` means a function:

* Does **not** have any side effects on external state or the supplied parameters
* **and** the result depends on some external state outside of the supplied parameters.

This contradicts with the interpretation in #1724 by @zonuexe which was accepted:

> Added the missing `#[Pure(true)]` attribute to some functions **that have side effects** [emphasis mine] but whose return value is important.

The reasoning for each change is documented in the commit messages.

As such it reverts some of the changes made in that PR. If my interpretation is wrong then reject this, but please improve the documentation.

### DIFF
--- a/Core/Core.php
+++ b/Core/Core.php
@@ -605,7 +605,7 @@ function get_class_vars(string $class): array {}
  * for the specified <i>object</i> in scope. If a property have
  * not been assigned a value, it will be returned with a null value.
  */
-#[Pure(true)]
+#[Pure]
 function get_object_vars(object $object): array {}
 
 /**
@@ -813,7 +813,7 @@ function create_function(string $args, string $code): false|string {}
  * by this function, the return value will be the string
  * Unknown.
  */
-#[Pure(true)]
+#[Pure]
 function get_resource_type($resource): string {}
 
 /**
--- a/Core/Core.php
+++ b/Core/Core.php
@@ -445,7 +445,6 @@ function property_exists($object_or_class, string $property): bool {}
  * @link https://secure.php.net/manual/en/function.trait-exists.php
  * @since 5.4
  */
-#[Pure(true)]
 function trait_exists(string $trait, bool $autoload = true): bool {}
 
 /**
@@ -460,7 +459,6 @@ function trait_exists(string $trait, bool $autoload = true): bool {}
  * @return bool true if <i>class_name</i> is a defined class,
  * false otherwise.
  */
-#[Pure(true)]
 function class_exists(string $class, bool $autoload = true): bool {}
 
 /**
@@ -476,7 +474,6 @@ function class_exists(string $class, bool $autoload = true): bool {}
  * <i>interface_name</i> has been defined, false otherwise.
  * @since 5.0.2
  */
-#[Pure(true)]
 function interface_exists(string $interface, bool $autoload = true): bool {}
 
 /**
@@ -508,7 +505,6 @@ function function_exists(string $function): bool {}
  * false otherwise.
  * @since 8.1
  */
-#[Pure(true)]
 function enum_exists(string $enum, bool $autoload = true): bool {}
 
 /**
--- a/standard/standard_8.php
+++ b/standard/standard_8.php
@@ -244,7 +244,6 @@ function ob_end_clean(): bool {}
  * @link https://php.net/manual/en/function.ob-get-flush.php
  * @return string|false the output buffer or false if no buffering is active.
  */
-#[Pure(true)]
 function ob_get_flush(): string|false {}
 
 /**
@@ -253,7 +252,6 @@ function ob_get_flush(): string|false {}
  * @return string|false the contents of the output buffer and end output buffering.
  * If output buffering isn't active then false is returned.
  */
-#[Pure(true)]
 function ob_get_clean(): string|false {}
 
 /**
--- a/standard/standard_9.php
+++ b/standard/standard_9.php
@@ -445,7 +445,6 @@ function array_intersect_uassoc(
  * array1 that are present in all the arguments.
  * @meta
  */
-#[Pure]
 function array_uintersect_uassoc(
     array $array,
     #[PhpStormStubsElementAvailable(from: '5.3', to: '7.4')] array $array2,
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -50,4 +50,4 @@ jobs:
         run: docker compose -f docker-compose.yml run test_runner vendor/bin/phpunit --no-progress --testsuite Structure
 
       - name: run cs fixer
-        run:  docker run --rm -v .:/opt/project -w /opt/project php:8.3-cli sh -c 'curl -sS https://getcomposer.org/installer | php && php composer.phar run cs'
+        run:  docker run --rm -e PHP_CS_FIXER_IGNORE_ENV=1 -v .:/opt/project -w /opt/project php:8.4-cli sh -c 'curl -sS https://getcomposer.org/installer | php && php composer.phar run cs'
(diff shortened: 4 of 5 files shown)


Core/Core.php
```diff
@@ -445,7 +445,6 @@ function property_exists($object_or_class, string $property): bool {}
  * @link https://secure.php.net/manual/en/function.trait-exists.php
  * @since 5.4
  */
-#[Pure(true)]
```
isfedorov, line 448: It looks like this change is incorrect since `trait_exists` actually depends on global state. For example 
```
var_dump(trait_exists('MyTrait')); // false

trait MyTrait {}

var_dump(trait_exists('MyTrait')); // true
```
And at the same time this function doesn't produce any side effects, that's why it should be marked with `#[Pure(true)]` attribute.

Core/Core.php
```diff
@@ -460,7 +459,6 @@ function trait_exists(string $trait, bool $autoload = true): bool {}
  * @return bool true if <i>class_name</i> is a defined class,
  * false otherwise.
  */
-#[Pure(true)]
```
isfedorov, line 463: The same concern as above

Core/Core.php
```diff
@@ -476,7 +474,6 @@ function class_exists(string $class, bool $autoload = true): bool {}
  * <i>interface_name</i> has been defined, false otherwise.
  * @since 5.0.2
  */
-#[Pure(true)]
```
isfedorov, line 479: The same concern as above

Core/Core.php
```diff
@@ -508,7 +505,6 @@ function function_exists(string $function): bool {}
  * false otherwise.
  * @since 8.1
  */
-#[Pure(true)]
```
isfedorov, line 511: The same concern as above

Core/Core.php
```diff
@@ -609,7 +605,7 @@ function get_class_vars(string $class): array {}
  * for the specified <i>object</i> in scope. If a property have
  * not been assigned a value, it will be returned with a null value.
  */
-#[Pure(true)]
```
isfedorov, line 612: It looks like this change is also incorrect since function actually depends on global state. The function depends on the structure of the class from which the object was instantiated. If the class definition changes (e.g., due to autoloading), the result can change. Also if string property names or values are affected by locale settings (e.g., setlocale()), it might affect the output. Please correct me if I'm wrong

Core/Core.php
```diff
@@ -817,7 +813,7 @@ function create_function(string $args, string $code): false|string {}
  * by this function, the return value will be the string
  * Unknown.
  */
-#[Pure(true)]
```
isfedorov, line 820: I suppose this change is also incorrect. The function operates on a resource handle and some of the resources can be managed globally such as database connections or network sockets. If a global resource pool is modified the resource type can change or become invalid. Also in case of this code it will return different results
```
$handle = fopen('example.txt', 'r');
echo get_resource_type($handle); // Outputs: stream

fclose($handle);
echo get_resource_type($handle); // Outputs: Unknown
```
Since the state of resource has been changed (resource connection closed)

standard/standard_8.php
```diff
@@ -244,7 +244,6 @@ function ob_end_clean(): bool {}
  * @link https://php.net/manual/en/function.ob-get-flush.php
  * @return string|false the output buffer or false if no buffering is active.
  */
-#[Pure(true)]
```
isfedorov, line 247: I suppose this is incorrect as well. The function relies on the current output buffering stack. If no output buffer is active, ob_get_flush() will return false. The output buffering state is part of PHP's global state.. Please correct me if I'm wrong.

standard/standard_8.php
```diff
@@ -253,7 +252,6 @@ function ob_get_flush(): string|false {}
  * @return string|false the contents of the output buffer and end output buffering.
  * If output buffering isn't active then false is returned.
  */
```
isfedorov, line 254: The same concern as above

standard/standard_9.php
```diff
@@ -445,7 +445,6 @@ function array_intersect_uassoc(
  * array1 that are present in all the arguments.
  * @meta
  */
-#[Pure]
```
isfedorov, line 448: IMO this change is incorrect as well. The function doesn't depend on global state and doesn't have any side effects. it just computes the intersection of arrays with User-defined comparison for values and User-defined comparison for keys
```
function value_compare($a, $b) {
    return strcmp($a, $b);
}
function key_compare($a, $b) {
    return strcmp($a, $b);
}

$array1 = ["a" => "apple", "b" => "banana"];
$array2 = ["a" => "apple", "b" => "berry"];

$result = array_uintersect_uassoc($array1, $array2, "value_compare", "key_compare");
print_r($result); // Only "a" => "apple" matches both value and key
```
Please correct me if I'm wrong.

Core/Core.php
cs278: Per the commit message these methods all trigger the autoloader which can (and is really intended) to have side effects namely defining symbols but nothing stopping it doing other things.

standard/standard_9.php
cs278: This is not straight forward because in your example this function is pure, but if you use a comparison function that isn't pure then this function is not pure.

The convention in this package seems to be that any function/method taking a callback is not marked as pure.

Core/Core.php
isfedorov: Ah, yes. I agree here. The function can produce indirect side effects via autoloader.

standard/standard_8.php
isfedorov: I see. Thank you! I agree.

standard/standard_9.php
isfedorov: Right. The function can produce side effects via callback. Thank you!

Core/Core.php
isfedorov: @cs278 Could you please comment on this? Am I right or do I miss something?

isfedorov (changes requested):

zonuexe:
The semantics of `#[JetBrains\PhpStorm\Pure(true)]` are explained in https://github.com/JetBrains/phpstorm-attributes/blob/1.2/src/Pure.php#L17C44-L17C127

> `$mayDependOnGlobalScope` Whether the function result may be dependendent on anything except passed variables

Since this explanation is [oxymoron](https://en.wikipedia.org/wiki/Oxymoron), I guess it's possible to remove all `#[Pure(true)]`, but I don't see how that would break backwards compatibility of PhpStorm. 🤪

VincentLanglet:
> The semantics of `#[JetBrains\PhpStorm\Pure(true)]` are explained in [JetBrains/phpstorm-attributes@`1.2`/src/Pure.php#L17C44-L17C127](https://github.com/JetBrains/phpstorm-attributes/blob/1.2/src/Pure.php?rgh-link-date=2025-05-23T17%3A50%3A26Z#L17C44-L17C127)
> 
> > `$mayDependOnGlobalScope` Whether the function result may be dependendent on anything except passed variables
> 
> Since this explanation is [oxymoron](https://en.wikipedia.org/wiki/Oxymoron), I guess it's possible to remove all `#[Pure(true)]`, but I don't know how that would break backwards compatibility of PhpStorm. 🤪

I don't consider this as oxymoron.
I understand like @cs278 as Pure(true) is just meant to say "hasNoSideEffect".

Are you still interested in this PR @cs278 ? I agree with the comment https://github.com/JetBrains/phpstorm-stubs/pull/1730#discussion_r2039064660 and `#[Pure(true)]` should be kept for `get_object_vars` and `get_resource_type`.

PR #1741, opened 2025-05-14
Fix DomElement::attributes type

VincentLanglet:
Hi @isfedorov

The description say
```
A <classname>DOMNamedNodeMap</classname> containing the attributes of this node (if it is a <classname>DOMElement</classname>) or NULL otherwise.
```
and the signature should be the same than DomNode.

### DIFF
--- a/dom/dom_c.php
+++ b/dom/dom_c.php
@@ -1720,7 +1720,7 @@ class DOMElement extends DOMNode implements DOMParentNode, DOMChildNode
     public $nextSibling;
 
     /**
-     * @var DOMNamedNodeMap<DOMAttr>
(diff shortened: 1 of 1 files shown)


VincentLanglet:
Error seems unrelated

dom/dom_c.php
```diff
@@ -1720,7 +1720,7 @@ class DOMElement extends DOMNode implements DOMParentNode, DOMChildNode
     public $nextSibling;
 
     /**
-     * @var DOMNamedNodeMap<DOMAttr>
+     * @var DOMNamedNodeMap<DOMAttr>|null
```
stof, line 1723: The type also need to be updated in the `LanguageLevelTypeAware` attribute defining the native type per PHP version.

PR #1745, opened 2025-05-27
Revert #1741 - \DOMElement::$attributes is never null

Jean85:
See https://github.com/phpstan/phpstan/issues/13076

This reverts #1741
From the PHP docs: https://www.php.net/manual/en/class.domnode.php#domnode.props.attributes

> A [DOMNamedNodeMap](https://www.php.net/manual/en/class.domnamednodemap.php) containing the attributes of this node (if it is a [DOMElement](https://www.php.net/manual/en/class.domelement.php)) or [null](https://www.php.net/manual/en/reserved.constants.php#constant.null) otherwise.

So, on `\DOMElement` it's correct to override and mark it as not nullable.

### DIFF
--- a/dom/dom_c.php
+++ b/dom/dom_c.php
@@ -1720,7 +1720,7 @@ class DOMElement extends DOMNode implements DOMParentNode, DOMChildNode
     public $nextSibling;
 
     /**
-     * @var DOMNamedNodeMap<DOMAttr>|null
+     * @var DOMNamedNodeMap<DOMAttr>
      * A <classname>DOMNamedNodeMap</classname> containing the attributes of this node (if it is a <classname>DOMElement</classname>) or NULL otherwise.
      * @link https://php.net/manual/en/class.domnode.php#domnode.props.attributes
      */
--- a/dom/dom_c.php
+++ b/dom/dom_c.php
@@ -1724,7 +1724,7 @@ class DOMElement extends DOMNode implements DOMParentNode, DOMChildNode
      * A <classname>DOMNamedNodeMap</classname> containing the attributes of this node (if it is a <classname>DOMElement</classname>) or NULL otherwise.
      * @link https://php.net/manual/en/class.domnode.php#domnode.props.attributes
      */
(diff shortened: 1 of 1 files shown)


VincentLanglet:
I agree with this PR @isfedorov, sorry my mistake

VincentLanglet:
> I'm sorry but I'm not sure I've got it.
> 
> > A [DOMNamedNodeMap](https://www.php.net/manual/en/class.domnamednodemap.php) containing the attributes of this node (if it is a [DOMElement](https://www.php.net/manual/en/class.domelement.php)) or [null](https://www.php.net/manual/en/reserved.constants.php#constant.null) otherwise.
> 
> So as I understand this, this property can contain null if the object isn't a DOMElement. But even if the object is a DOMElement reflection still shows that the type of property is nullable. [3v4l.org/HPtTl](https://3v4l.org/HPtTl) Could you please clarify?

The logic behind is kinda
```
get_type($dom->attributes) = $dom instanceof DOMELement ? DOMNamedNodeMap : null;
```

Indeed seems like the reflection is telling the property might be nullable, but it's untrue for a `DOMELement`.
- The property is initialized
- The property is read only so cannot be re-set to null

see https://onlinephp.io/c/7b879

isfedorov:
@VincentLanglet I completely understand the reasoning of this change but what I don't like about this change is that we are trying to align stubs with _actual runtime value_ instead of PHP native signature. What will happen if PHP implementation changes the default assigned value of property to `null` in DomElement implementation? The signature of the property will remain the same as it is at the moment, right? PhpStorm will not give you a warning that property can be null, but since the actual value will be `null` you can easily get an NPE in runtime. But as far as I can see this issue produces a lot of noise by other linters so let's make this change. The only thing that should be fixed additionally - type in LanguageLevelTypeAware attribute.

VincentLanglet:
Oh indeed.

@Jean85 Can you change
```
#[LanguageLevelTypeAware(['8.1' => 'DOMNamedNodeMap|null'], default: '')]
```
to
```
#[LanguageLevelTypeAware(['8.1' => 'DOMNamedNodeMap'], default: '')]
```
?

PR #1749, opened 2025-05-30
DateTimeInterface::format() with Pure attribute

integer:


LolGleb:
Thank you for your PR! After reviewing it, I believe that adding the `#[Pure]` attribute to the `DateTimeInterface::format()` method is not appropriate. Here’s why:

This method depends not only on the input argument, but also on the internal state of the object.

For example, calling the method with the same $format value can yield different results depending on the instance’s internal state:
```
$date1 = new DateTime('2023-02-03');
$date2 = new DateTime('2024-04-05');
echo $date1->format('Y-m-d'); // "2023-02-03"
echo $date2->format('Y-m-d'); // "2024-04-05"
```

This violates one of the key conditions of purity: the output of the method must be consistent given the same input arguments.

If you think I am wrong, please point out where I am mistaken.

PR #1750, opened 2025-06-01
Sync RedisCluster stub with phpredis stub

RobiNN1:
Synchronized (with https://github.com/phpredis/phpredis/blob/develop/redis_cluster.stub.php) and also added missing methods. Maybe in the next PR I will add missing phpdocs, for now this should be enough for proper autocomplete.

### DIFF
--- a/redis/RedisCluster.php
+++ b/redis/RedisCluster.php
@@ -2,12 +2,12 @@
 
 /**
  * Helper autocomplete for php redis cluster extension.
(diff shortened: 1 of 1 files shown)


RobiNN1:
CS Fixer fails on unrelated files

Files that were not fixed due to errors reported during linting before fixing:
   1) /opt/project/tests/Parsers/StubParser.php
   2) /opt/project/tests/Parsers/Visitors/ASTVisitor.php
   3) /opt/project/Core/Core.php

PR #1752, opened 2025-06-12
Add generics to array_find

JurianArie:


standard/standard_10.php
```diff
@@ -1,8 +1,13 @@
...
 /**
+ * @template TKey
+ * @template TValue
+ * @param array<TKey, TValue> $array
+ * @param callable(TValue, TKey): bool
+ * @return TValue|null
  * @since 8.4
  */
-function array_find(array $array, callable $callback): mixed {}
+function array_find(array $array, callable $callback) {}
```
isfedorov, line 10: @JurianArie return type should stay here to match reflection data. And a failing test notifies about this
Screenshot 2025-06-12 at 16 48 56
Other test failures are not related to your changes so I'll fix them myself. Could you please just return back `mixed` return type?

isfedorov (changes requested):

standard/standard_10.php
JurianArie: The mixed type has been added back. That issue does indeed look related!

PR #1754, opened 2025-06-23
E_USER_ERROR is deprecated in PHP 8.4

staabm:
see https://wiki.php.net/rfc/deprecations_php_8_4#deprecate_passing_e_user_error_to_trigger_error

technically this constant is not deprecated yet, but only its use with `trigger_error`.


LolGleb:
Thank you for your PR!
From my understanding, the E_USER_ERROR constant itself is not being deprecated. Only its usage with trigger_error() is being deprecated. This distinction is important because the constant is still valid in other contexts.

In PhpStorm, this case is highlighted correctly: we show a deprecation warning only when E_USER_ERROR is passed to trigger_error().
image

PR #1760, opened 2025-07-20
Add @throws tag on createFromFormat

VincentLanglet:
According to https://www.php.net/manual/en/datetime.createfromformat.php#refsect1-datetime.createfromformat-errors
since PHP 8, `createFromFormat` can throws ValueError when the datetime contains NULL-bytes.


VincentLanglet:
Is it ok for you @LolGleb ? :)

PR #1761, opened 2025-07-20
Improve DomNode methods return type

VincentLanglet:
cc @isfedorov


dom/dom_c.php
```diff
@@ -162,14 +162,15 @@ class DOMNode
...
+     * @template TNode of DOMNode
+     * @param TNode $node <p>
      * The new node.
      * </p>
      * @param null|DOMNode $child [optional] <p>
      * The reference node. If not supplied, newnode is
      * appended to the children.
      * </p>
-     * @return DOMNode The inserted node.
+     * @return TNode The inserted node.
```
VincentLanglet, line 173: According to the phpdoc and https://www.php.net/manual/en/domnode.insertbefore.php
it returns the node inserted, which is the one coming from the param.

dom/dom_c.php
```diff
@@ -179,35 +180,38 @@ public function insertBefore(
...
      * The new node. It must be a member of the target document, i.e.
      * created by one of the DOMDocument->createXXX() methods or imported in
      * the document by .
      * </p>
-     * @param DOMNode $child <p>
+     * @param TNode $child <p>
      * The old node.
      * </p>
-     * @return DOMNode|false The old node or false if an error occur.
+     * @return TNode|false The old node or false if an error occur.
```
VincentLanglet, line 192: According to the phpdoc and https://www.php.net/manual/en/domnode.replacechild.php
the old node is returned which is the second param if I understand.

dom/dom_c.php
```diff
@@ -179,35 +180,38 @@ public function insertBefore(
...
     /**
      * Removes child from list of children
      * @link https://php.net/manual/en/domnode.removechild.php
-     * @param DOMNode $child <p>
+     * @template TNode of DOMNode
+     * @param TNode $child <p>
      * The removed child.
      * </p>
-     * @return DOMNode If the child could be removed the functions returns the old child.
+     * @return TNode|false If the child could be removed the functions returns the old child.
```
VincentLanglet, line 203: This returns the child removed (which is the one in param)
AND `false` was missing

https://www.php.net/manual/en/domnode.removechild.php

dom/dom_c.php
```diff
@@ -179,35 +180,38 @@ public function insertBefore(
...
     /**
      * Adds new child at the end of the children
      * @link https://php.net/manual/en/domnode.appendchild.php
-     * @param DOMNode $node <p>
+     * @template TNode of DOMNode
+     * @param TNode $node <p>
      * The appended child.
      * </p>
-     * @return DOMNode The node added.
+     * @return TNode|false The node added.
```
VincentLanglet, line 214: This returns the node added (which is coming form the param)
AND `false` was missing

https://www.php.net/manual/en/domnode.appendchild.php

dom/dom_c.php
```diff
@@ -162,14 +162,15 @@ class DOMNode
...
+     * @template TNode of DOMNode
+     * @param TNode $node <p>
      * The new node.
      * </p>
      * @param null|DOMNode $child [optional] <p>
      * The reference node. If not supplied, newnode is
      * appended to the children.
      * </p>
-     * @return DOMNode The inserted node.
+     * @return TNode|false The inserted node.
```
VincentLanglet, line 173: According to the phpdoc and https://www.php.net/manual/en/domnode.insertbefore.php
it returns the node inserted, which is the one coming from the param.

AND `false` was missing

dom/dom_c.php
```diff
@@ -226,7 +230,7 @@ public function hasChildNodes(): bool {}
      * Indicates whether to copy all descendant nodes. This parameter is
      * defaulted to false.
      * </p>
-     * @return static The cloned node.
+     * @return static|false The cloned node.
```
VincentLanglet, line 233: cf https://www.php.net/manual/en/domnode.clonenode.php#refsect1-domnode.clonenode-description
and https://github.com/php/php-src/blob/d7f9caf37c0915d1096c398671f4c92f0a11a1d3/ext/dom/php_dom.stub.php#L427

PR #1763, opened 2025-07-22
Fix incorrect @param type for $previous in ErrorException::__construct docblock

TomC2187:
This PR updates the PHPDoc for the ErrorException::__construct method to correctly reflect the accepted type for the $previous parameter.

The actual constructor accepts ?Throwable, not just Exception. Updating the PHPDoc aligns the documentation with the method signature and resolves static analysis issues in tools like PHPStan and Psalm, which currently detect a mismatch.

This improves type accuracy and ensures better compatibility with exception chaining involving Error.


TomC2187:
Dupe

PR #1767, opened 2025-07-28
Fix createFromInterface return type

VincentLanglet:
See:
- https://github.com/php/php-src/blob/e48099a03c677bea5d0c615cefcab094f8b5253b/ext/date/php_date.stub.php#L361-L362
- https://github.com/php/php-src/blob/e48099a03c677bea5d0c615cefcab094f8b5253b/ext/date/php_date.stub.php#L564-L565

### DIFF
--- a/date/date_c.php
+++ b/date/date_c.php
@@ -471,7 +471,7 @@ class DateTimeImmutable implements DateTimeInterface
 
     /**
      * @param DateTimeInterface $object
-     * @return DateTimeImmutable
+     * @return static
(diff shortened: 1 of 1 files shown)


LolGleb:
Hi, thanks for the PR. As far as I can tell, your changes look good. We'll need some time on our end to figure out how to adapt the tests, which will take a while.

If you'd like to help with that, we'd be super grateful, but just a heads-up, framework is quite complex

PR #1768, opened 2025-07-30
Add @throws for request_parse_body()

GottemHams:
According to the [PHP docs](https://www.php.net/manual/en/function.request-parse-body.php#refsect1-function.request-parse-body-errors):

> When the request body is invalid, according to the Content-Type header, a [RequestParseBodyException](https://www.php.net/manual/en/class.requestparsebodyexception.php) is thrown.
>
> A [ValueError](https://www.php.net/manual/en/class.valueerror.php) is thrown when options contains invalid keys, or invalid values for the corresponding key.

### DIFF
--- a/standard/standard_10.php
+++ b/standard/standard_10.php
@@ -34,7 +34,6 @@ function http_clear_last_response_headers(): void {}
  * @param array|null $options
  * @return array<int, array>
  * @throws RequestParseBodyException if the request body uses an invalid/unsupported content type
- * @throws ValueError if <i>options</i> contains an invalid key or value
  */
 function request_parse_body(?array $options = null): array {}
 /**
--- a/standard/standard_10.php
+++ b/standard/standard_10.php
@@ -33,6 +33,8 @@ function http_clear_last_response_headers(): void {}
  * @since 8.4
  * @param array|null $options
  * @return array<int, array>
+ * @throws RequestParseBodyException if the request body uses an invalid/unsupported content type
+ * @throws ValueError if <i>options</i> contains an invalid key or value
  */
 function request_parse_body(?array $options = null): array {}
 /**


standard/standard_10.php
```diff
@@ -33,6 +33,8 @@ function http_clear_last_response_headers(): void {}
  * @since 8.4
  * @param array|null $options
  * @return array<int, array>
+ * @throws RequestParseBodyException if the request body uses an invalid/unsupported content type
+ * @throws ValueError if <i>options</i> contains an invalid key or value
```
VincentLanglet, line 37:
```suggestion
 * @throws RequestParseBodyException if the request body uses an invalid/unsupported content type
```
I wonder if only RequestParseBodyException should be added...

ValueError/TypeError is something which can basically be thrown by every existing php functions, but still the `@throws` tag was not added 

As an exemple `explode` might throw a `ValueError` but it's not added to the phpdoc
https://www.php.net/manual/en/function.explode.php because it rely on the fact that the function is used in a correct way.

GottemHams: I think in this specific case it might be best to keep the `@throws ValueError` though. It can also be thrown if you call `request_parse_body()` from CLI for example, which may not be immediately obvious for everyone. Or that it really only supports 2 content types (at least for now), so you would pretty much always have to catch the `ValueError`. Otherwise you'd have to keep track of the types that PHP supports natively:

```php
if($content_type === 'application/x-www-form-urlencoded' || $content_type === 'multipart/form-data') {
  [$data, $files] = request_parse_body();
}
else {
  // Do something, e.g. respond with 400 Bad Request
}
```

Versus:

```php
try {
  [$data, $files] = request_parse_body();
}
catch(ValueError) {
  // Do something, e.g. respond with 400 Bad Request
}
```

Things like `explode()` are a lot more "difficult" to break, since it only throws for an empty separator string and you can already use a string like an array anyway. :>

PR #1777, opened 2025-08-08
Improve solr signatures

VincentLanglet:
Mainly improving `optional` param signature by adding a default value.
cf https://www.php.net/manual/en/class.solrdismaxquery.php for instance

And a missing `@return void`

### DIFF
--- a/solr/Documents/SolrDocument.php
+++ b/solr/Documents/SolrDocument.php
@@ -178,8 +178,9 @@ final class SolrDocument implements ArrayAccess, Iterator, Serializable
     /**
      * (PECL solr &gt;= 0.9.2)<br/>
      * Returns a SolrInputDocument equivalent of the object
+     *
      * @link https://php.net/manual/en/solrdocument.getinputdocument.php
-     * @return SolrInputDocument|null <p>
+     * @return SolrInputDocument <p>
      * Returns a SolrInputDocument on success and <b>NULL</b> on failure.
      * </p>
      */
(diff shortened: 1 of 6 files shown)


solr/Documents/SolrDocument.php
```diff
@@ -179,7 +179,7 @@ public function getFieldNames() {}
      * (PECL solr &gt;= 0.9.2)<br/>
      * Returns a SolrInputDocument equivalent of the object
      * @link https://php.net/manual/en/solrdocument.getinputdocument.php
-     * @return SolrInputDocument <p>
+     * @return SolrInputDocument|null <p>
```
VincentLanglet, line 182: It say `Returns a SolrInputDocument on success and <b>NULL</b> on failure.`

isfedorov (changes requested):

solr/Queries/SolrQuery.php
```diff
@@ -82,14 +82,14 @@ public function addFacetDateField($dateField) {}
      * @param string $value <p>
      * The value to use.
      * </p>
-     * @param string $field_override <p>
+     * @param string $field_override [Optional] <p>
```
isfedorov, line 85: [Optional] can be removed here, actually. Optionality of a parameter is obvious from its default value. Stubs use `[optional]` in some other places because due to some reasons it's impossible to add a default value in signature, but it's not a case here

isfedorov (changes requested):

PR #1784, opened 2025-09-01
[phpstorm-stubs] Add support for PHP 8.5 in tests

isfedorov:


RobiNN1:
There is probably a problem here.

In this valid code, the round() function reports:
Function has a parameter whose default value is incompatible with its declared type

```php
$memory_info = $info['memory'];
$used_memory = (int) ($memory_info['used_memory']);
$max_memory = (int) ($memory_info['maxmemory']);
if ($max_memory > 0) {
    $memory_usage = round(($used_memory / $max_memory) * 100, 2);
}
```

isfedorov:
true. Already fixed it and soon will merge to master

PR #1794, opened 2025-09-12
Force array_first/-last input types through template

florianschaefer-jb:
Prior to this change, PHPStorm's type inference would always return `mixed` for the return types of `array_first` and `array_last`. By upgrading the input types to templates, we implicitly enforce the use of a more advanced generics type provider that deals with array elements correctly.


isfedorov:
Failures are related to updated php 8.5 beta 3 thus not related to the changes so merging PR

PR #1801, opened 2025-10-03
Improve class property type for reflections

VincentLanglet:


Reflection/ReflectionMethod.php
```diff
@@ -68,7 +68,7 @@ class ReflectionMethod extends ReflectionFunctionAbstract
      * </code>
      *
      * @link https://php.net/manual/en/reflectionmethod.construct.php
-     * @param string|object $objectOrMethod Classname, object
+     * @param class-string|object $objectOrMethod Classname, object
```
uuf6429, line 71: Note that when called with only one argument, the behaviour is different: the argument **must be a string** and in the form of **<fully-qualified-class>::<method name>** (deprecated as of PHP 8.4): https://www.php.net/manual/en/reflectionmethod.construct.php#:~:text=The%20alternative%20signature

I don't know if this logic can be somehow modelled here... I even doubt if PHPStan can handle it. But saying it is only `class-string` feels somewhat wrong.

On an unrelated note, the `@param` description here is a bit poor and unclear.

PR #1805, opened 2025-10-17
Various developer quality of life fixes

uuf6429:
- Renamed `ReflectionData.json` to `ReflectionData.dat` since it does not contain valid json (I picked 'dat' since there isn't a well-defined format for PHP and 'dat' is generic enough).
- `.gitignore`'d (and removed) `.idea/` since I couldn't see how this is helping anyone (and in fact made it worse by committing files by mistake)
- Fixed permissions of a few files (0755 to 0644) - done "for free" by git when removing `.idey`.
- Updated `README.md` to mention `runTests.sh` + a slight formatting improvement
- Slightly improved `runTests.sh` readability by adding a few blank lines
- Made `runTests.sh` support running against arbitrary PHP versions passed as parameters (instead of all versions)
  - Maybe it might make sense to use that script in the github workflow too? It would decrease repetitive code and ensure that the script is working.
- Fixed tiny typo and removed unused env var from github workflow to avoid confusion

Relates to https://youtrack.jetbrains.com/issue/WI-82914/PHPStorm-stubs-project-state

PR #1810, opened 2025-10-26
Improve Reflection API type hinting

uuf6429:
- in certain places, it's practically guaranteed that a non-empty-string is expected since symbols cannot have an empty string, so I updated them accordingly
- I've replaced `T` template type name with something more readable - in one case this even fixed a bug because `T` was defined for the entire class AND on a specific method.
- when possible, I replace the general `array` type with something more specific - list or indexed array etc.
- added some missing templates types, mainly on constructors e.g. `@template TReflectedClass of UnitEnum`

Dislcaimer:
- I was not able to run tests locally, as per https://github.com/JetBrains/phpstorm-stubs/pull/1805#issuecomment-3418666930
- It's likely this change might cause static analyser errors in projects using it.

E.g. the following would now fail (but it [should have  already](https://3v4l.org/cAEln)): https://phpstan.org/r/09d7d483-6aad-41df-805e-75d44edbec32
_It's worth noting that now it would also fail in "unsecured" cases e.g. if the method name was retrieved from somewhere returning `string` instead of `non-empty-string`_


Reflection/ReflectionAttribute.php
```diff
@@ -5,10 +5,13 @@
 /**
  * @since 8.0
  *
- * @template T of object
+ * @template TAttributeClass of Attribute
```
SerafimArts: Why is there a bound on the `\Attribute` here? If this is a reflection of an arbitrary `object` that is used as an attribute?

Reflection/ReflectionEnum.php
```diff
@@ -17,12 +21,13 @@ public function __construct(object|string $objectOrClass) {}
     public function hasCase(string $name): bool {}
 
     /**
-     * @return ReflectionEnumUnitCase[]|ReflectionEnumBackedCase[]
+     * @return (TReflectedClass is BackedEnum ? list<ReflectionEnumBackedCase<TReflectedClass>> : list<ReflectionEnumUnitCase<TReflectedClass>>)
```
SerafimArts, line 24: PHPStorm doesn't support conditional types =(

so it seems more appropriate to declare them in the return statement via a union, and for tools with prefixes, i.e.:
```php
@phpstan-return (TReflectedClass is BackedEnum 
    ? list<ReflectionEnumBackedCase<TReflectedClass>> 
    : list<ReflectionEnumUnitCase<TReflectedClass>>)
@return list<ReflectionEnumBackedCase<TReflectedClass>>|list<ReflectionEnumUnitCase<TReflectedClass>>
```

What do you think?

Reflection/ReflectionFunctionAbstract.php
```diff
@@ -316,13 +316,13 @@ public function returnsReference(): bool {}
     public function hasReturnType(): bool {}
 
     /**
-     * @template T
+     * @template TAttributeClass of Attribute
```
SerafimArts: This is the same problem as above. It should be an `object`, not an instance of `\Attribute`.

Reflection/ReflectionProperty.php
```diff
@@ -91,8 +92,8 @@ class ReflectionProperty implements Reflector
      * Construct a ReflectionProperty object
      *
      * @link https://php.net/manual/en/reflectionproperty.construct.php
-     * @param class-string|object $class The class name, that contains the property.
-     * @param string $property The name of the property being reflected.
+     * @param class-string<TReflectedClass>|TReflectedClass $class The class name, that contains the property.
+     * @param non-empty-string $property The name of the property being reflected.
```
SerafimArts: The property name may be empty.

```php
$object = (object)['' => 42];

echo $object->{''}; // int(42)

var_dump(new ReflectionProperty($object, '')); // object(ReflectionProperty) { name: string(''), ... }
```

Reflection/ReflectionProperty.php
```diff
@@ -26,14 +27,14 @@ class ReflectionProperty implements Reflector
     public const IS_VIRTUAL = 512;
 
     /**
-     * @var string Name of the property, same as calling the {@see ReflectionProperty::getName()} method
+     * @var non-empty-string Name of the property, same as calling the {@see ReflectionProperty::getName()} method
```
SerafimArts: Property name can be empty

Reflection/ReflectionParameter.php
```diff
@@ -277,13 +277,13 @@ public function isVariadic(): bool {}
     public function isPromoted(): bool {}
 
     /**
-     * @template T
+     * @template TAttributeClass of Attribute
```
SerafimArts: Docblocks should be structured in the following order:
- Title
- Description
- Annotations

So, it's best to move the `@template` docblock after the class description, if that's not too much trouble.

Reflection/ReflectionAttribute.php
uuf6429: I guess at the time I was thinking that attributes must extend/implement some common interface, but that is not true.

Reflection/ReflectionEnum.php
uuf6429: agreed.

Reflection/ReflectionProperty.php
uuf6429: I could have sworn it didn't work when I tried it. 🤔

uuf6429: On the other hand, it seems this doesn't work with static properties:
- https://3v4l.org/EcMS0
- https://3v4l.org/Cd5bu

Reflection/ReflectionParameter.php
```diff
@@ -81,8 +81,7 @@ public function isPassedByReference(): bool {}
      * Returns whether this parameter can be passed by value
      *
      * @link https://php.net/manual/en/reflectionparameter.canbepassedbyvalue.php
-     * @return bool|null {@see true} if the parameter can be passed by value, {@see false} otherwise.
-     * Returns {@see null} in case of an error.
```
uuf6429, line 85: It never really returns null, that seems to be a documentation bug (I'll report it separately).

See: https://github.com/php/php-src/blob/3ed80a154d589e3ea1b02f43fa1899ea9c46d70f/ext/reflection/php_reflection.c#L2872

uuf6429: Edit: Apparently it might have returned null in the past: https://github.com/php/doc-en/pull/5563

Is there any sane way of "deprecating" the old usage?

Reflection/ReflectionProperty.php
SerafimArts: Yes, I haven't tested static fields for this behavior.

SerafimArts:
> [uuf6429](https://github.com/uuf6429) requested a review from [SerafimArts](https://github.com/SerafimArts) [4 hours ago](https://github.com/JetBrains/phpstorm-stubs/pull/1810#event-25609845123)

I don't work for JetBrains and have no right to review their code.

All my comments arose because I accidentally saw this PR with some mistakes and decided to suggest some minor improvements =)

Reflection/ReflectionEnum.php
isfedorov: Sorry, but I have quite an opposite opinion. It's better to make return type as `@return (TReflectedClass is BackedEnum 
    ? list<ReflectionEnumBackedCase<TReflectedClass>> 
    : list<ReflectionEnumUnitCase<TReflectedClass>>)` and if some cases don't work in PhpStorm, create a related youtrack issue since in general support of conditional return types was done in [WI-66465](https://youtrack.jetbrains.com/issue/WI-66465/Support-parsing-of-PhpStans-conditional-return-types), [WI-69618](https://youtrack.jetbrains.com/issue/WI-69618/Conditional-return-type-Support-templates-in-condition), [WI-69644](https://youtrack.jetbrains.com/issue/WI-69644/Add-support-of-parsing-for-multiline-conditional-return-type), [WI-78351](https://youtrack.jetbrains.com/issue/WI-78351/Type-not-correctly-inferred-from-phpstan-psalm-conditional-generic-return-type)

isfedorov:
@uuf6429 Other than my comment regarding conditional return types this PR looks good to me! Thank you! Unfortunately, there were a lot of changes in the test framework, and due to this, the PR has conflicts with master. Could you please sync the latest version of the repo and update your PR? After that, we will merge it

isfedorov:
@uuf6429 Also, please note that since test framework has been changed, it's required now to regenerate caches of parsed stubs (see README.md for details) but TLDR; you need to run `docker compose -f docker-compose.yml run --rm test_runner php tests/run-stubs-parser.php` to generate those caches and commit caches along with other changes

Reflection/ReflectionEnum.php
uuf6429: I just tried it out:
```php
/**
 * @param bool $return
 * @return ($return is true ? ReflectionClass: null)
 */
function test(bool $return)
{
    return $return ? new ReflectionClass(ReflectionClass::class) : null;
}


echo test(false)->name; // PHPStorm does not complain about name (it should),
                        // but it does suggest symbols from ReflectionClass,
                        // so it at least it seems to partially work.
```

For me either way is fine. I'll change it as suggested. 👍

uuf6429:
@isfedorov I did the requested changes. Seems there's a lot of "noise" in the stubs cache, but it looks like it could be an issue about (lack of?) sorting?

isfedorov:
@uuf6429 
> but it looks like it could be an issue about (lack of?) sorting?

Exactly! Working on it right now.

isfedorov:
@uuf6429 please try to rebase one more time

uuf6429:
Hmm... now I don't know why a tonne of lines have been simply removed (despite running the tool multiple times in docker). 🤔 

@isfedorov one thing is sure - don't forget to squash (and not just merge) the changes when the PR is ready.

isfedorov:
@uuf6429 I've restricted the used reflection docker images to linux/amd64 and sorted records, so now the runs on any machines should produce equal reflection/stubs cache data.  Could you please try rebasing and force-pushing one more time? Sorry for missing these nuances in the first place

uuf6429:
@isfedorov no problem - it would have had to happen at some point anyway, so sooner is probably better than later.

That being said, the problem seems to still persist. I give it a quick look - you know that the stubs thingy is running in a "plain" php 8.5 alpine image? I don't think it would find things like mongodb, pthreads and so on.

isfedorov:
@uuf6429 
> That being said, the problem seems to still persist

I think I've found a root cause, but I'll fix it after merge of you PR. I'll just push fix with new cache files.

uuf6429:
OK, let me know if I can do anything else. Feel free to merge/revert changes into my PR, if needs be.

PR #1836, opened 2026-01-13
[phpstorm-stubs] Update `#[Pure]` attribute in `mb_str_pad` to `#[Pure(true)]`

LolGleb:


staabm:
whats the difference of `mb_str_pad` in comparison to all other `mb_str_*` functions so it is `#[Pure(true)]` but all others stay/are `#[Pure]`?

staabm:
https://github.com/JetBrains/phpstorm-stubs/pull/1793#discussion_r2349668514

> Shouldn't the functions be marked as #[Pure(true)] instead, since AFAIU they can depend on global state if $encoding parameter isn't passed. In this case, mb_internal_encoding() will be used, which can depend on the global state. Could you please double-check?

I think this comment said, that all functions which depend on `$encoding` should be `#[Pure(true)]` (or not)?

PR #1860, opened 2026-03-16
Narrow ReflectionClass::getConstant() return-type

staabm:
per https://www.php.net/manual/en/language.oop5.constants.php constants can only ever hold certain values

> class constants can have a scalar type such as bool, int, float, string, or even array. When using array, the contents can only be other scalar types.


VincentLanglet:
cc @staabm @isfedorov This is wrong, there is also NULL
See https://3v4l.org/69Jraf#v8.5.3

I opened https://github.com/JetBrains/phpstorm-stubs/pull/1879

PR #1867, opened 2026-04-02
Remove `Serializable` interface from GMP class

LeonMelis:
Trying to use `$gmp->serialize()` or `$gmp->unserialize()` as defined in the stubs will result in `Error: Call to undefined method GMP::serialize()`

This is because GMP class does not actually implement the `Serializable` interface (although it does implement `__serialize()` and `__unserialize`). Note that with those functions present, PHP could have implement the serializable class for GMP, but probably omitted that deliberately as PHP does not provide an object-oriented interface to manipulate GMP objects, only a procedural GMP API. 

See: https://www.php.net/manual/en/class.gmp.php

> _"No object-oriented interface is provided to manipulate GMP objects. Please use the procedural GMP API."_

This can be confirmed in the PHP source code, the GMP class does not implement the `Serializable` interface, see: https://github.com/php/php-src/blob/master/ext/gmp/gmp.c#L584

```c
gmp_ce = register_class_GMP();
```

If `Serializable` interface was implemented in the GMP class in the PHP source code, then we would have expected to see:

```c
gmp_ce = register_class_GMP(zend_ce_serializable);
```

### DIFF
--- a/gmp/gmp.php
+++ b/gmp/gmp.php
@@ -798,7 +798,7 @@ define('GMP_VERSION', "6.3.0");
 
 define('GMP_MPIR_VERSION', '3.0.0');
 
-final class GMP implements Serializable
+final class GMP
 {
     /**
      * @since 8.2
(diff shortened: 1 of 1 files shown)


stof:
> Note that with those functions present, PHP could have implement the serializable class for GMP, but probably omitted that deliberately as PHP does not provide an object-oriented interface to manipulate GMP objects, only a procedural GMP API.

This statement is wrong. `__serialize` and `__unserialize` are a *separate* serialization mechanism than `Serializable`.

The reason why `Serializable` is not implemented is because GMP uses the modern serialization mechanism.

PR #1874, opened 2026-04-17
Update throws annotation on createFromDateString

VincentLanglet:
Hi @isfedorov 

See https://www.php.net/manual/en/dateinterval.createfromdatestring.php
Since PHP 8.3 the method can now throws `DateMalformedStringException`

Since there is no LanguageLevelTypeAware about throw type, I think I have to duplicate the stub with a PhpStormStubsElementAvailable element.

NB: The LanguageLevelAware should be about 8.3 and not 8.4, but if I try to fix it, tests are failing ; is there something we can do ?

### DIFF
--- a/date/date_c.php
+++ b/date/date_c.php
@@ -1127,7 +1127,21 @@ class DateInterval
      * @link https://php.net/manual/en/dateinterval.createfromdatestring.php
      */
     #[TentativeType]
-    #[LanguageLevelTypeAware(['8.4' => 'DateInterval'], default: 'DateInterval|false')]
+    #[PhpStormStubsElementAvailable(from: '5.3', to: '8.2')]
+    #[LanguageLevelTypeAware(['8.3' => 'DateInterval'], default: 'DateInterval|false')]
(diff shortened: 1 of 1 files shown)


LolGleb:
Thank you for the PR! I've checked why this happens using this piece of code:
```
<?php

$method = new ReflectionMethod(DateInterval::class, 'createFromDateString');

echo 'PHP version: ' . PHP_VERSION . PHP_EOL;
echo 'Return type: ' . ($method->getTentativeReturnType() ?? 'none') . PHP_EOL;
```
The results were:
```
PHP version: 8.3.30                                                                                                                                                           
Return type: DateInterval|false

PHP version: 8.4.20                                                                                                                                                           
Return type: DateInterval
```
This confirms that the return type declaration in PHP source code was updated from `DateInterval|false` to `DateInterval` only in version 8.4, which means `LanguageLevelTypeAware(['8.4' => 'DateInterval'])` is exactly right. The tests were correct and your PR is absolutely valid. LGTM:)

PR #1878, opened 2026-04-25
Add ext-ds v2 stubs

simPod:
## Summary
- Add a separate `ds_v2` stub directory for ext-ds 2.0's breaking API changes.
- Register `ds_v2` as an alternate non-core stub directory and exclude it from generated stub map output, matching the `couchbase_v2` pattern.
- Include v2 symbols such as `Ds\Seq`, `Ds\Heap`, `Ds\Key`, and functional constructors.

### DIFF
--- a/ds_v2/ds.php
+++ b/ds_v2/ds.php
@@ -1,0 +1,629 @@
+<?php
+
+/**
+ * PHP Data Structure stubs, a PECL extension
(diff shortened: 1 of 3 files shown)


simPod:
I don't know how to correctly approach v2 of ext here

PR #1879, opened 2026-04-25
Revert getConstant return type narrow

VincentLanglet:
It can be a scalar OR NULL
Reflection/ReflectionClass.php

Edit: `object` is also possible see https://3v4l.org/aD9Bc#v8.5.3
And nested arrays are allowed too `array<array<scalar>>` for instance...

I propose to revert the change which didn't had lot of values.

cc @isfedorov @LolGleb


LolGleb:
Thanks for catching this! You're right that `scalar|array<scalar>` was too narrow. Constant values can also be `null` or an `object`, so I believe that reverting to `mixed` is the correct approach here. Merging!

PR #1888, opened 2026-05-04
Fix DateInterval::createFromDateString() exception type

mpesari:
The PHP documentation had wrong exception type for `DateInterval::createFromDateString()`:

- https://github.com/php/doc-en/pull/5521


VincentLanglet:
I agree with this @LolGleb ; I made the mistake in https://github.com/JetBrains/phpstorm-stubs/pull/1874

PR #1889, opened 2026-05-05
Correct type of $matches in preg_match_all()

AJenbo:


staabm:
please show example code

AJenbo:
https://github.com/AJenbo/agcms/blob/820bf8b578f66d38b8d58e8947f05f82fe9076d2/application/inc/Http/Controllers/Admin/ExportController.php#L283

```php
        $matches = null;
        $success = preg_match_all(
            '/<img[^>]+src="([^"]+)"[^>]*>/iu',
            $html,
            $matches
        );
```

https://www.php.net/manual/en/function.preg-match-all.php

> [array](https://www.php.net/manual/en/language.types.array.php) &$matches = [null](https://www.php.net/manual/en/reserved.constants.php#constant.null),

PR #1890, opened 2026-05-07
WI-84821 SNMP2 and SNMP3 stubs $host param should be $hostname

StefNieuwhuis:
[WI-84821](https://youtrack.jetbrains.com/issue/WI-84821) SNMP2 and SNMP3 stubs $host param should be $hostname

Changed $host paramter to $hostname in all snmp2 and snmp3 functions to be consistent with the actual extension.


PhpStormStubsMap.php
StefNieuwhuis, line 1: uh, just followed the README don't know if this should have happened tho..

LolGleb:
Thanks for the PR. A couple of asks before this can land:
1. Drop the `PhpStormStubsMap.php` changes please. The proposed map is missing valid mappings;
2. Fix `snmp2_real_walk docblock`. Line 601 still says `@param string $host` while the signature on line 619 was renamed to `$hostname`;
3. (Optional) Would you mind covering the snmp1 family too? `snmpgetnext`, `snmprealwalk`, and `snmpset` still use `$host` in both signature and docblock, and php.net documents them as `$hostname`. Otherwise the file is left half-converted.

PR #1896, opened 2026-06-07
Further reflection api improvements

uuf6429:
Changes:
- fixes a mistake in my previous reflection PR regarding getAttributes() return type
- changes method name strings to `non-falsy-string` to align with phpstan stubs
- improves type hinting of reflaction class initializers (heavily inspired from phpstan stubs)
- since reflection classes are now mainly generic, fixed the ones with undefined templates
- a few other minor fixes

Some relevant links:
- https://phpstan.org/r/78cf39ee-1d76-4e14-b2a6-990d856f994c
- https://3v4l.org/S0Oon#v8.5.7

### DIFF
--- a/Reflection/ReflectionAttribute.php
+++ b/Reflection/ReflectionAttribute.php
@@ -5,7 +5,7 @@ use JetBrains\PhpStorm\Pure;
 /**
  * @since 8.0
  *
- * @template TAttributeClass of object
+ * @template-covariant TAttributeClass of object
  */
 class ReflectionAttribute implements Reflector
 {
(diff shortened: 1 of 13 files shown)


uuf6429:
@ondrejmirtes when you have time, could you maybe let me know if I'm missing something? I'd gladly do the relevant changes on phpstan side as well (as I started already).

My remaining question so far is the 2x stub definitions in phpstan regarding the `*WithLazyObjects` cases, if/how I should handle that (in particular, I don't understand why they would require `@template-covariant`).

PR #1897, opened 2026-06-08
Revert "Improve Reflection API type hinting (#1810)"

uuf6429:
This reverts commit c96a600691033709d31b689d31ff5f385885cdcd due to backward compatibility issues as [described here](https://github.com/phpstan/phpstan/discussions/14788#discussioncomment-17216591).

Short term: the revert allows phpstan to keep making use of these stubs for the moment.
Long term: I was already going to contribute some further changes with #1896, so instead I'll put the revert changes in there and then we'll see how to have this release without stepping on anyones' feet.


uuf6429:
Thanks @isfedorov and apologies for the extra work.

PR #1899, opened 2026-06-11
Add FileReference attribute for file reference injections

LevStolpner:


isfedorov:
Will fix failing test separately as it's not related to changes

PR #1907, opened 2026-07-15
replace secure.php.net with www.php.net

jordikroon:
This PR normalises the links to php.net as a few variants live within the stubs. 
At least we should move towards https not depend on secure.php.net. So if this change is too much, we can get rid of the second commit.

Please see commits 1 by 1:

Commit 1: replaces secure.php.net with www.php.net
```bash
LC_ALL=C find . -type f -exec sed -i '' 's/secure\.php\.net/www.php.net/g' {} +
```

Commit 2: replaces https://php.net with https://www.php.net
```bash
LC_ALL=C find . -type f -exec sed -i '' 's|https://php\.net|https://www.php.net|g' {} +
```

Commit 3: replaces http://php.net with https://www.php.net
```bash
LC_ALL=C find . -type f -exec sed -i '' 's|http://php\.net|https://www.php.net|g' {} +
```

Commit 4: fixes tests, intentionally using http:// here.


isfedorov:
@jordikroon That's a good point, but could you please clarify, what is a benefit of normalizing links to `//www.php.net` instead of `//php.net` (other than avoiding 301 redirect)? It's longer form and amount of required changes is huge (stubs have ~1300 occurrences of `www.php.net` while ~8800 of `php.net`). So I agree that links should be normalized, but I'd suggest to do it in a vice versa way. All existing `www.php.net` and `secure.php.net` convert to `//php.net`. Also please update related stubs check to verify that new added links are aligned with this rule (see `PhpDocLinksCheck::checkLinks`) and unit tests that will test this new change in PhpDocLinksCheck (see PhpDocLinksCheckTest). Oh, and please pay attention to https://github.com/JetBrains/phpstorm-stubs/blob/master/CONTRIBUTING.md#cache-files and avoid committing those Reflection*.json files.

isfedorov:
Fixed in scope of other massive PHPDoc fix

PR #1909, opened 2026-07-21
Update Redis classes

RobiNN1:
Added missing methods, docsblocks and fixed return types in all redis classes. It should now be up to date with the latest phpredis stubs.

Also in RedisSentinel is 2x __construct(). Can I remove old one, or it's for compatibility? It probably cause this issue in PHPStan `Class RedisSentinel constructor invoked with 1 parameter, 2-6 required.`
I haven't found RedisSentinel in PHPStan signatures so it need to be fixed here I guess.

### DIFF
--- a/redis/Redis.php
+++ b/redis/Redis.php
@@ -105,6 +105,12 @@ class Redis
      */
     public function __construct(?array $options = null) {}
 
+    /**
+     * Destructor to clean up the Redis object.
+     *
+     * This method will disconnect from Redis. If the connection is persistent
+     * it will be stashed for future reuse.
+     */
     public function __destruct() {}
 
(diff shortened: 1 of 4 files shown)


isfedorov:
@RobiNN1 constructor with 1 parameter was added in scope of https://youtrack.jetbrains.com/issue/WI-75592
Screenshot 2026-07-22 at 09 26 59
And it's still actual https://github.com/phpredis/phpredis/blob/release/6.3.0/sentinel.md

RobiNN1:
I want to remove old with multiple parameters.


@isfedorov should be ok now, merged constructors so it accept both old version and new with array options.

PR #1919, opened 2026-08-03
`ReflectionType::__toString()` is not deprecated

cs278:
This method has been deprecated on and off, but is currently not deprecated.

https://github.com/phpstan/phpstan/issues/14963


cs278:
Not sure what I'm supposed to do with the test failure, yes the method is marked as deprecated in 7.4 but it was never removed and undeprecated in 8.0 so yeah. :shrug:

isfedorov:
@cs278 Right, thank you for this PR. I'll update test suite and add this case to known problems to avoid test failures.

PR #1936, opened 2026-08-19
[dom] Remove non-Dom class stubs from the `Dom` namespace

kaznovac:
Removed 'stranded' stubs in `Dom` namespace:
- Sqlite (Pdo)
- Mysql (Pdo)
- RandomError (Random)
- BrokenRandomEngineError (Random)
- RandomException (Random)


isfedorov:
@kaznovac Thank you for the fix! These are stale classes indeed! Unfortunately, the test framework is one-directional, that's. why didn't catch this problem

PR #1940, opened 2026-09-06
fix `pg_loopen` phpdoc return type

staabm:
it should be analog `pg_lo_open` (in the same file)

### DIFF
--- a/pgsql/pgsql.php
+++ b/pgsql/pgsql.php
@@ -1977,7 +1977,7 @@ function pg_lounlink(
  * @param $connection
  * @param $oid [optional]
  * @param $mode [optional]
- * @return resource
+ * @return resource|false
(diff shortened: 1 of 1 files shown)


pgsql/pgsql.php
```diff
@@ -1977,7 +1977,7 @@ function pg_lounlink(
  * @param $connection
  * @param $oid [optional]
  * @param $mode [optional]
- * @return resource
+ * @return resource|false
  * @deprecated 8.0
  */
 #[LanguageLevelTypeAware(['8.1' => '\PgSql\Lob|false'], default: 'resource|false')]
```
staabm, line 1983: after the fix, the phpdoc return type is compatible with this annotation based default-type

